PR #10, opened 2019-03-02
Tech :: move out step deifnitions in a specific module

adimascio:
It should be simple to read and maintain as this. On the way, fixed the steps documentation and introduce a more formal type definition for each step.

See individual commits for more explanation. (This PR is not supposed to be squashed before merging)


piotch:
This is not directly linked to this PR but since we are fixing the doc: `This step is meant to select a specific domain (using MongoDB terminology).` sounds off, "domain" is not Mongo terminology, it's Toucan, maybe you meant "collection" instead of "domain" ?

src/lib/steps.ts
```diff
@@ -0,0 +1,51 @@
...
+  oldname: string;
+  newname: string;
+}
+
+export interface DeleteStep {
+  name: 'delete';
+  columns: Array<string>;
+}
+
+export interface NewColumnStep {
```
piotch: this one is not documented

adimascio: :+1:

adimascio:
> This is not directly linked to this PR but since we are fixing the doc: `This step is meant to select a specific domain (using MongoDB terminology).` sounds off, "domain" is not Mongo terminology, it's Toucan, maybe you meant "collection" instead of "domain" ?

I agree that "domain" might be a bad wording. "Collection" on the other hand might sound too much mongo-ish, don't you think? What about simply "dataset" ?

piotch:
> I agree that "domain" might be a bad wording. "Collection" on the other hand might sound too much mongo-ish, don't you think? What about simply "dataset" ?

dataset 👍

piotch:
sorry, button confusion

src/lib/steps.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * This module defines the supported unit-of-transformation steps.
+ */
+
+export interface DomainStep {
```
Dalanir: Rename to `selectDataset` I guess following previous comments

adimascio: :+1: I will do this in a distinct PR once this one is merged.

PR #12, opened 2019-03-03
[doc] document `newcolumn` step

adimascio:
Add missing documentation about the `newcolumn` step.

### DIFF
--- a/doc/steps.md
+++ b/doc/steps.md
@@ -68,6 +68,18 @@ Delete a column.
 }
 ```
 
+### `newcolumn` step
+
+Compute a new column based on another one or a custom query.
+
+```javascript
+{
+    name: 'newcolumn',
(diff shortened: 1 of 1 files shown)


doc/steps.md
```diff
@@ -68,6 +68,18 @@ Delete a column.
...
 
+### `newcolumn` step
+
+Compute a new column based on another one or a custom query.
+
+```javascript
+{
+    name: 'newcolumn',
+    column: 'my-new-column',
+    query: { $concat: ['$col1', ' - ', '$col2'] }
```
Dalanir: Why propose this alternative ?

adimascio: In order to avoid designing (for now) an explicit language to perform operations on other columns, we just specify the transformation in terms of the target query langage (e.g. mongo in this specific case).

PR #17, opened 2019-03-13
tech :: update example App component to match valid step types

adimascio:
Use `Arrray<PipelineStep>` instead of `Array<any>` in `App.vue` component.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,33 +1,29 @@
 <template>
   <div id="app">
     <img alt="Vue logo" src="./assets/logo.png">
-    <Pipeline :steps="steps" :domains-list="domainsList" @selectedPipeline="setSteps" />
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -1,33 +1,29 @@
 <template>
   <div id="app">
     <img alt="Vue logo" src="./assets/logo.png">
-    <Pipeline :steps="steps" :domains-list="domainsList" @selectedPipeline="setSteps" />
+    <Pipeline :steps="steps" :domains-list="domainsList" @selectedPipeline="setSteps"/>
```
davinov: Maybe we could set up [Prettier](https://github.com/prettier/prettier) on this project to enforce a standard code-style?
Or/also: set up pug, which is a very light and convenient way of wring templates

PR #19, opened 2019-03-14
tech :: Convert vue components to TypeScript

davinov:
So we can benefit from type checking inside vue components.

Docs to learn how to use these decorators and class syntax:
- https://vuejs.org/v2/guide/typescript.html
- https://github.com/vuejs/vue-class-component
- https://github.com/kaorun343/vue-property-decorator

Note: storybook now uses a specific bundle, also built with the same rollup configuration.
It's therefore needed to export any component to showcase in `.storybook/bundle.ts`.

### DIFF
--- a/.storybook/bundle.ts
+++ b/.storybook/bundle.ts
@@ -1,0 +1,14 @@
+/*
+  This file creates the bundle needed by storybook to import the UI components.
+  All components available in the Storybook must be exported here.
+*/
+
+import DomainSelector from '../src/components/DomainSelector.vue'
+import Pipeline from '../src/components/Pipeline.vue'
+import Step from '../src/components/Step.vue'
+
+export {
+  DomainSelector,
+  Pipeline,
+  Step,
+}
--- a/.storybook/rollup.config.js
+++ b/.storybook/rollup.config.js
@@ -1,0 +1,13 @@
+import mainConfig from '../rollup.config'
+
+export default {
+  ...mainConfig,
+  input: '.storybook/bundle.ts',
+  output: [
+      {
+          file: 'dist/storybook/components.js',
+          format: 'esm'
+      }
+  ],
+}
+
--- a/src/components/DomainSelector.vue
+++ b/src/components/DomainSelector.vue
@@ -24,7 +24,10 @@ import { Component, Prop } from 'vue-property-decorator'
   name: 'domain-selector'
 })
 export default class DomainSelector extends Vue {
-  @Prop({ default: () => [] })
+  @Prop({
+    default: () => [],
+    type: Array
+  })
   readonly domainsList!: Array<string>
 
   @Prop()
--- a/src/components/DomainSelector.vue
+++ b/src/components/DomainSelector.vue
@@ -16,22 +16,28 @@
   </div>
 </template>
 
-<script>
-export default {
-  name: 'domain-selector',
-  props: {
-    domainsList: Array,
-    selectedDomain: String,
-  },
-  methods: {
-    isDomainSelected(name) {
-      return this.selectedDomain === name;
-    },
-    select(newDomain) {
-      this.$emit('selectedDomain', newDomain);
-    },
-  },
-};
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+
+@Component({
+  name: 'domain-selector'
+})
+export default class DomainSelector extends Vue {
+  @Prop({ default: () => [] })
+  readonly domainsList!: Array<string>
+
+  @Prop()
+  readonly selectedDomain!: string | undefined
+
+  isDomainSelected(name: string) {
+    return this.selectedDomain === name;
+  }
+
+  select(newDomain: string) {
+    this.$emit('selectedDomain', newDomain);
+  }
+}
 </script>
 
 <style lang="scss" scoped>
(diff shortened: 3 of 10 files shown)


src/components/DomainSelector.vue
```diff
@@ -16,22 +16,31 @@
...
+
+@Component({
+  name: 'domain-selector'
+})
+export default class DomainSelector extends Vue {
+  @Prop({
+    default: () => [],
+    type: Array
+  })
+  readonly domainsList!: Array<string>
```
davinov: So there is two type declaration here:
- the one in the decorator @Prop is meant for vue, and will be enforced at _run time_
- the other directly after the prop name is meant for typescript, and will be enforced at _transpile time_

.storybook/bundle.ts
```diff
@@ -0,0 +1,16 @@
+/*
+  This file creates the bundle needed by storybook to import the UI components.
+  All components available in the Storybook must be exported here.
+*/
+
+import Cell from '../src/components/dataset-viewer/Cell.vue'
```
adimascio: This component seems to be part of another PR, you should remove the import here.

.storybook/rollup.config.js
```diff
@@ -0,0 +1,14 @@
+import mainConfig from '../rollup.config'
+
+let storybookComponentsPackageConfig = Object.assign({}, mainConfig)
```
adimascio: Use `const` instead of `let` ? I don't see any new assignment to `storybookComponentsPackageConfig`.

Also, this is a just a matter of style but I like to use the spread operator for object assignments. Here, `Object.assign{{}, mainconfig)` could become `{...mainConfig}`. What do you think?

.storybook/rollup.config.js
```diff
@@ -0,0 +1,14 @@
...
+storybookComponentsPackageConfig.input = '.storybook/bundle.ts'
+storybookComponentsPackageConfig.output = [
+  {
+    file: 'dist/storybook/components.js',
+    format: 'esm'
+  }
+]
+
+export default storybookComponentsPackageConfig
+
```
adimascio: Actually, the whole assignment could become:

```javascript

export default {
    ...mainConfig,
    input: '.storybook/bundle.ts',
    output: [
        {
            file: 'dist/storybook/components.js',
            format: 'esm'
        }
    ],
};
```

adimascio: again, this is purely a matter of stlye :)

src/components/DomainSelector.vue
adimascio: :+1:

src/components/DomainSelector.vue
```diff
@@ -16,22 +16,31 @@
...
+<script lang="ts">
+import Vue from 'vue'
+import { Component, Prop } from 'vue-property-decorator'
+
+@Component({
+  name: 'domain-selector'
+})
+export default class DomainSelector extends Vue {
+  @Prop({
+    default: () => [],
```
adimascio: Why not `deafult: []` directly? Is it a matter of mutability and default being evaluated once and for all?

src/components/DomainSelector.vue
```diff
@@ -16,22 +16,31 @@
...
+    type: Array
+  })
+  readonly domainsList!: Array<string>
+
+  @Prop()
+  readonly selectedDomain!: string | undefined
```
Dalanir: the bang operator is not supposed to exclude `undefined` value ?

src/components/DomainSelector.vue
adimascio: :+1:

src/components/DomainSelector.vue
adimascio: The `!` tells the typescript compiler not to crash if the attribute ist not initialized in constructor. In case of props, I guess it's a bit trickier. Probably related to https://github.com/vuejs/vue-class-component/issues/221

.storybook/rollup.config.js
davinov: And clarity, I strongly agree ;)

.storybook/rollup.config.js
davinov: I'm not that sure here: I know that `const` only forbids the identifier to be reassigned, yet it passes the message that is object is quite "constant", i.e. immutable.

The lines just below reassigned properties of this object, so I was not that inclined to use `const`.

src/components/DomainSelector.vue
davinov: Yes, otherwise it's shared among every instance of this component: https://vuejs.org/v2/api/#props

.storybook/bundle.ts
davinov: Oops, it's effectively in the next one ;)

.storybook/rollup.config.js
adimascio: I understand but still, I disagree :-). My preference would be to stick with the language semantic and maybe even enforce the eslint `prefer-const: "error"` rule. Any objection?

src/components/DomainSelector.vue
adimascio: :+1:

src/components/DomainSelector.vue
davinov: Yep, @adimascio is right: as vue will assign itself the prop depending on what the parent component bound to it, we do not pass an initiaa value

vue-class-component suggest to use the `!` to still get type hinting on props

davinov:
@adimascio Thanks for the feedback, I've made the changes and rebased for clarity

PR #21, opened 2019-03-14
Dataset viewer :: cell

davinov:
Add a basic `Cell` component, customizable enough to:
- display raw and formatted values, with custom formatters
- conditionally add css class modifiers

### DIFF
--- a/src/components/dataset-viewer/Cell.vue
+++ b/src/components/dataset-viewer/Cell.vue
@@ -69,7 +69,7 @@ export default Vue.extend({
 })
 </script>
 
-<style lang="scss" scoped>
+<style lang="scss">
   @import '../../styles/Cell';
 </style>
 
--- a/src/components/dataset-viewer/Cell.vue
+++ b/src/components/dataset-viewer/Cell.vue
@@ -4,10 +4,12 @@ import Vue from 'vue'
 
 import detectType from './detect-type'
 
-const defaultFormatter = (value: any, row: object, column: string) =>
+const defaultFormatter = (value: any, row: object, column: string): string =>
   _.isObject(value) ? JSON.stringify(value) : _.toString(value)
 
 const CSS_CLASS = 'dataset-viewer__cell'
+
+type ClassModifierGenerator = (value: any, row: object, column: string, displayedValue: string) => string | undefined
 
 export default Vue.extend({
   name: 'dataset-viewer-cell',
@@ -36,7 +38,7 @@ export default Vue.extend({
     let classes = [CSS_CLASS]
     let type = detectType(value)
 
-    let displayedValue
+    let displayedValue: string
     if (ctx.props.displayFormatted) {
       classes.push(`${CSS_CLASS}--formatted`)
       displayedValue = ctx.props.formatter(value, ctx.props.row, ctx.props.column)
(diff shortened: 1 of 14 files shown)


src/components/dataset-viewer/Cell.vue
```diff
@@ -0,0 +1,75 @@
...
+
+    let displayedValue: string
+    if (ctx.props.displayFormatted) {
+      classes.push(`${CSS_CLASS}--formatted`)
+      displayedValue = ctx.props.formatter(value, ctx.props.row, ctx.props.column)
+    } else {
+      displayedValue = defaultFormatter(value, ctx.props.row, ctx.props.column)
+    }
+
+    classes = classes.concat(
```
adimascio: purely matter of style, but what do you think of `classes.push(...ctx.props.classModifiers.map(/* ... */))`? It might be _sligthly_ more consistent with the other `push` instructions above. (I let you choose to ignore this :))

adimascio: and then use `const` instead of `let` at variable definition :-)

src/components/dataset-viewer/Cell.vue
```diff
@@ -0,0 +1,75 @@
...
+    classModifiers: {
+      default: () => [],
+      type: Array
+    },
+  },
+
+  functional: true,
+
+  render (createElement, ctx) {
+    let value = ctx.props.row[ctx.props.column]
```
adimascio:
```suggestion
     const value = ctx.props.row[ctx.props.column]
```

src/components/dataset-viewer/Cell.vue
```diff
@@ -0,0 +1,75 @@
...
+    let classes = [CSS_CLASS]
+    let type = detectType(value)
```
adimascio:
```suggestion
    const type = detectType(value)
```

src/components/dataset-viewer/Cell.vue
```diff
@@ -0,0 +1,75 @@
...
+      displayedValue = defaultFormatter(value, ctx.props.row, ctx.props.column)
+    }
+
+    classes = classes.concat(
+      ctx.props.classModifiers
+        .map( (generateClassModifier: ClassModifierGenerator) =>
+          generateClassModifier(value, ctx.props.row, ctx.props.column, displayedValue)
+        )
+        // Filter out modifiers who generated nothing
+        .filter( (modifier: string | undefined) => modifier )
```
adimascio:
```suggestion
        .filter(Boolean)
```

The only caveat I know of is when we want to differentiate `0` and other falsy but that is also true but this is also true when filtering with the identity function

stories/cell.js
```diff
@@ -0,0 +1,26 @@
...
+          return {
+            column: "Some content"
+          }
+        }
+      }
+    },
+    template: `
+      <cell
+        :row="row"
+        :column="'column'"
```
adimascio: Any reason for the double quoting here?

src/components/dataset-viewer/Cell.vue
```diff
@@ -0,0 +1,75 @@
...
+    displayFormatted: {
+      default: false,
+      type: Boolean
+    },
+    formatter: {
+      default: defaultFormatter,
+      type: Function
+    },
+    classModifiers: {
+      default: () => [],
```
adimascio: Could we add some documentation about those props? Especially about class modifiers that seem to be chained?

davinov:
Note: we've decided not to share the table component with Tucana for now, so I'll need to rework this to remove features needed by Tucana that adds complexity (formatters in particular)

adimascio:
Closing this PR since we now chose another path, see https://github.com/ToucanToco/vue-query-builder/pull/104 for intsance.

PR #62, opened 2019-03-22
feature :: translator toolchain

adimascio:
Implement a simple translator toolchain to (hypothetically) support multiple backends.

For now, the only backend supported is `mongo3.6`. 

This PR also adds a typesafe `StepMatcher` interface that should help to detect, at compile-time,
step types that are not handled by the translation chain.

### DIFF
--- a/src/lib/matcher.ts
+++ b/src/lib/matcher.ts
@@ -1,0 +1,70 @@
+/**
+ * This module provides the basic type machinery to build pipeline
+ * translators efficiently and safely.
+ *
+ * It exposes 2 main building blocks:
+ * - `StepMatcher`: a type mapping each available pipeline steps to a
+ *   translation function
+ * - `matchStep`: a function accepting a `StepMatcher` and that return a function
+ *   transforming a pipeline step to an arbiratry output.
+ *
+ * This system is supposed to be typesafe, which means that if you add a
+ * new type of `PipelineStep`, the typescript compiler is expected to fail
+ * if your matcher implementation doesn't provide a mapping for this new kind
+ * of step.
+ */
+import { PipelineStep, PipelineStepName } from './steps';
+
+/**
+ * `StepByType` is a conditional type that expects the `PipelineStep` type and
+ * a valid step name and return the only possible step type matching this name.
+ *
+ * For instance, `StepByType<PipelineStep, 'delete'>` is equivalent to `DeleteStep`
+ */
+export type StepByType<A, T> = A extends { name: T } ? A : never;
+
+/**
+ * `StepMatcher` is a type built dynamically according to each available step
+ * name that associates each of this step name to a specific processing function
+ * accepting a step of corresponding type. For instance, it will generate the
+ * `rename: (step: RenameStep) => T` mapping entry automatically.
+ *
+ * This type will help to enforce that each kind of step has a corresponding
+ * translation function at compile time.
+ */
+export type StepMatcher<T> = { [K in PipelineStepName]: (step: StepByType<PipelineStep, K>) => T };
+
+/**
+ * Helper exception to handle unreachable conditions in a switch statement.
+ * This will help the typescript compiler to raise an error if one or several
+ * kind of steps are missing in a switch / case statement.
+ *
+ */
+class Unreachable extends Error {
+  constructor(value: never) {
+    super(`value ${value} is not valid`);
+  }
+}
+
+export function matchStep<T>(matcher: StepMatcher<T>): (step: PipelineStep) => T {
+  return (step: PipelineStep): T => {
+    switch (step.name) {
+      case 'domain':
+        return matcher.domain(step);
+      case 'filter':
+        return matcher.filter(step);
+      case 'select':
+        return matcher.select(step);
+      case 'rename':
+        return matcher.rename(step);
+      case 'delete':
+        return matcher.delete(step);
+      case 'newcolumn':
+        return matcher.newcolumn(step);
+      case 'custom':
+        return matcher.custom(step);
+      default:
+        throw new Unreachable(step);
+    }
+  };
+}
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -1,0 +1,91 @@
+/** This module contains mongo specific translation operations */
+
+import { PipelineStep } from '@/lib/steps';
+
+/**
+ * MongoStep interface. For now, it's basically an object with any property.
+ */
+export interface MongoStep {
+  [propName: string]: any;
+}
+
+/**
+ * Transform a standard pipeline into a list of mongo steps.
+ *
+ * - 'domain' steps are transformed into `$match` statements,
+ * - 'select', 'rename', 'delete' and 'newcolumn' steps are transformed into
+ *   `$project` statements,
+ * - 'filter' steps are transformed into `match` statements.
+ *
+ * @param pipeline the input pipeline
+ *
+ * @returns the list of corresponding mongo steps
+ */
+export function pipeToMongo(pipeline: Array<PipelineStep>): Array<MongoStep> {
+  const mongoSteps: Array<MongoStep> = [];
+  for (const step of pipeline) {
+    if (step.name === 'domain') {
+      mongoSteps.push({ $match: { domain: step.domain } });
+    } else if (step.name === 'select') {
+      const projection: { [propName: string]: number } = {};
+      for (const column of step.columns) {
+        projection[column] = 1;
+      }
+      mongoSteps.push({ $project: projection });
+    } else if (step.name === 'rename') {
+      mongoSteps.push({ $project: { [step.newname]: `$${step.oldname}` } });
+    } else if (step.name === 'delete') {
+      const projection: { [propName: string]: number } = {};
+      for (const column of step.columns) {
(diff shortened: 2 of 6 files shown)


piotch:
Great to see this already moving on! Thanks! 👍 

> (hypothetically) support multiple backends

hold your breath!

Just to make sure that I understand correctly: this mean that a `backend` has to support (or pretend to support) all the types of steps or it will not compile ? And conversely if you add a step type you have to add (or pretend to add it) to all `backend` ? If I understood well, could you elaborate why doing it this way rather than requiring the `backend` to give a list of steps that it support ? 

Also some naming remarks:

* are `backend` and `translator` the same thing ? If so should we not use only one term ?
* `translator` and `transformer` I think I get the difference but then I don't

adimascio:
> Just to make sure that I understand correctly: this mean that a `backend` has to support (or pretend to support) all the types of steps or it will not compile ? 

More or less, yes. My idea was to make sure all code paths were covered. If a translator doesn't support a transformation step, I had imagined that we could provide a specific `NotImplemented` primitve or something alike but still it would have to be declared explicitly.

> And conversely if you add a step type you have to add (or pretend to add it) to all `backend` ?

Yes, but again I would see this as "explicitly not support it" rather than "pretend to add it".
 
> If I understood well, could you elaborate why doing it this way rather than requiring the `backend` to give a list of steps that it support ?

My insipration came from languages such as OCaml where you would have to make sure you could not leave any case apart.

> Also some naming remarks:
> 
>     * are `backend` and `translator` the same thing ? If so should we not use only one term ?

`backend` here is more a symbolic names (e.g. `mongo3.6`, `pandas`, `mongo4+python`, `sql+python`, etc.). `translator` would be the translator ("compiler" ?) function itself.

>     * `translator` and `transformer` I think I get the difference but then I don't

Bad naming I guess, thanks for pointing this out!

piotch:
Ok thanks. I think it's good to have a way to declare explicitly that a `translator` does not support a `step`. Could it be documented / implemented in this PR ? 

> Yes, but again I would see this as "explicitly not support it" rather than "pretend to add it".

Is there a way to make this explicit in the `translator` API, without having to call the method and catch an exception ? 

I think of a few use cases where it would be useful, given a `backend` to know "in advance" (i.e. before calling the method) if a type of step is supported or not. 

For example : 

* I'm picking a backend, I would like the UI elements which are adding un-supported steps to be disabled, instead of being able to add steps that won't be translated later.
* I'm changing from one backend to another, I'd like to know which backends are available given a list of already created steps, instead of being able to pick any and then it fails when I run the pipeline.

src/lib/translators/mongo.ts
```diff
@@ -0,0 +1,89 @@
...
+      continue;
+    }
+    lastStep = step;
+    outputSteps.push(lastStep);
+  }
+  return outputSteps;
+}
+
+export const translators = {
+  mongo36: pipeToMongo,
```
davinov: Do you anticipate multiple mongo versions here?
Otherwise I think it's easier to have juste one `translators.ts` file that will import all the available translators

davinov: I think that's what you do in the toolchain just below ^^
So I would have called this file `translators`so we can `import { getTranslator } from 'lib/translators'`instead of toolchain.
I would have guessed that "tools/utils" were exported in the toolchain.

src/lib/translators/toolchain.ts
```diff
@@ -0,0 +1,39 @@
...
+/** return a function translating an array of pipeline steps to expected backend
+ * steps
+ *
+ * Example usage:
+ * ```typescript
+ * const translator = getTranslator('mongo36');
+ * const result = translator(myPipeline);
+ * ```
+ */
+export function getTranslator(backend: string) {
```
davinov: I was wondering why was this function better than just exporting the `TRANSLATORS` dictionary.
Is it to also have some safety at runtime?

I feel we either have to choose between being protected at compile time or at runtime.
If we go for the second option, I would have ditched this and ask users to rely on the first one.
Otherwise, we'll have to constantly support two error handling systems, and I fear we'll eventually forget to handle the second one at some point

src/lib/matcher.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Helper exception to handle unreachable conditions in a switch statement.
+ * This will help the typescript compiler to raise an error if one or several
+ * kind of steps are missing in a switch / case statement.
+ *
+ */
+class Unreachable extends Error {
+  constructor(value: never) {
+    super(`value ${value} is not valid`);
+  }
```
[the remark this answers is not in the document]
adimascio: I meant `unreachable` as in [unreachable code](https://en.wikipedia.org/wiki/Unreachable_code).

It really is a way to give a hint to the compiler. While "compiling" the different `swtich / case` branches, typescript will narrow down the possible types (string values here are actually types). If some `cases` are not written down (e.g. `my-step-name`), when typescript will analyze the "unreachable" branch, it will complain about an `my-step-name` being incompatible with the `never` type.

src/lib/matcher.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function matchStep<T>(matcher: StepMatcher<T>): (step: PipelineStep) => T {
+  return (step: PipelineStep): T => {
+    switch (step.name) {
```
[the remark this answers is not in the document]
adimascio: I think that by writing this code, you won't allow typescript to do the "exhaustiveness" test but I can't dig a bit more on that topic.

src/lib/translators/mongo.ts
adimascio: > Do you anticipate multiple mongo versions here?

I think we may have to target `mongo36` and `mongo4` at some point but I'm not certain about this.

src/lib/translators/toolchain.ts
adimascio: > I was wondering why was this function better than just exporting the `TRANSLATORS` dictionary.
> Is it to also have some safety at runtime?

I wanted to define an API rather than a raw object (which admittedly is also an API). This seemed like a reasonable sustainable encapsulation. If at some point we need to perform additional checks, the function will allow this more easily

src/lib/matcher.ts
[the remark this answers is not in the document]
davinov: Ah, I see. You mean the TS analyzer would follow switch/cases branches but not check avery key of an object?

src/lib/matcher.ts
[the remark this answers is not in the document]
davinov: Ah OK! I haven't thought of that type of unreachability.
It still feels weird to have code that will appear on the final bundle to throw errors that we only need at compile time.

Why shouldn't we just define an interface then:
```
interface Matcher {
    domain(): ... 
    filter(): ...
   ...
}
```
That seems more idiomatic and will provide simple messages saying '<method> is not implemented'
Also that means transforming matchers into classes that extends the Matcher interface.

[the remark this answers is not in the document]
davinov: Cool then! :100: Thanks for the explanations

src/lib/matcher.ts
```diff
@@ -0,0 +1,46 @@
...
+ * It exposes 2 main building blocks:
+ * - `StepMatcher`: a type mapping each available pipeline steps to a
+ *   translation function
+ * - `matchStep`: a function accepting a `StepMatcher` and that return a function
+ *   transforming a pipeline step to an arbiratry output.
+ *
+ * This system is supposed to be typesafe, which means that if you add a
+ * new type of `PipelineStep`, the typescript compiler is expected to fail
+ * if your matcher implementation doesn't provide a mapping for this new kind
+ * of step.
```
piotch: is this comment still true ? Reading the doc I understood that the it would automatically detect a new serie of unsuported steps in translators in this case.

src/lib/matcher.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * `StepMatcher` is a type built dynamically according to each available step
+ * name that associates each of this step name to a specific processing function
+ * accepting a step of corresponding type. For instance, it will generate the
+ * `rename: (step: RenameStep) => T` mapping entry automatically.
+ *
+ * This type will help to enforce that each kind of step has a corresponding
+ * translation function at compile time.
+ */
```
piotch: is this still the case?

src/lib/translators/base.ts
```diff
@@ -0,0 +1,192 @@
...
+  custom(step: S.CustomStep) {}
+
+  /**
+   * translates an input pipeline into a list of steps that makes sense for the
+   * targeted backend.
+   *
+   * @param pipeline the array of input steps.
+   * @returns the list of translated output steps.
+   */
+  translate(pipeline: Array<S.PipelineStep>): Array<OutputStep> {
```
piotch: in the case of a SQL translator, it would return a list of strings for example, that the caller would be have to join into a query string ?

adimascio: Probably, yes

src/lib/matcher.ts
adimascio: Yes it is

src/lib/matcher.ts
adimascio: The `StepMatcher` is still the backbone the system relies on.

PR #71, opened 2019-03-29
Temporary :: old 'replace' PR

vdestraitt:
Add a transformation function for the 'replace' step into mongo step

Related to issue: https://github.com/ToucanToco/vue-query-builder/issues/61

### DIFF
--- a/src/lib/pipeline.ts
+++ b/src/lib/pipeline.ts
@@ -4,7 +4,7 @@
  * This module define the mongo → standard pipeline steps implementation.
  */
 
-import { PipelineStep } from './steps';
+import { PipelineStep, ReplaceStep } from './steps';
 
 /**
  * MongoStep interface. For now, it's basically an object with any property.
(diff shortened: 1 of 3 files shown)


src/lib/pipeline.ts
```diff
@@ -175,6 +177,14 @@ function simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<MongoStep> {
 
   for (const step of mongoSteps.slice(1)) {
     if (step.$project !== undefined && lastStep.$project !== undefined) {
+      for (const key in step.$project) {
+        // We do not want to merge two $project with common keys
+        if (key in lastStep.$project) {
```
adimascio:
```suggestion
        if (lastStep.$project.hasOwnProperty(key)) {
```

In our cases, the `in` operator should perfectly do the job but there might be some subtle differences if at some point we get an object extending another one's prototoype. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty

src/lib/pipeline.ts
```diff
@@ -175,6 +177,14 @@ function simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<MongoStep> {
...
+          outputSteps.push(step);
+          lastStep = step;
+          continue
+        }
+      }
```
adimascio: Good catch! I think we should test this case specifically and even probably in a separate commit (or PR).

PR #77, opened 2019-04-11
Fix :: simplify mongo pipeline

vdestraitt:
First we made the test case more coherent to give more sense to the query.

Then we managed 2 cases that were not supported:

- We do not want to merge two Mongo steps with common keys

- We do not want to merge two $project steps or two $addFields steps if the second step includes a reference to a key that was declared (created or modified) in the first step.

Closes https://github.com/ToucanToco/vue-query-builder/issues/76

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -92,7 +92,7 @@ const mapper: StepMatcher<MongoStep> = {
 export class Mongo36Translator extends BaseTranslator {
   translate(pipeline: Array<PipelineStep>) {
     const mongoSteps = super.translate(pipeline).flat();
-    return simplifyMongoPipeline(mongoSteps);
+    return _simplifyMongoPipeline(mongoSteps);
   }
 }
 Object.assign(Mongo36Translator.prototype, mapper);
@@ -107,7 +107,7 @@ Object.assign(Mongo36Translator.prototype, mapper);
  *
  * @returns the list of simplified mongo steps
  */
-function simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<MongoStep> {
+export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<MongoStep> {
   const outputSteps: Array<MongoStep> = [];
   let lastStep: MongoStep = mongoSteps[0];
   outputSteps.push(lastStep);
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -24,6 +24,8 @@ function fromkeys(keys: Array<string>, value = 0) {
 function filterstepToMatchstep(step: FilterStep): MongoStep {
   if (step.operator === undefined || step.operator === 'eq') {
     return { $match: { [step.column]: step.value } };
+  } else if (step.operator === 'ne') {
+    return { $match: { [step.column]: { $ne: step.value } } };
   } else {
     throw new Error(`Operator ${step.operator} is not handled yet.`);
   }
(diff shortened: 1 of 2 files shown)


src/lib/translators/mongo.ts
```diff
@@ -105,23 +105,42 @@ Object.assign(Mongo36Translator.prototype, mapper);
  *
  * @returns the list of simplified mongo steps
  */
-function simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<MongoStep> {
+export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<MongoStep> {
+  let merge: boolean = true;
```
adimascio: In such cases, you should let typescript infer type automatically, you don't have to specify `boolean` explicitly.

```suggestion
  let merge = true;
```

src/lib/translators/mongo.ts
```diff
@@ -105,23 +105,42 @@ Object.assign(Mongo36Translator.prototype, mapper);
...
   for (const step of mongoSteps.slice(1)) {
-    if (step.$project !== undefined && lastStep.$project !== undefined) {
-      // merge $project steps together
-      lastStep.$project = { ...lastStep.$project, ...step.$project };
-      continue;
-    } else if (step.$match !== undefined && lastStep.$match !== undefined) {
-      // merge $match steps together
-      lastStep.$match = { ...lastStep.$match, ...step.$match };
-      continue;
+    const stepOperator: string = Object.keys(step)[0];
```
adimascio: In such cases, you should let typescript infer type automatically, you don't have to specify string explicitly.

```suggestion
    const stepOperator = Object.keys(step)[0];
```

src/lib/translators/mongo.ts
```diff
@@ -105,23 +105,42 @@ Object.assign(Mongo36Translator.prototype, mapper);
...
+    const isMergeable: boolean =
```
adimascio: In such cases, you should let typescript infer type automatically, you don't have to specify boolean explicitly.

src/lib/translators/mongo.ts
adimascio: There's always a single key here right ? In this case, you could write

```suggestion
    const [stepOperator] = Object.keys(step);
```

Otherwise you should not rely on key order.

src/lib/translators/mongo.ts
```diff
@@ -105,23 +105,42 @@ Object.assign(Mongo36Translator.prototype, mapper);
...
+      stepOperator === '$project' || stepOperator === '$addFields' || stepOperator === '$match';
+    if (isMergeable && lastStep[stepOperator] !== undefined) {
+      for (const key in step[stepOperator]) {
+        // We do not want to merge two $project with common keys
+        if (lastStep[stepOperator].hasOwnProperty(key)) {
+          merge = false;
```
adimascio: Shoudn't we `break` once `merge` is set to `false`?

src/lib/translators/mongo.ts
```diff
@@ -105,23 +105,42 @@ Object.assign(Mongo36Translator.prototype, mapper);
...
+        } else if (stepOperator !== '$match') {
+          // We do not want to merge two $project or $addFields with a `step`
+          // key referencing as value a`lastStep` key
+          const valueString: string = JSON.stringify(step[stepOperator][key]);
+          for (const lastKey in lastStep[stepOperator]) {
+            const regex: RegExp = new RegExp(`.*['"]\\$${lastKey}['"].*`);
+            if (regex.test(valueString)) {
+              merge = false;
```
adimascio: Shoudn't we `break` once `merge` is set to `false`?

src/lib/translators/mongo.ts
vdestraitt: @adimascio I tried to use `break` but it seems a little tricky to use as it breaks out from parent loops to. So basically as soon as I met a `merge: false`case, my Mongo pipeline was over as I went out of the root loop (`for (const step of mongoSteps.slice(1))`). 

But open to discussion if I (probably) miss something !

vdestraitt: @adimascio yes right, thx

src/lib/translators/mongo.ts
vdestraitt: @adimascio I tried to use `break` but it seems a little tricky to use as it breaks out from parent loops to. So basically as soon as I met a `merge: false`case, my Mongo pipeline was over as I went out of the root loop (`for (const step of mongoSteps.slice(1))`). 

But open to discussion if I (probably) miss something !

src/lib/translators/mongo.ts
vdestraitt: cf. comment above

vdestraitt:
@adimascio thank you for your review. I took your comments into account in the latest commit

PR #91, opened 2019-04-12
Improve :: `aggregate` step mongo

vdestraitt:
Restrict AggFunctionStep interface to only support basic aggregation functions and improve transformAggregate to support every basic aggregation function among:

- sum
- avg
- count
- min
- max

Related to issue https://github.com/ToucanToco/vue-query-builder/issues/30
But dos not support more complex aggregations such as count_distinct

### DIFF
--- a/src/lib/steps.ts
+++ b/src/lib/steps.ts
@@ -40,7 +40,7 @@ interface AggFunctionStep {
   /** name of the agg function step, (typically the name of the output column) */
   name: string;
   /** the aggregation operation (e.g. `sum` or `count`) */
-  aggfunction: string;
+  aggfunction?: 'sum' | 'avg' | 'count' | 'min' | 'max';
   /** the column the aggregation function is working on */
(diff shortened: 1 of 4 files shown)


src/lib/steps.ts
```diff
@@ -40,7 +40,7 @@ interface AggFunctionStep {
   /** name of the agg function step, (typically the name of the output column) */
   name: string;
   /** the aggregation operation (e.g. `sum` or `count`) */
-  aggfunction: string;
+  aggfunction?: 'sum' | 'avg' | 'count' | 'min' | 'max';
```
adimascio: I don't see why `aggfunction` should be optional and as far as I can see, the `if / else` branch in the `transformAggregate` function assume it's not empty.

vdestraitt: Yes you’re right it should not be optional !! I will fix this thank you

PR #92, opened 2019-04-16
Feature :: `sort` step mongo

vdestraitt:
Support translation of a `sort` step in Mongo Aggregation Pipeline. Sort can be expressed on one or several columns; for any column, sorting order can be either "asc" or "desc".

Closes https://github.com/ToucanToco/vue-query-builder/issues/70

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -89,12 +89,10 @@ function transformReplace(step: ReplaceStep): MongoStep {
 
 /** transform a 'sort' step into corresponding mongo steps */
 function transformSort(step: SortStep): MongoStep {
-  let sortMongo: PropMap<number> = {};
-  if (step.order === undefined) {
-    step.order = Array(step.columns.length).fill('asc');
-  }
+  const sortMongo: PropMap<number> = {};
+  const sortOrders = step.order === undefined ? Array(step.columns.length).fill('asc') : step.order;
   for (let i = 0; i < step.columns.length; i++) {
-    const order = step.order[i] === 'asc' ? 1 : -1;
+    const order = sortOrders[i] === 'asc' ? 1 : -1;
     sortMongo[step.columns[i]] = order;
   }
   return { $sort: sortMongo };
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -1,6 +1,6 @@
 /** This module contains mongo specific translation operations */
 
-import { AggregationStep, FilterStep, PipelineStep, ReplaceStep } from '@/lib/steps';
+import { AggregationStep, FilterStep, PipelineStep, ReplaceStep, SortStep } from '@/lib/steps';
 import { StepMatcher } from '@/lib/matcher';
 import { BaseTranslator } from '@/lib/translators/base';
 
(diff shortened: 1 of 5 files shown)


src/lib/translators/mongo.ts
```diff
@@ -79,6 +79,19 @@ function transformReplace(step: ReplaceStep): MongoStep {
   };
 }
 
+/** transform a 'sort' step into corresponding mongo steps */
+function transformSort(step: SortStep): MongoStep {
+  let sortMongo: PropMap<number> = {};
```
adimascio: You should use `const` instead of `let` here since the `sortMongo` identifier (vs. its value) is not modified in the function.

```suggestion
   const sortMongo: PropMap<number> = {};
```

src/lib/translators/mongo.ts
```diff
@@ -79,6 +79,19 @@ function transformReplace(step: ReplaceStep): MongoStep {
...
+  if (step.order === undefined) {
+    step.order = Array(step.columns.length).fill('asc');
```
adimascio: I think we should not mutate `step` in order to avoid side effects. Maybe should we simply do something like:

```javascript
const sortOrders = step.order === undefined ? Array(step.columns.length).fill('asc')  : step.order;
# ...
```

tests/unit/mongo.spec.ts
```diff
@@ -250,4 +251,89 @@ describe('Pipeline to mongo translator', () => {
...
+  it('can generate a basic sort step on one column', () => {
+    const pipeline: Array<PipelineStep> = [
+      {
+        name: 'sort',
+        columns: ['foo'],
+        order: ['desc'],
+      },
+    ];
+    const querySteps = mongo36translator.translate(pipeline);
+    Mongo36Translator;
```
adimascio: I think this line is an error and should be removed (along with the corresponding `import` statement)

tests/unit/mongo.spec.ts
```diff
@@ -250,4 +251,89 @@ describe('Pipeline to mongo translator', () => {
...
+  it('can generate a sort step on multiple columns', () => {
+    const pipeline: Array<PipelineStep> = [
+      {
+        name: 'sort',
+        columns: ['foo', 'bar'],
+        order: ['asc', 'desc'],
+      },
+    ];
+    const querySteps = mongo36translator.translate(pipeline);
+    Mongo36Translator;
```
adimascio: same as before : I think this line is an error and should be removed (along with the corresponding `import` statement)

tests/unit/mongo.spec.ts
```diff
@@ -250,4 +251,89 @@ describe('Pipeline to mongo translator', () => {
...
+
+  it('can generate a sort step on multiple columns with default order', () => {
+    const pipeline: Array<PipelineStep> = [
+      {
+        name: 'sort',
+        columns: ['foo', 'bar'],
+      },
+    ];
+    const querySteps = mongo36translator.translate(pipeline);
+    Mongo36Translator;
```
adimascio: same as before : I think this line is an error and should be removed (along with the corresponding `import` statement)

tests/unit/mongo.spec.ts
```diff
@@ -250,4 +251,89 @@ describe('Pipeline to mongo translator', () => {
...
+      {
+        $sort: {
+          foo: 1,
+          bar: 1,
+        },
+      },
+    ]);
+  });
+
+  // it('can generate a basic sort step on one column', () => {
```
adimascio: I would suggest either to remove the comment or make the test pass :)

Issue #25, opened 2019-03-20
Implement the 2-panes view


adimascio:
The left-pane will host a graphic rendering of the pipeline. The right-pane (720px min ?) will host the dataset viewer.

PR #93, opened 2019-04-16
Feature :: Resizable Panels Component

Dalanir:
That PR consists in creating a component that has two resizable panels with a story and some basic unit tests.

Linked Issue: #25

### DIFF
--- a/src/components/ResizablePanels.vue
+++ b/src/components/ResizablePanels.vue
@@ -42,7 +42,7 @@ export default class ResizablePanels extends Vue {
     const containerSize = this.$el.getBoundingClientRect();
     const containerSizeOnAxis = containerSize.width;
 
-    const mousemoveListener = (e: any) => {
+    const mousemoveListener = (e: MouseEvent) => {
       this.ratio = this.ratio + e.movementX / containerSizeOnAxis;
       this.isResized = true;
     };
--- a/src/components/ResizablePanels.vue
+++ b/src/components/ResizablePanels.vue
@@ -64,26 +64,5 @@ export default class ResizablePanels extends Vue {
 }
 </script>
 <style lang="scss" scoped>
-  .resizable-panels {
-    display: flex;
-  }
-
-  .resizable-panels__panel {
-    display: flex;
-    height: 100%;
-  }
-
-  .resizable-panels__resizer {
-    display: flex;
-    justify-content: space-between;
-    width: 5px;
-    height: 100%;
-    cursor: ew-resize;
-  }
-
-  .resizable-panels__line {
-    width: 1px;
-    height: 100%;
-    background-color: rgba(0, 0, 0, .3);
-  }
+   @import '../styles/ResizablePanels';
 </style>
--- a/src/components/ResizablePanels.vue
+++ b/src/components/ResizablePanels.vue
@@ -21,7 +21,6 @@ import { Component } from 'vue-property-decorator';
   name: 'resizable-panels',
 })
 export default class ResizablePanels extends Vue {
-  private isResized = false;
   private ratio: number = 0.4;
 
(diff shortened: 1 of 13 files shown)


src/components/ResizablePanels.vue
```diff
@@ -0,0 +1,68 @@
...
+    return {
+      width: `${1 - (this.ratio * 100)}%`,
+    };
+  }
+
+  startResize() {
+    const containerSize = this.$el.getBoundingClientRect();
+    const containerSizeOnAxis = containerSize.width;
+
+    const mousemoveListener = (e: any) => {
```
davinov: `e` should be of type `MouseEvent`?
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

tests/unit/resizable-panels.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+      slots: {
+        'right-panel': '<div class="slot-right-panel"/>'
+      }
+    });
+    const wrapperRightPanelChild = wrapper.find('.slot-right-panel');
+
+    expect(wrapper.exists()).toBeTruthy();
+    expect(wrapperRightPanelChild.exists()).toBeTruthy();
+  });
+
```
davinov: We could test here:
- the default ratio
- the resize behavior?

tests/unit/resizable-panels.spec.ts
```diff
@@ -34,4 +34,59 @@ describe('Resizable Panels', () => {
...
+  describe('when resizing', () => {
+    it('should know that is being resized', () => {
+      const wrapper = shallowMount(ResizablePanels, {
+        attachToDocument: true,
+      });
+      const resizerWrapper = wrapper.find('.resizable-panels__resizer');
+
+      resizerWrapper.trigger('mousedown');
+      wrapper.trigger('mousemove', { movementX: 10 });
+      expect(wrapper.vm.$data.isResized).toBeTruthy();
```
davinov: Isn't `rapper.vm.isResized` sufficient?

I'm also not sure we should test for internal data properties.
What is the effect of this data on the rendered component?
If there is one, I think that's what we should test for. Otherwise, the test can be dropped.

tests/unit/resizable-panels.spec.ts
```diff
@@ -34,4 +34,59 @@ describe('Resizable Panels', () => {
...
+          height: 1000,
+        };
+      };
+      wrapper.vm.$el.getBoundingClientRect = jest.fn(fakeGetBoundingClientRect);
+
+      resizerWrapper.trigger('mousedown');
+      wrapper.trigger('mousemove', { movementX: 100 });
+      // baseRatio: 0.4
+      // If I move my mouse of 100px - that is 1/10 compare to its width - then I increase my ratio by 0.1
+      expect(wrapper.vm.$data.ratio).toEqual(0.5);
```
davinov: :+1: Very clear!

tests/unit/resizable-panels.spec.ts
Dalanir: That was TypeScritpt compiler that enforce that. Indeed, the `vm` object don't know about internal vars.
But you're right, that property isn't used so I remove it from the component.

src/components/ResizablePanels.vue
```diff
@@ -0,0 +1,59 @@
...
+</template>
+<script  lang="ts">
+import Vue from 'vue';
+import { Component } from 'vue-property-decorator';
+
+@Component({
+  name: 'resizable-panels',
+})
+export default class ResizablePanels extends Vue {
+  private ratio: number = 0.4;
```
adimascio: You could let typescript infer the data type.

```suggestion
  private ratio = 0.4;
```

Issue #73, opened 2019-03-31
add code coverage checks


adimascio:
We should have a check on code coverage. By default, we should block pull requests if new code is not fully covered.

PR #94, opened 2019-04-17
Tech :: Add Coverage

Dalanir:
Need express in issue: #73

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    patch: off
+    project: off
+  ignore:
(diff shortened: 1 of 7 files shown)


codecov.yml
```diff
@@ -0,0 +1,6 @@
+coverage:
```
adimascio: Shouldn't we use:

```yaml
coverage:
  status:
    patch:
      default:
        enabled: yes
        target: 100%
```

?

Dalanir: Yes you're right, we should enforce 100% coverage of new code !

PR #95, opened 2019-04-18
Feature: fillna step translation to mongo

vdestraitt:
Replace null values by a given value in a column.

Closes https://github.com/ToucanToco/vue-query-builder/issues/69

### DIFF
--- a/src/lib/steps.ts
+++ b/src/lib/steps.ts
@@ -2,6 +2,8 @@
  * This module defines the supported unit-of-transformation steps.
  */
 
+type primitiveType = number | boolean | string | Date;
+
 export interface DomainStep {
   name: 'domain';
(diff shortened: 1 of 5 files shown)


src/lib/steps.ts
```diff
@@ -72,6 +72,12 @@ export interface SortStep {
   order?: Array<'asc' | 'desc'>;
 }
 
+export interface FillnaStep {
+  name: 'fillna';
+  column: string;
+  value: string | number;
+}
```
adimascio: What about deifning a new type for such cases, like:

```
type primitiveType = number | boolean | string | Date;
```

vdestraitt: Done, thank you :)

PR #97, opened 2019-04-19
Feature :: Top step translation to mongo

vdestraitt:
Return top N rows, by group if `groups` is specified, else over full dataset.

```javascript
{
  name: 'top',
  groups: ['foo']
  value: 'bar',
  sort: 'desc' // or 'asc'
  limit: 10
}
```

Closes https://github.com/ToucanToco/vue-query-builder/issues/66

### DIFF
--- a/doc/steps.md
+++ b/doc/steps.md
@@ -155,3 +155,17 @@ Replace null values by a given value in a column.
     value: "bar"
 }
 ```
+
+### 'top' step
+
+Return top N rows by group if `groups` is specified, else over full dataset
+
+```javascript
+{
+  name: 'top',
+  groups: ['foo']
+  value: 'bar',
+  sort: 'desc' // or 'asc'
+  limit: 10
+}
+```
--- a/doc/steps.md
+++ b/doc/steps.md
@@ -164,7 +164,7 @@ Return top N rows by group if `groups` is specified, else over full dataset.
 {
   name: 'top',
   groups: ['foo'],
-  value: 'bar',
+  rank_on: 'bar',
   sort: 'desc', // or 'asc'
   limit: 10
 }
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -108,7 +108,6 @@ function transformSort(step: SortStep): MongoStep {
 /** transform an 'top' step into corresponding mongo steps */
 function transformTop(step: TopStep): Array<MongoStep> {
   const sortOrder = step.sort === 'asc' ? 1 : -1;
-  const groupMongo: MongoStep = {};
   let groupCols: PropMap<string> | null = {};
 
   // Prepare the $group Mongo step
@@ -119,14 +118,10 @@ function transformTop(step: TopStep): Array<MongoStep> {
   } else {
     groupCols = null;
   }
-  groupMongo['$group'] = {
-    _id: groupCols,
-    _tcAppArray: { $push: '$$ROOT' },
-  };
 
   return [
     { $sort: { [step.value]: sortOrder } },
-    groupMongo,
+    { $group: { _id: groupCols, _tcAppArray: { $push: '$$ROOT' } } },
     { $project: { _tcAppTopElems: { $slice: ['$_tcAppArray', step.limit] } } },
     { $unwind: '$_tcAppTopElems' },
     { $replaceRoot: { newRoot: '$_tcAppTopElems' } },
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -1,6 +1,13 @@
 /** This module contains mongo specific translation operations */
 
-import { AggregationStep, FilterStep, PipelineStep, ReplaceStep, SortStep } from '@/lib/steps';
+import {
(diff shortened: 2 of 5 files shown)


doc/steps.md
```diff
@@ -155,3 +155,17 @@ Replace null values by a given value in a column.
     value: "bar"
 }
 ```
+
+### 'top' step
+
+Return top N rows by group if `groups` is specified, else over full dataset
```
adimascio:
```suggestion
Return top N rows by group if `groups` is specified, else over full dataset.
```

doc/steps.md
```diff
@@ -155,3 +155,17 @@ Replace null values by a given value in a column.
...
+
+```javascript
+{
+  name: 'top',
+  groups: ['foo']
```
adimascio:
```suggestion
  groups: ['foo'],
```

doc/steps.md
```diff
@@ -155,3 +155,17 @@ Replace null values by a given value in a column.
...
+  value: 'bar',
+  sort: 'desc' // or 'asc'
```
adimascio:
```suggestion
  sort: 'desc', // or 'asc'
```

doc/steps.md
```diff
@@ -155,3 +155,17 @@ Replace null values by a given value in a column.
...
+
+### 'top' step
+
+Return top N rows by group if `groups` is specified, else over full dataset
+
+```javascript
+{
+  name: 'top',
+  groups: ['foo']
+  value: 'bar',
```
adimascio: I would suggest `key` or `sortkey` instead of `value` which sounds a bit more "static" to me.

Besides, shouldn't we accept any expression here (e.g. not just a column name)? If so, may an example would be useful.

src/lib/translators/mongo.ts
```diff
@@ -98,6 +105,34 @@ function transformSort(step: SortStep): MongoStep {
...
+    groupCols = null;
+  }
+  groupMongo['$group'] = {
+    _id: groupCols,
+    _tcAppArray: { $push: '$$ROOT' },
+  };
+
+  return [
+    { $sort: { [step.value]: sortOrder } },
+    groupMongo,
```
adimascio: This is purely a matter of style but I wonder if it would not be more readable to avoid an intermediate `groupMongo` variable and let the step be explicit:

```suggestion
    { $group: { _id: groupCols, _tcAppArray: { $push: '$$ROOT' } } },
```

tests/unit/mongo.spec.ts
```diff
@@ -536,4 +536,69 @@ describe('Pipeline to mongo translator', () => {
       { $addFields: { foo: { $cond: [{ $eq: ['$foo', null] }, 'bar', '$foo'] } } },
     ]);
   });
+
+  it('can generate a fillna step', () => {
```
adimascio: This test doesn't seem to belong here :)

doc/steps.md
vdestraitt: @adimascio Hello, not sure to understand your point. `value` receives the name of the column in which to find values to be sorted before returning top rows. What kind of generic expression should we accept ?

Ok to change the parameter name but I would rather suggest something like `values_column` or `column_to_sort` ?

adimascio: I imagined something like `value: "lowercase(my_col)"` (putting aside potential)

vdestraitt: Hum... Sorry but I do not see the use case. We just need to receive a regular and unique column name (I say that based on existing operations that we perform in code mode)

vdestraitt: Basic use case in my thoughts: I want top 10 revenue contributors by country:

```javascript
{
  name: 'top'
  groups: ['country']
  value_column: 'revenue'
}
```

May you please provide an example of yours so I can better understand ?

adimascio: I don't have any precise use case, I just think of the `key` parameter of the python's `sorted` builtin which I find very convenient. I guess a small adapation of your use case would be : "I want the 10 biggest difference in revenue since last year":

```
{
  name: 'top'
  groups: ['country']
  value: 'revenue'
  key: 'abs(revenue - last_year_revenue)'
}
```

Is it too far fetched for our standard use cases?

vdestraitt: Ok so in such a case `value` is not pertinent anymore because you sort only based on `key` right ? I do think it is a bit overkill at least for our first milestones. I think it makes the step more complex to understand and use and it is not something that you want to do often in such a direct fashion. Actually if you want to do this, you may use a `formula` step first to compute your revenue difference, and then apply a `top` step based on the new computed column. So we are not stuck for such a use case, and I believe that in terms of data pipeline it is easier and clearer to do it in 2 steps.

I do agree on the convenience offered by powerful features in programmatic tools such as python, but I believe that for the purpose of our graphical interface we have to keep our transformation steps as straightforward and pedagogic as possible, while promoting a modular rather than parametrical  approach to data wrangling. :) Let me know if you still disagree or if I miss something.

vdestraitt: @adimascio would `rank_values_on` be a better candidate to replace the parameter `value`?

Issue #27, opened 2019-03-20
[table] select a column


adimascio:
It must be possible to select (and highlight) a column in the table. It should also be possible to select (and highlight) more than one column at once with `ctrl+click`

PR #104, opened 2019-04-24
Feature :: Simple data viewer 

### LINKED ISSUE #28: [table] display the data
In the right pane, we need a table that displays the dataset. The table should be scrollable horizontally.

Dalanir:
A simple first iteration to create a component that display data. There are the basis to column selection (just change the UI right now)

Some pics
Capture d’écran 2019-04-26 à 10 29 35
Capture d’écran 2019-04-26 à 10 29 41
Capture d’écran 2019-04-26 à 10 29 48

closes #28 and #27


Dalanir:
[ETQC, je visualise le tableau de données](https://trello.com/c/ITgAZFcp/3188-etqc-je-visualise-le-tableau-de-donn%C3%A9es)

tests/unit/data-viewer-cell.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { shallowMount } from '@vue/test-utils';
+import DataViewerCell from '../../src/components/DataViewerCell.vue';
```
adimascio:
```suggestion
import DataViewerCell from '@/components/DataViewerCell.vue';
```

src/components/DataViewer.vue
```diff
@@ -0,0 +1,92 @@
...
+    DataViewerCell,
+  },
+})
+export default class DataViewer extends Vue {
+  @Prop({
+    default: () => [],
+    type: Array,
+  })
+  dataset!: Array<object>;
+
```
adimascio: let's remove the empty line :)

```suggestion
```

src/components/DataViewer.vue
```diff
@@ -0,0 +1,92 @@
...
+  selectedColumns: Array<string> = [];
+
+  get datasetIsntEmpty() {
+    return this.dataset.length !== 0;
+  }
+
+  get columnNames() {
+    return _.keys(this.dataset[0]);
```
adimascio: I suggest we use standard JS methods when possible:

```suggestion
     return Object.keys(this.dataset[0]);
```

src/components/DataViewer.vue
```diff
@@ -0,0 +1,92 @@
...
+      return {
+        name: d,
+        class: {
+          'data-viewer__header-cell': true,
+          'data-viewer__header-cell--active': this.isSelected(d),
+        },
+      };
+    });
+  }
+
```
adimascio: For the same reason as above, I would suggest:

```
  get formattedColumns() {
    return this.columnNames.map(d => {
      return {
        name: d,
        class: {
          'data-viewer__header-cell': true,
          'data-viewer__header-cell--active': this.isSelected(d),
        },
      };
    });
  }
```

src/components/DataViewer.vue
```diff
@@ -0,0 +1,92 @@
...
+  isSelected(column: string) {
+    return _.includes(this.selectedColumns, column);
```
adimascio: For the same reason as above:

```suggestion
    return this.selectedColumns.includes(column);
```

src/components/DataViewer.vue
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  getValue(row: DataRow, column: string) {
+    return row[column];
+  }
+
+  toggleColumnSelection(column: string) {
+    if (_.includes(this.selectedColumns, column)) {
```
adimascio:
```suggestion
     if (this.selectedColumns.includes(column)) {
```

src/components/DataViewer.vue
```diff
@@ -0,0 +1,92 @@
...
+      this.selectedColumns = _.without(this.selectedColumns, column);
+    } else {
+      this.selectedColumns = _.concat(this.selectedColumns, column);
```
adimascio:
```suggestion
      this.selectedColumns = [...this.selectedColumns, column];
```

src/components/DataViewer.vue
Dalanir: Hmm not sur, available to discuss this convention

Dalanir: Discuss with @adimasci : we separate here `props` from intern `data` for readability

src/components/DataViewer.vue
```diff
@@ -0,0 +1,126 @@
...
+   * @param {DataRow} row - A row from our dataset
+   * @param {string} column - A column name
+   * @return {any}
+   */
+  getValue(row: DataRow, column: string) {
+    return row[column];
+  }
+
+  /**
+   * Set or unset a column name from selectedColumns
```
davinov: Would it be more accurately "select or deselect a column by its name"?
(I don't understand what's to set set a name, outside a renaming operation)

src/components/DataViewerCell.vue
```diff
@@ -0,0 +1,32 @@
...
+    :class="{'data-viewer-cell': true, 'data-viewer-cell--active': props.isSelected }"
+  >{{ props.value }}</td>
+</template>
+<script lang="ts">
+import Vue from 'vue';
+import { Component, Prop } from 'vue-property-decorator';
+
+/**
+ * @name DataViewerCell
+ * @description A cell display in a DataViewer
```
davinov: +ed

src/components/DataViewerCell.vue
```diff
@@ -0,0 +1,32 @@
...
+export default class DataViewerCell extends Vue {
+  @Prop({
+    default: () => false,
+  })
+  isSelected!: boolean;
+
+  @Prop({
+    default: () => '',
+  })
+  value!: string;
```
davinov: Is this always a string?
If values are not formatted, this can be pretty much anything like Number, null/undefined or even Object (for _id fields sent by Mongo)

src/styles/DataViewer.scss
```diff
@@ -0,0 +1,65 @@
...
+  flex-shrink: 0;
+  justify-content: space-between;
+}
+
+.data-viewer-cell {
+  background-color: #fafafa;
+}
+
+.data-viewer-cell--active {
+  // It's trick to have its left side colored cause of border-collapse
```
davinov: :+1:

stories/data-viewer.js
```diff
@@ -0,0 +1,38 @@
+import { storiesOf } from '@storybook/vue';
+
+import { DataViewer } from '../dist/storybook/components';
+
+const stories = storiesOf('DataViewer', module);
+
+stories.add('empty', () => ({
+  components: { DataViewer },
+  template: `
+    <div style="width: 800px; height: 800px">
```
davinov: You don't need to set up a size like this in storybook 5. There is a button in the workspace that let you resize and rotate the viewport. Therefore it's better to let the component take the space available.

tests/unit/data-viewer-cell.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      context: {
+        props: {
+          value: 'my_value',
+          isSelected: true,
+        },
+      },
+    });
+
+    expect(wrapper.classes()).toContain('data-viewer-cell--active');
+  });
```
davinov: I would add a test for undefined value and for some other value types, so it's clearly specified how they are stringified.

src/components/DataViewerCell.vue
```diff
@@ -0,0 +1,32 @@
...
+import { Component, Prop } from 'vue-property-decorator';
+
+/**
+ * @name DataViewerCell
+ * @description A cell display in a DataViewer
+ * @param {boolean} isSelected - Wether the cell is selected or is in a selected column
+ * @param {string} value - The cell's value
+ */
+@Component({
+  name: 'data-viewer-cell',
```
davinov: I think you should make this component functional as:
- it has no state
- the number of cells can be quite high, and the memory I'm of so many components is not to be neglected
- there is no real interest seeing this component in vue dev tools

tests/unit/data-viewer.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+  });
+
+  it('should display a message when no data', () => {
+    const wrapper = shallowMount(DataViewer, {
+      propsData: {
+        dataset: [],
+      },
+    });
+
+    expect(wrapper.text()).toEqual('No Data Available');
```
davinov: No capitalization needed in this message I think

tests/unit/data-viewer.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      });
+
+      const firstRowWrapper = wrapper.find('.data-viewer__row');
+      const firstCellWrapper = firstRowWrapper.find('dataviewercell-stub');
+      // Syntax specific to stubbed functional Vue Component
+      expect(firstCellWrapper.attributes('value')).toEqual('value1');
+    });
+  });
+
+  describe('when clicking on first head cell', () => {
```
davinov: This is more of a `context` (starts with when)
Otherwise it's describing the row selection behavior (`describe('row selection)`)

tests/unit/data-viewer.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+        },
+      });
+
+      const headerCellsWrapper = wrapper.findAll('.data-viewer__header-cell');
+      expect(headerCellsWrapper.at(0).text()).toEqual('columnA');
+      expect(headerCellsWrapper.at(1).text()).toEqual('columnB');
+      expect(headerCellsWrapper.at(2).text()).toEqual('columnC');
+    });
+
+    describe('when clicking', () => {
```
davinov: Cell selection?

src/components/DataViewer.vue
Dalanir: 👍

src/components/DataViewerCell.vue
Dalanir: I choose to consider it like as it's "just" the display in the table. It seems that specifying string here cast automatically cf. fake_dataset. But open to discussion

src/components/DataViewerCell.vue
Dalanir: It is, the declaration is done on the template

tests/unit/data-viewer.spec.ts
Dalanir: Jest doesn't provide `context` keyword like mocha :/

Dalanir: So ok to rename

tests/unit/data-viewer.spec.ts
Dalanir: Why ?

davinov: https://www.grammarbook.com/punctuation/capital.asp

src/components/DataViewerCell.vue
davinov: I don't think adding a type in TS does an automatic casting.
But using `{{ value }}` in the template, I think vue does the casting automatically.
Still, the TS compiler will fail if we pass a dataset which have different value types.

PR #111, opened 2019-04-25
Fix :: simplifyMongoPipeline

### LINKED ISSUE #114: simplifyMongoPipeline :: don't mix included and excluded fields
Mongo won't support mixing inclusions (e.g. `mycol: 1`) and exclusions (e.g. `mycol`: 0) in a single `$project` step. We have to make sure `simplifyMongoPipeline` doesn't generate such mixed steps.

vdestraitt:
Do not allow mix of inclusions ("\<column\>: <any expr != 0>") and exclusions ("\<column\>: 0") when trying to merge $project steps, as it source of error in Mongo

Closes #114

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -233,8 +233,8 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
          * includes any inclusion, we do not want to merge those steps.
          */
         if (stepOperator === '$project') {
-          const excluded = Number(step['$project'][key] === 0);
-          merge = !Object.values(lastStep.$project).some(value => value === excluded);
+          const included = Boolean(step['$project'][key]);
+          merge = Object.values(lastStep.$project).every(value => value == included);
         }
         if (lastStep[stepOperator].hasOwnProperty(key)) {
           // We do not want to merge two $project with common keys
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -233,8 +233,8 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
          * includes any inclusion, we do not want to merge those steps.
          */
         if (stepOperator === '$project') {
-          const included = Boolean(step['$project'][key]);
-          merge = Object.values(lastStep.$project).every(value => value == included);
+          const included = Boolean(step.$project[key]);
+          merge = Object.values(lastStep.$project).every(value => Boolean(value) === included);
         }
         if (lastStep[stepOperator].hasOwnProperty(key)) {
           // We do not want to merge two $project with common keys
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -229,17 +229,12 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
       for (const key in step[stepOperator]) {
         /**
          * In Mongo, exclusions cannot be combined with any inclusion, so if we
-         * have such an exclusion in a $project step, and that the previous one
-         * include any inclusion, we do not want to merge the steps.
+         * have an exclusion in a $project step, and that the previous one
+         * includes any inclusion, we do not want to merge those steps.
          */
         if (stepOperator === '$project') {
-          const bool = step[stepOperator][key] === 0;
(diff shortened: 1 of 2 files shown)


src/lib/translators/mongo.ts
```diff
@@ -227,11 +227,26 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
...
       for (const key in step[stepOperator]) {
-        // We do not want to merge two $project with common keys
+        /**
+         * In Mongo, exclusions cannot be combined with any inclusion, so if we
+         * have such an exclusion in a $project step, and that the previous one
+         * include any inclusion, we do not want to merge the steps.
+         */
+        if (stepOperator === '$project') {
+          const bool = step[stepOperator][key] === 0;
+          for (const lastKey in lastStep[stepOperator]) {
```
adimascio: Maybe this *for-loop* could be replaced by a call to `some`? I have something like this in mind:

```javascript
  const excluded = Number(step.$project[key] === 0);
  const merge = !Object.values(lastStep.$project).some(value => value === excluded);
```

Also, I have the feeling that this would even be enough:

```javascript
  const merge = !Object.values(lastStep.$project).some(value => value === step.$project[key]);
```

What do you think?

src/lib/translators/mongo.ts
```diff
@@ -227,11 +227,26 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
...
     if (isMergeable && lastStep[stepOperator] !== undefined) {
       for (const key in step[stepOperator]) {
-        // We do not want to merge two $project with common keys
+        /**
+         * In Mongo, exclusions cannot be combined with any inclusion, so if we
+         * have such an exclusion in a $project step, and that the previous one
+         * include any inclusion, we do not want to merge the steps.
+         */
+        if (stepOperator === '$project') {
+          const bool = step[stepOperator][key] === 0;
```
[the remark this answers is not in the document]
vdestraitt: Agreed !

src/lib/translators/mongo.ts
```diff
@@ -227,11 +227,26 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
...
+          for (const lastKey in lastStep[stepOperator]) {
+            if (lastStep[stepOperator][lastKey] === Number(bool)) {
```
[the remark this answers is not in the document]
vdestraitt: Yes it would, and it is :)

src/lib/translators/mongo.ts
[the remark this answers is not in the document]
vdestraitt: Better naming indeed

src/lib/translators/mongo.ts
vdestraitt: Thank you for the `some` tip ! Much nicer and clearer.

Your second proposal does not test what we want.
An exclusion matches the pattern `column: 0` and an inclusion matches the pattern `column: \<something strictly different than 0\>`
We do not want two merge a step that includes an inclusion with another including at least one exclusion.

That being said, my tests are not enough => see you in the next commits ;)

src/lib/translators/mongo.ts
```diff
@@ -233,8 +233,8 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
          * includes any inclusion, we do not want to merge those steps.
          */
         if (stepOperator === '$project') {
-          const excluded = Number(step['$project'][key] === 0);
-          merge = !Object.values(lastStep.$project).some(value => value === excluded);
+          const included = Boolean(step['$project'][key]);
```
adimascio:
```suggestion
          const included = Boolean(step.$project[key]);
```

src/lib/translators/mongo.ts
```diff
@@ -233,8 +233,8 @@ export function _simplifyMongoPipeline(mongoSteps: Array<MongoStep>): Array<Mong
...
+          merge = Object.values(lastStep.$project).every(value => value == included);
```
adimascio: I'm afraid you can't rely on `==` here. `1 == true` but `2 != true` for instance. You're probably stuck with converting both sides to booleans and use `===`. Or maybe use the `xor` operator:

```
          merge = Object.values(lastStep.$project).every(value => !(value ^ included));
```

PR #118, opened 2019-04-30
Feature :: unpivot step translation to mongo

vdestraitt:
Unpivot a list of columns to rows.

 ```javascript
{
  name: 'unpivot',
  keep: ['COMPANY', 'COUNTRY'], // columns to keep fixed around which to unpivot columns
  unpivot: ['NB_CLIENTS', 'REVENUES'], // columns to unpivot
  unpivot_column_name: 'KPI', // name of the new dimension column created after unpivot
  value_column_name: 'VALUE', // name of the new value column created after unpivot
  dropna: true // whether null values have to be kept or the corresponding rows discarded
}

Closes https://github.com/ToucanToco/vue-query-builder/issues/60

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -6,6 +6,7 @@ import {
   ArgminStep,
   FilterStep,
   PipelineStep,
+  PivotStep,
   ReplaceStep,
   SortStep,
   TopStep,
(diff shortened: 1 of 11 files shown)


vdestraitt:
[ETQC, je suis capable de traduire une step `unpivot` en mongo](https://trello.com/c/AyxOXsga/3211-etqc-je-suis-capable-de-traduire-une-step-unpivot-en-mongo)

src/lib/translators/mongo.ts
```diff
@@ -222,6 +223,43 @@ function transformTop(step: TopStep): Array<MongoStep> {
   ];
 }
 
+/** transform an 'unpivot' step into corresponding mongo steps */
+function transformUnpivot(step: UnpivotStep): Array<MongoStep> {
+  let mongoPipeline: Array<MongoStep> = [];
+  let projectCols: PropMap<string> = {};
```
adimascio: I think we could use `const` here:

```suggestion
  const projectCols: PropMap<string> = {};
```

src/lib/translators/mongo.ts
```diff
@@ -222,6 +223,43 @@ function transformTop(step: TopStep): Array<MongoStep> {
...
+  let objectToArray: PropMap<string> = {};
```
adimascio: I think we could use `const` here:

```suggestion
  const objectToArray: PropMap<string> = {};
```

src/lib/translators/mongo.ts
```diff
@@ -222,6 +223,43 @@ function transformTop(step: TopStep): Array<MongoStep> {
...
+
+  // Prepare the projectCols to be included in Mongo $project steps
+  for (const col of step.keep) {
```
adimascio: Since we now depend on lodash, I suggest the following:

```typescript
import _ from 'lodash';

const projectCols: PropMap<string> = _.fromPairs(step.keep.map(col => [col, `$${col}`]));
```

src/lib/translators/mongo.ts
```diff
@@ -222,6 +223,43 @@ function transformTop(step: TopStep): Array<MongoStep> {
...
+    projectCols[col] = `$${col}`;
+  }
+
+  // Prepare the objectToArray object to be included in the first Mongo $project step
+  for (const col of step.unpivot) {
```
adimascio: Same comment as above:

```typescript
import _ from 'lodash';

const objectToArray: PropMap<string> = _.fromPairs(step.unpicot.map(col => [col, `$${col}`]));
```

I even wonder if this specific oneliner should not be in a utils function since we do this kind of mapping (`col` -> `$col`) quite often.

adimascio: I have a small ongoing PR about that.

PR #119, opened 2019-04-30
Improve :: replace step translation to mongo

vdestraitt:
We have to support multiple replacements at once.

API:

```javascript
{
   name: 'replace',
   search_column: "COUNTRY",
   new_column: "REGION"
   to_replace: [
     ['France', 'Europe'],
     ['USA', 'North America'],
  ]
}
```

Closes https://github.com/ToucanToco/vue-query-builder/issues/115

### DIFF
--- a/doc/steps.md
+++ b/doc/steps.md
@@ -362,7 +362,10 @@ Rename a column.,
 ### `replace` step
 
 Replace one or several values in a column, and write resulting column inplace or in a new column.
+<<<<<<< HEAD
+=======
 Replaced values must be strings, so first cast your `search_column`first if needed.
+>>>>>>> 56647b1fb702c5b6f0f6a943dd628ecef57a408e
 
 A replace step has the following strucure:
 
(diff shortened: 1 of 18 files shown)


src/lib/translators/mongo.ts
```diff
@@ -173,16 +172,14 @@ function transformPercentage(step: PercentageStep): Array<MongoStep> {
 
 /** transform an 'replace' step into corresponding mongo steps */
 function transformReplace(step: ReplaceStep): MongoStep {
+  const branches: Array<MongoStep> = [];
+  for (const k in step.to_replace) {
```
adimascio: I think this could be written as:

```typescript
  const branches: Array<MongoStep> = [];
  for (const [oldval, newval] of Object.entries(step.to_replace)) {
    branches.push({ case: { $eq: [`$${step.search_column}`, oldval] }, then: newval });
  }
```

or directly with a `map`:

```typescript
    const branches: Array<MongoStep> = Object.entries(step.to_replace).map((oldval, newval) => (
        { case: { $eq: [`$${step.search_column}`, oldval] }, then: newval }
    );
```

doc/steps.md
```diff
@@ -309,20 +309,78 @@ Rename a column.,
 
 ### `replace` step
 
-Replace value in a column, and write resulting column inplace or in a new column.
+Replace one or several values in a column, and write resulting column inplace or in a new column.
+Replaced values must be strings, so first cast your `search_column`first if needed.
```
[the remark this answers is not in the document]
vdestraitt: @adimasci I agree that it is not optimal to force value to be strings... But as the proposed API use an object, keys in `to_replace` need to be strings right ? Maybe the API is not fine then ?

src/lib/translators/mongo.ts
vdestraitt: Nice ! Thank you for the suggestion :)

vdestraitt: Nice ! Thank you for the suggestion :)

doc/steps.md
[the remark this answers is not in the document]
adimascio: As discussed earlier, I think we should use a list of lists:

```
{
  to_replace: [
    ['old1', 'old2'],
    [0, 1],
  ]
```

Furthermore, we'll probably be glad to have a predictible order for replacements at some point.

src/lib/steps.ts
```diff
@@ -3,6 +3,7 @@
  */
 
 type PrimitiveType = number | boolean | string | Date;
+type PropAny = { [prop: string]: any };
```
adimascio: This line is not necessary, you should remove it.

```suggestion
```

src/lib/translators/mongo.ts
```diff
@@ -231,16 +231,14 @@ function transformPivot(step: PivotStep): Array<MongoStep> {
 
 /** transform an 'replace' step into corresponding mongo steps */
 function transformReplace(step: ReplaceStep): MongoStep {
+  const branches: Array<MongoStep> = [];
```
adimascio: I suggest destructuring `rep` into `[oldval, newval]` and use `map`

```typescript
const branches = step.to_replace.map(([oldval, newval]) => ({
    case: { $eq: [$$(step.search_column), oldval] },
    then: newval,
}));
```

vdestraitt: Thank you for the suggestion !

PR #120, opened 2019-04-30
feature :: define a basic dataset type

adimascio:
DataSet type is defined as two properties:

- `headers`: a list of objects defining columns
  - each column is itself defined by a `name` and
    an optional `type`. We'll probably add extra
    metadata later,

- `data`: a list of list of values

Use this type in the `DataViewer` component.

Related to https://github.com/ToucanToco/vue-query-builder/issues/26

### DIFF
--- a/src/lib/dataset.ts
+++ b/src/lib/dataset.ts
@@ -8,7 +8,7 @@ export type DataSetColumn = {
 };
 
 export type DataSet = {
-  columns: Array<DataSetColumn>;
+  headers: Array<DataSetColumn>;
   data: Array<Array<any>>;
 };
 
@@ -31,15 +31,15 @@ function setUnion<T>(set1: Set<T>, set2: Set<T>): Set<T> {
  * a `DataSet` structure
  */
 export function mongoResultsToDataset(results: MongoResults): DataSet {
-  const dataset: DataSet = { columns: [], data: [] };
+  const dataset: DataSet = { headers: [], data: [] };
   if (results.length) {
     // each document migh have a different set of keys therefore we need
     // to loop over all documents and make the union of all keys
     const colnames = results.map(row => new Set(Object.keys(row))).reduce(setUnion, new Set());
     // transform set of names to list of DataSetColumn objects
-    dataset.columns = Array.from(colnames).map(name => ({ name }));
+    dataset.headers = Array.from(colnames).map(name => ({ name }));
     for (const row of results) {
-      dataset.data.push(dataset.columns.map(coldef => row[coldef.name]));
+      dataset.data.push(dataset.headers.map(coldef => row[coldef.name]));
     }
   }
   return dataset;
--- a/src/lib/dataset.ts
+++ b/src/lib/dataset.ts
@@ -1,0 +1,48 @@
+/**
+ * This module contains helpers and definitions related to datasets.
+ */
+
+export type DataSetColumn = {
+  name: string;
+  type?: 'integer' | 'float' | 'boolean' | 'string' | 'date' | 'object';
+};
+
+export type DataSet = {
+  columns: Array<DataSetColumn>;
+  data: Array<Array<any>>;
+};
+
+export type MongoDocument = { [k: string]: any };
+export type MongoResults = Array<MongoDocument>;
+
+/**
+ * implement set union
+ * > setUnion(new Set([1, 2, 3]), new Set([2, 4, 6]))
+ * Set([1, 2, 3, 4, 6])
+ */
+function setUnion<T>(set1: Set<T>, set2: Set<T>): Set<T> {
+  const result = new Set(set1);
+  set2.forEach(val => result.add(val));
+  return result;
+}
+
+/**
+ * transform a mongo resultset (i.e. a list of json documents) into
+ * a `DataSet` structure
+ */
+export function mongoResultsToDataset(results: MongoResults): DataSet {
+  const dataset: DataSet = { columns: [], data: [] };
+  if (results.length) {
+    // each document migh have a different set of keys therefore we need
+    // to loop over all documents and make the union of all keys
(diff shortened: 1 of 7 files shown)


src/lib/dataset.ts
```diff
@@ -0,0 +1,48 @@
...
+  const result = new Set(set1);
+  set2.forEach(val => result.add(val));
+  return result;
+}
+
+/**
+ * transform a mongo resultset (i.e. a list of json documents) into
+ * a `DataSet` structure
+ */
+export function mongoResultsToDataset(results: MongoResults): DataSet {
```
vdestraitt: In terms of architecture, do we want to mimic a principle similar to our translators ? With a  file defining the generic dataset API and a mongo.ts file for the specifics of how to get to it through a mongo backend ?

src/lib/dataset.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Set([1, 2, 3, 4, 6])
+ */
+function setUnion<T>(set1: Set<T>, set2: Set<T>): Set<T> {
+  const result = new Set(set1);
+  set2.forEach(val => result.add(val));
+  return result;
+}
+
+/**
+ * transform a mongo resultset (i.e. a list of json documents) into
```
vdestraitt:
```suggestion
 * transform a mongo results set (i.e. a list of json documents) into
```

stories/data-viewer.js
```diff
@@ -1,29 +1,37 @@
...
 
-import { DataViewer } from '../dist/storybook/components';
+import {DataViewer} from '../dist/storybook/components';
 
 const stories = storiesOf('DataViewer', module);
 
 stories.add('empty', () => ({
-  components: { DataViewer },
-  template: `
+                       components: {DataViewer},
```
vdestraitt: Bad indentation ?

tests/unit/dataset.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      { col1: 'bar', col2: 7, col4: '?' },
+    ];
+    const dataset = mongoResultsToDataset(mongoResults);
+    expect(dataset.columns).toEqual([
+      { name: 'col1' },
+      { name: 'col2' },
+      { name: 'col3' },
+      { name: 'col4' },
+    ]);
+    expect(dataset.data).toEqual([['foo', undefined, true, undefined], ['bar', 7, undefined, '?']]);
```
vdestraitt: Ok nice one: so we are guaranteed that all rows will always be arrays of same length, with null values represented as `undefined`=> maybe we can document that clearly in the function implemented in dataset.ts ?

src/lib/dataset.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+export type DataSetColumn = {
+  name: string;
+  type?: 'integer' | 'float' | 'boolean' | 'string' | 'date' | 'object';
+};
+
+export type DataSet = {
+  columns: Array<DataSetColumn>;
+  data: Array<Array<any>>;
```
vdestraitt: Maybe an example dataset would be nice here. We have to understand at once what is the orientation of data (it is an arrays of rows, not columns)

src/lib/dataset.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * transform a mongo resultset (i.e. a list of json documents) into
+ * a `DataSet` structure
+ */
+export function mongoResultsToDataset(results: MongoResults): DataSet {
+  const dataset: DataSet = { columns: [], data: [] };
+  if (results.length) {
+    // each document migh have a different set of keys therefore we need
+    // to loop over all documents and make the union of all keys
+    const colnames = results.map(row => new Set(Object.keys(row))).reduce(setUnion, new Set());
```
vdestraitt: @adimasci Do we want to try to get this meta information directly through an API call to the Mongo backend ? Thus avoiding to scan the full dataset ?

src/lib/dataset.ts
```diff
@@ -0,0 +1,48 @@
...
+    // transform set of names to list of DataSetColumn objects
+    dataset.columns = Array.from(colnames)
+      .sort()
```
vdestraitt: @adimasci so we always get columns in alphabetical order ? Is this necessary ?

adimascio: This is not strictly necessary but there is no good order here and it's easier to test.

src/lib/dataset.ts
adimascio: I would vote a manual scan for now and keep in mind we might want to use yh backend capabilities to get this information directly.

tests/unit/dataset.spec.ts
adimascio: Indeed

src/lib/dataset.ts
adimascio: That was intended, but OK :)

src/lib/dataset.ts
adimascio: Let's do that

src/lib/dataset.ts
vdestraitt: agreed :)

src/lib/dataset.ts
vdestraitt: I understand, but it may be disturbing at rendering, for exemple if I rename the column it will reappear elsewhere. It is an additional effort for the user to keep track of it.

vdestraitt: Just another comment: maybe we may use the term `headers` rather than `columns`, as a column is actually, by definition, a combination of `headers` and associated `data`

tests/unit/dataset.spec.ts
adimascio: I've eventually decided to go for `null` since I don't see the point of differentiating `null` values (i.e. specified but null) and `undefined` values.

PR #130, opened 2019-05-12
feature :: use vuex for application state

### LINKED ISSUE #131: Use vuex to manage application state

adimascio:
Closes #131

### DIFF
--- a/playground/App.vue
+++ b/playground/App.vue
@@ -1,13 +1,8 @@
 <template>
   <div id="app">
     <ResizablePanels>
-      <Pipeline
-        slot="left-panel"
-        :steps="steps"
-        :domains-list="domainsList"
-        @selectedPipeline="setSteps"
-      />
-      <DataViewer slot="right-panel" :dataset="dataset"/>
+      <Pipeline slot="left-panel"/>
+      <DataViewer slot="right-panel"/>
     </ResizablePanels>
     <pre><code>{{ code }}</code></pre>
   </div>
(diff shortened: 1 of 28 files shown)


playground/App.vue
```diff
@@ -1,70 +1,50 @@
...
 export default class App extends Vue {
-  dataset: DataSet = {
-    headers: [{ name: 'col1' }, { name: 'col2' }, { name: 'col3' }],
-    data: [
-      [12, 13, 14],
-      [15, 19, 15],
-      [16, 20, 16],
-    ]
-  };
+  @State('pipeline') steps!: Pipeline;
```
vdestraitt: Should we remain consistent and keep the `pipeline`wording rather than `steps`?

tests/unit/pipeline.spec.ts
```diff
@@ -1,28 +1,36 @@
...
   it('renders steps', () => {
-    const wrapper = shallowMount(Pipeline, {
-      propsData: {
-        steps: [{ name: 'domain' }, { name: 'rename' }, { name: 'group' }],
-      },
-    });
+    const pipeline: Pipeline = [
+      { name: 'domain', domain: 'GoT' },
+      { name: 'replace', search_column: 'characters', to_replace: [['Snow', 'Targaryen']] },
+      { name: 'sort', columns: ['death'] },
```
Dalanir: Like this test aha !

stories/data-viewer.js
```diff
@@ -1,44 +1,41 @@
...
+        ],
+      data: [
+        ['value1', 'value2', 'value3'],
+        ['value4', 'value5', 'value6'],
+        ['value7', 'value8', 'value9'],
+        ['value10', 'value11', 'value12'],
+        ['value10', { obj: 'value14' }, null],
+      ]
+    },
+  }, [], false),
```
Dalanir: Why is autobind `false` and why specify it ? It looks like it's the default value

adimascio: good catch, it's a mistake.

PR #136, opened 2019-05-20
Feature :: Popover Component

Dalanir:


src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+
+  @Prop({
+    type: Boolean,
+    default: () => false,
+  })
+  active!: boolean;
+
+  @Prop({
+    default: () => POPOVER_ALIGN.CENTER,
+    validator: value => _.includes(POPOVER_ALIGN, value),
```
adimascio:
```suggestion
     validator: value => POPOVER_ALIGN.includes(value),
```

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+
+  @Prop({
+    type: Boolean,
+    default: () => false,
+  })
+  bottom!: boolean;
+
+  elementStyle: ElementPosition = {};
+  parent: HTMLElement | null = null;
+  parents: Array<HTMLElement> = [];
```
adimascio:
```suggestion
   parents: HTMLElement[] = [];
```

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+
+  @Prop({
+    default: () => POPOVER_ALIGN.CENTER,
+    validator: value => _.includes(POPOVER_ALIGN, value),
+  })
+  align!: string;
+
+  @Prop({
+    type: Boolean,
+    default: () => false,
```
adimascio:
```suggestion
    default: false,
```

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+
+    // Move the popover into the body to prevent its hiding by an
+    // `overflow: hidden` container
+    document.body.appendChild(this.$el);
+
+    // Get all scrollable parents
+    const parents = [];
+    let { parent } = this;
+
+    while (parent !== document.body && !_.isNull(parent)) {
```
adimascio: Any reason not to do:
```suggestion
    while (parent !== document.body && parent !== null) {
```

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+
+  beforeDestroy() {
+    // Cleanup listeners
+    window.removeEventListener('orientationchange', this.orientationchangeListener);
+    window.removeEventListener('resize', this.resizeListener);
+    _.forEach(this.parents, parent => {
+      return parent.removeEventListener('scroll', this.scrollListener);
+    });
+    // Cleanup DOM
+    if (!_.isNull(this.$el.parentElement)) {
```
adimascio:
```suggestion
    if (this.$el.parentElement !== null) {
```

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+    if (isActive) {
+      return this.updatePosition();
+    }
+  }
+
+  beforeDestroy() {
+    // Cleanup listeners
+    window.removeEventListener('orientationchange', this.orientationchangeListener);
+    window.removeEventListener('resize', this.resizeListener);
+    _.forEach(this.parents, parent => {
```
adimascio: I would suggest

```javascript
    for (const p of this.parents) {
        parent.removeEventListener('scroll', this.scrollListener);
    }
```

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+      case POPOVER_ALIGN.LEFT:
+        return this.alignLeft;
+      case POPOVER_ALIGN.RIGHT:
+        return this.alignRight;
+      default:
+        return () => ({});
+    }
+  }
+
+  get isActive() {
```
adimascio: Why do we need a getter if it just returns a plain instance attribute?

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+    return this.active;
+  }
+
+  get isBottom() {
```
adimascio: same here, why do we need a getter if it just returns a plain instance attribute?

tests/unit/popover.spec.ts
```diff
@@ -0,0 +1,671 @@
...
+          return createElement(
+            'div',
+            {
+              style: parentStyle,
+            },
+            [
+              createElement(
+                Popover,
+                {
+                  props,
```
adimascio:
```suggestion
                  ...props,
```
maybe?

src/components/Popover.vue
```diff
@@ -0,0 +1,289 @@
...
+      parents.push(parent);
+      parent = parent.parentElement;
+    }
+
+    this.parents = parents;
+
+    // Attach listeners
+    window.addEventListener('orientationchange', this.orientationchangeListener);
+    window.addEventListener('resize', this.resizeListener);
+    _.forEach(this.parents, parent => {
```
adimascio:
```suggestion
    for (const p of this.parents) {
        parent.addEventListener('scroll', this.scrollListener);
    }
```

src/lib/popover.ts
```diff
@@ -0,0 +1,8 @@
+export const POPOVER_ALIGN = {
```
adimascio: I would rather not put any display logic in `src/lib`. Could we put the module in `src/components` ? Either by creating a `Popover` directory (e.g. `Popover/index.vue` and `Popover/constants.vue`) or just by dropping it directly in `src/components`.

PR #143, opened 2019-05-24
Feature :: UI rename step

vdestraitt:
Main features:

- Click on column header to open the rename step configuration form in the left panel
- Autocomplete "oldname" based on selected column in the dataset viewer
- Change autocomplete "oldname" field when changing selected column in the dataset viewer 
- Return an error when "newname" is empty
- On validation, validate the form schema, render the new step in pipeline and launch the query

Related to https://github.com/ToucanToco/vue-query-builder/issues/112


src/components/FormRenameStep.vue
```diff
@@ -0,0 +1,128 @@
...
+  step: RenameStepConf = this.initialValue;
+
+  @Mutation toggleColumnSelection!: (column: string) => void;
+
+  @Getter selectedColumns!: string[];
+  @Getter columnNames!: string[];
+
+  @Watch('selectedColumns')
+  onSelectedColumnsChanged(val: string[], oldVal: string[]) {
+    if (val != oldVal) {
```
adimascio: Beware!, in javascript, array comparison doesn't behave as you seem to expect (e.g. not like python):

```javascript
> [1, 2] === [1, 2]
false
``` 

To avoid reinventing the wheel, use `_.isEqual` from `lodash`

PR #146, opened 2019-05-25
[rename-step] edit inside pipeline

adimascio:
This PR is for allowing the edition of a rename step when clicking on such an existing step's edit icon:

- toggle the edition mode when clicking on the edit icon, on validate or on cancel buttons
- populate the rename step form with already filled information when the step was created
- make the data viewer show the data resulting from the pipeline stoping at the step just before the step being edited
- manage the state correctly depending on whether the edition mode is toggled for a step creation or a step modification

Other additions:
- allow for inserting a step in the pipeline (rather than systematically appending it at the end)
- tests

Related to https://github.com/ToucanToco/vue-query-builder/issues/33 and https://github.com/ToucanToco/vue-query-builder/issues/112


Dalanir:
[ETQU, je peux modifier le nom d'une colonne via une interface](https://trello.com/c/NdOB9amJ/3209-etqu-je-peux-modifier-le-nom-dune-colonne-via-une-interface)

Issue #34, opened 2019-03-20
Toy backend for easier testing


adimascio:
As a developer, it would be nice to test the editor in real conditions with real data (e.g. CSV files or a toy mongo database).

One could imagime something like:

```bash
git clone https://github.com/ToucanToco/vue-query-builder/issues/new
python examples/toy-backend.py my-config-file.yml
yarn start-demo
```

piotch:
yo @adimascio we could maybe do this using node rather than python ? It avoids having two lang and al. to install 

- https://www.npmjs.com/package/mongodb-memory-server
- https://www.npmjs.com/package/csv

If you agree with this approach, happy to try to do something.

adimascio:
> yo @adimascio we could maybe do this using node rather than python ? It avoids having two lang and al. to install

Hey @piotch. Indeed, we could definitely go with a pure JS approach to avoid mixing languages and package managers.

PR #150, opened 2019-05-27
[components] move vqb logic to src/components

adimascio:
Dry-out playground to remove all code that belongs to VQB itself:

- assembling of `ResizePanel`, `DataViewer` and `Pipeline`,
- basic update logic

Now export a `Vqb` component which should be the main component imported
in host code, along with the `getTranslator()` factory.

For a concrete example, just checkout the new `playground/App.vue` code.

Next steps might be to make a pure-JS playground without bundling at all
to speed up development process.

Related to #34


playground/App.vue
```diff
@@ -1,105 +1,37 @@
...
-    ResizablePanels,
+    Vqb,
   },
+
 })
 export default class App extends Vue {
   @State pipeline!: Pipeline;
-  @State domains!: Array<string>;
 
   @Getter activePipeline!: Pipeline;
```
vdestraitt: @adimasci Is not `activePipeline` the only piece of data that we need to get from the store in this file as such ?
`pipeline`, and `setDomains` and `setDataset` mutations are not directly used here right ?

vdestraitt:
@adimascio when I launch the playground I cannot see the name of the collection anymore in the very first step of the pipeline ("Domain" dropdown)

PR #152, opened 2019-05-31
Improve :: Select pipeline step

vdestraitt:
Make a pipeline step fully clickable (not only the time travel dot)

Closes https://github.com/ToucanToco/vue-query-builder/issues/45

### DIFF
--- a/src/components/DataViewer.vue
+++ b/src/components/DataViewer.vue
@@ -30,7 +30,6 @@
   <div v-else>No data available</div>
 </template>
 <script lang="ts">
-import _ from 'lodash';
 import Vue from 'vue';
 import { Getter, Mutation, State } from 'vuex-class';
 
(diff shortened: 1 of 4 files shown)


adimascio:
Is the `@click` listener on the dot stil necessary then?

vdestraitt:
> Is the `@click` listener on the dot stil necessary then?

@adimasci I think the user can expect the whole step object to be clickable for time travel purposes. Without any data to prove the contrary I suggest to let the interaction available on the dot too to avoid any frustration...

adimascio:
> > Is the `@click` listener on the dot stil necessary then?
> 
> @adimasci I think the user can expect the whole step object to be clickable for time travel purposes. Without any data to prove the contrary I suggest to let the interaction available on the dot too to avoid any frustration...

Sure.. My point was not clear enough. I did not check but I think that when a user clicks on the dot, the event will bubble up to the enclosing div, thus not requiring to listen to `@click` directly on the dot. The div will catch it anyway.

PR #153, opened 2019-06-01
Improve :: column selection

vdestraitt:
A column can be selected when clicking on any of its cell in the dataset viewer.

Related to https://github.com/ToucanToco/vue-query-builder/issues/27


Dalanir:
It seems to me that is not an UX that we want to hilight because in the future we will select a speicif cell for actions. I close that PR for the moment, maybe we will re open it later.

PR #166, opened 2019-06-06
[Rename step] :: error if newname corresponds to an existing column name

vdestraitt:
Return an error if trying to rename a column with an already used column name.

Closes https://github.com/ToucanToco/vue-query-builder/issues/162

### DIFF
--- a/src/mixins/FormMixin.vue
+++ b/src/mixins/FormMixin.vue
@@ -2,9 +2,12 @@
 import { Component, Vue } from 'vue-property-decorator';
 import Ajv, { ValidateFunction, ErrorObject } from 'ajv';
 
-@Component    
+type VqbError = Partial<ErrorObject>;
+
+@Component
 export default class FormMixin extends Vue {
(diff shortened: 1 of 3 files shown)


src/mixins/FormMixin.vue
```diff
@@ -2,9 +2,12 @@
 import { Component, Vue } from 'vue-property-decorator';
 import Ajv, { ValidateFunction, ErrorObject } from 'ajv';
 
-@Component    
+type VqbError = Partial<ErrorObject>;
+
+@Component
 export default class FormMixin extends Vue {
-  errors?: ErrorObject[] | null = null;
+  // errors?: ErrorObject[] | null = null;
```
Dalanir:
```suggestion
```

vdestraitt:
> LGTM but that means that all forms have to implement that validation, can't we directly overload Ajv validator ?

@Dalanir not all forms (only those that will add a column). But we should for sure think about another approach, maybe in a refactoring PR when the code base will have matured a bit on this topic ?

PR #173, opened 2019-06-12
Feature :: Action toolbar popover 

zegonz:
- add new properties for button in `constant.ts`
- add new object `ACTION_CATEGORIES`
- add property `enable` to display a special style / remove hover cursor for disabled button

### DIFF
--- a/src/components/ActionToolbar.vue
+++ b/src/components/ActionToolbar.vue
@@ -6,11 +6,13 @@
         <span class="action-toolbar__btn-txt">Column</span>
       </button>
       <action-toolbar-button
-        v-for="(button, index) in buttons"
+        v-for="(button, index) in formattedButtons"
         :icon="button.icon"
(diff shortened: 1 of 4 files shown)


src/components/ActionToolbar.vue
```diff
@@ -7,10 +7,15 @@
...
 
       <action-toolbar-button
-        v-for="button in buttons"
+        v-for="(button, index) in formattedButtons"
         :icon="button.icon"
         :label="button.label"
         :key="button.icon"
+        :category="button.category"
+        :is-active="button.isActionToolbarMenuOpened"
+        :class="{ 'action-toolbar__btn--active': button.isActionToolbarMenuOpened, 'action-toolbar__btn--disable': !button.enable }"
```
Dalanir: A good practice is to used a computed here in order to be concise and keep conventions from Tucana

PR #179, opened 2019-06-19
Feature :: Avoid bad toggling of step edition mode

vdestraitt:
When a step edition form is already open and that we click on another operation, we want to keep the edition mode open, with the new form instantiated.

Before this PR, there was a trivial toggle of the edition mode which was based on the hypothesis that the edition mode was off before clicking on any operation, which is not reasonable. When the edition mode was already on it led us to click twice before opening the correct form: the first time to close the existing form, the second one to open the new form. Bad experience that is now improved !

### DIFF
--- a/src/components/ActionMenu.vue
+++ b/src/components/ActionMenu.vue
@@ -38,11 +38,13 @@ export default class ActionMenu extends Vue {
   columnName!: string;
 
   @State pipeline!: Pipeline;
+  @State isEditingStep!: boolean;
 
   @Getter computedActiveStepIndex!: number;
 
   @Mutation selectStep!: MutationCallbacks['selectStep'];
   @Mutation setPipeline!: MutationCallbacks['setPipeline'];
+  @Mutation toggleStepEdition!: () => void;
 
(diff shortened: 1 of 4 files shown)


src/components/Vqb.vue
```diff
@@ -61,6 +61,9 @@ export default class Vqb extends Vue {
 
   openStepForm(params: PipelineStep, index: number) {
     // after that, we delete from params to pass down the others keys to initialValue
+    if (this.isEditingStep) {
```
adimascio: Please, add a comment here in the code explaining why toggling step is required if an edition is ongoing

src/components/ActionMenu.vue
```diff
@@ -66,6 +68,9 @@ export default class ActionMenu extends Vue {
     const newPipeline: Pipeline = [...this.pipeline];
     const index = this.computedActiveStepIndex + 1;
     const deletecolumnStep: PipelineStep = { name: 'delete', columns: [this.columnName] };
+    if (this.isEditingStep) {
```
adimascio: Please add a comment here explaining why

PR #182, opened 2019-06-23
[steps] make step type definitons writable

adimascio:
Step type definitions have no good reason to be readonly. In fact,
it is actually a pain to have steps being readonly in step forms since
we can't use the step types directly on a data attribute without wrapping
them with a specific `Writable` type that removes the `readonly` modiifier.

What we actually want is to make steps readonly during the translation process,
so just declare transform step function parameters as such and everything is
fine.

### DIFF
--- a/src/lib/translators/base.ts
+++ b/src/lib/translators/base.ts
@@ -34,7 +34,7 @@ export class StepNotSupported extends Error {
  * @param descriptor the method descriptor
  */
 function unsupported(target: BaseTranslator, propertyKey: S.PipelineStepName, descriptor: any) {
-  descriptor.value = function(step: S.PipelineStep) {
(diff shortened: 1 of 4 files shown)


src/lib/translators/base.ts
```diff
@@ -157,7 +157,7 @@ export class BaseTranslator implements StepMatcher<OutputStep> {
       // otherwise it will complain about
       // `((x: DomainStep) => void)) | ((x: FilterStep) => void) | ((x: ...) => void)`
       // not being assignable to `((x: DomainStep | FilterStep | ...) => void)`
-      const callback = <TransformStep>this[step.name];
+      const callback = this[step.name] as TransformStep;
```
vdestraitt: What is the (subtle I guess) difference?

PR #188, opened 2019-06-24
PR :: UI Aggregate Step

vdestraitt:
Introduce a complete UI for creating aggregations including:

- A new "WidgetMultiselect" component to manage the selction of multiple columns at once
- A new "WidgetList" component to manage repeatable form inputs

The Aggregate step form can be opened via the "Aggregate" button of the action toolbar.

Closes https://github.com/ToucanToco/vue-query-builder/issues/193
Closes https://github.com/ToucanToco/vue-query-builder/issues/194
Closes https://github.com/ToucanToco/vue-query-builder/issues/195

### DIFF
--- a/src/components/stepforms/AggregateStepForm.vue
+++ b/src/components/stepforms/AggregateStepForm.vue
@@ -1,0 +1,116 @@
+<template>
+  <div>
+    <step-form-title :title="title"></step-form-title>
+    <WidgetMultiselect
+      id="groupbyColumnsInput"
+      v-model="editedStep.on"
+      name="Group by:"
+      :options="columnNames"
+      @input="setSelectedColumns({ column: editedStep.on[0] })"
+      placeholder="Add columns"
+    ></WidgetMultiselect>
+    <WidgetList
+      addFieldName="Add aggregation"
+      id="toremove"
+      name="Aggregations:"
+      v-model="aggregations"
+      :defaultItem="defaultAggregation"
+      :widget="'widget-aggregation'"
+      :automatic-new-field="false"
+    ></WidgetList>
+    <step-form-buttonbar :errors="errors" :cancel="cancelEdition" :submit="submit"></step-form-buttonbar>
+  </div>
+</template>
+
+<script lang="ts">
+import { Prop } from 'vue-property-decorator';
+import { AggFunctionStep, AggregationStep } from '@/lib/steps';
+import aggregateSchema from '@/assets/schemas/aggregate-step__schema.json';
+import WidgetMultiselect from './WidgetMultiselect.vue';
+import WidgetList from './WidgetList.vue';
+import BaseStepForm from './StepForm.vue';
+import { StepFormComponent } from '@/components/formlib';
+
+@StepFormComponent({
+  vqbstep: 'aggregate',
+  name: 'aggregate-step-form',
+  components: {
+    WidgetList,
+    WidgetMultiselect,
+  },
+})
+export default class AggregateStepForm extends BaseStepForm<AggregationStep> {
+  @Prop({ type: Object, default: () => ({ name: 'aggregate', on: [], aggregations: [] }) })
+  initialStepValue!: AggregationStep;
+
+  readonly title: string = 'Delete Column Step';
+
+  editedStepModel = aggregateSchema;
+
+  get defaultAggregation() {
+    const agg: AggFunctionStep = {
+      column: '',
+      newcolumn: '',
+      aggfunction: 'sum',
+    };
+    return agg;
+  }
+
+  get aggregations() {
+    if (this.editedStep.aggregations.length) {
+      return this.editedStep.aggregations;
+    } else {
+      return [this.defaultAggregation];
+    }
+  }
+
+  set aggregations(newval) {
+    this.editedStep.aggregations = [...newval];
+  }
+
+  submit() {
+    /**
+     * If different aggregations have to be performed on the same column, add a suffix
+     * to the automatically generated newcolumn name
+     */
+    const newcolumnOccurences: { [prop: string]: number } = {};
+    for (const agg of this.editedStep.aggregations) {
+      agg.newcolumn = agg.column;
+      newcolumnOccurences[agg.newcolumn] = (newcolumnOccurences[agg.newcolumn] || 0) + 1;
+    }
+    for (const agg of this.editedStep.aggregations) {
+      if (newcolumnOccurences[agg.newcolumn] > 1) {
+        agg.newcolumn = `${agg.newcolumn}-${agg.aggfunction}`;
+      }
+    }
+    this.$$super.submit();
+  }
+}
+</script>
+<style lang="scss" scoped>
+@import '../../styles/_variables';
+.widget-form-action__button {
+  @extend %button-default;
+}
+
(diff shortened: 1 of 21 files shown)


src/components/stepforms/WidgetList.vue
```diff
@@ -0,0 +1,165 @@
...
+  addFieldName!: string;
+
+  @Prop({ type: String, default: null })
+  id!: string;
+
+  @Prop({ type: String, default: '' })
+  name!: string;
+
+  @Prop({ type: String, default: '' })
+  placeholder!: string;
```
adimascio: Looks like this prop is unused in `WidgetList`.

src/components/stepforms/WidgetList.vue
```diff
@@ -0,0 +1,165 @@
...
+      valueCopy.push(this.defaultChildValue);
+    }
+    valueCopy.forEach(value => {
+      children.push({
+        isRemovable: valueCopy.length !== 1,
+        value,
+      });
+    });
+    return children;
+  }
```
adimascio: What about

```typescript
    return valueCopy.map(value => ({
        isRemovable: valueCopy.length !== 1,
        value,
    });
```

and then removing `const children: object[] = [];` declaration?

tests/unit/widget-list.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        propsData: {
+          automaticNewField: false,
+          name: 'Aggregation',
+        },
+      });
+      const addButtonWrapper = wrapper.find('button');
+      addButtonWrapper.trigger('click');
+
+      expect(wrapper.emitted()['input']).to.exist;
+      expect(wrapper.emitted()['input'][0][0][0]).to.eql('');
```
adimascio:
```suggestion
      expect(wrapper.emitted()['input'][0][0][0]).to.equal('');
```

src/components/stepforms/AggregateStepForm.vue
```diff
@@ -0,0 +1,122 @@
...
+})
+export default class AggregateStepForm extends BaseStepForm<AggregationStep> {
+  @Prop({ type: Object, default: () => ({ name: 'aggregate', on: [], aggregations: [] }) })
+  initialStepValue!: AggregationStep;
+
+  @Prop({
+    type: Boolean,
+    default: true,
+  })
+  isStepCreation!: boolean;
```
adimascio: This property should not be necessary because it is defined in the parent class.

src/components/stepforms/AggregateStepForm.vue
```diff
@@ -0,0 +1,122 @@
...
+
+  editedStepModel = aggregateSchema;
+
+  get defaultAggregation() {
+    const agg = {
+      column: '',
+      newcolumn: '',
+      aggfunction: 'sum',
+    };
+    return agg as AggFunctionStep;
```
adimascio: Is there a reason not to define type at declaration time (i.e. `const agg: AggFunctionStep =...`) rather than using an explicit cast?

src/components/stepforms/AggregateStepForm.vue
```diff
@@ -0,0 +1,122 @@
...
+    this.editedStep.aggregations = newval;
+  }
+
+  submit() {
+    /**
+     * If different aggregations have to be performed on the same column, add a suffix
+     * to the automatically generated newcolumn name
+     */
+    const newcolumnOccurences: { [prop: string]: number } = {};
+    this.editedStep.aggregations.forEach(agg => {
```
adimascio: what about using a simple `for ... of` loop instead of a `.forEach` and `agg.newcolumn` instead of `agg["newcolumn"]` ?

src/components/stepforms/AggregateStepForm.vue
```diff
@@ -0,0 +1,122 @@
...
+      agg['newcolumn'] = agg['column'];
+      newcolumnOccurences[agg.newcolumn] = (newcolumnOccurences[agg.newcolumn] || 0) + 1;
+    });
+    this.editedStep.aggregations.map(agg => {
```
adimascio: there's no reason to use `.map` here. We don't actually build a new list from the first one. A simple `for...of` loops would do the job.

tests/unit/aggregate-step-form.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+      await wrapper.find(WidgetMultiselect).trigger('input');
+      expect(store.state.selectedColumns).to.eql(['foo']);
+    });
+  });
+
+  describe('WidgetList', () => {
+    it('should have exactly on WidgetList component', () => {
+      const wrapper = shallowMount(AggregateStepForm, { store: emptyStore, localVue });
+      const widgetWrappers = wrapper.findAll('widgetlist-stub');
+      expect(widgetWrappers.length).to.eql(1);
```
adimascio:
```suggestion
      expect(widgetWrappers.length).to.equal(1);
```

tests/unit/aggregate-step-form.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+    it('should pass down the "aggregations" prop to the WidgetList value prop', async () => {
+      const wrapper = shallowMount(AggregateStepForm, { store: emptyStore, localVue });
+      wrapper.setData({
+        editedStep: {
+          name: 'aggregate',
+          on: [],
+          aggregations: [{ column: 'foo', newcolumn: 'bar', aggfunction: 'sum' }],
+        },
+      });
+      await Vue.nextTick();
```
adimascio: I know there are some bad examples in our code but I think it would be better to use `await localVue.nextTick()` here, `Vue` refers to another instance (as far as I understand)

tests/unit/aggregate-step-form.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+      expect(wrapper.find('widgetlist-stub').props().value).to.eql([
+        { column: 'foo', newcolumn: 'bar', aggfunction: 'sum' },
+      ]);
+    });
+
+    it('should have expected default aggregation parameters', () => {
+      const wrapper = mount(AggregateStepForm, { store: emptyStore, localVue });
+      const widgetWrappers = wrapper.findAll(WidgetAutocomplete);
+      expect(widgetWrappers.at(0).props().value).to.eql('');
```
adimascio:
```suggestion
      expect(widgetWrappers.at(0).props().value).to.equal('');
```

tests/unit/aggregate-step-form.spec.ts
```diff
@@ -0,0 +1,288 @@
...
+      expect(widgetWrappers.at(1).props().value).to.eql('sum');
```
adimascio:
```suggestion
      expect(widgetWrappers.at(1).props().value).to.equal('sum');
```

src/components/stepforms/AggregateStepForm.vue
vdestraitt: Absolutely not ! :)

src/components/stepforms/AggregateStepForm.vue
vdestraitt: No problem ! Is this suggestion for improved readability ?

PR #189, opened 2019-06-24
Feature :: Domain Step in the Pipeline 

Dalanir:
In order to have the possibility to time travel back to the domain choice, we now consider it like a Step unremovable (I'd like feedback on this, we can maybe improve the check). We had to aslo create a form component in order to change the domain because is no long a dropdown.

### DIFF
--- a/src/components/Vqb.vue
+++ b/src/components/Vqb.vue
@@ -71,7 +71,8 @@ export default class Vqb extends Vue {
     this.initialValue = { ...params }; // make a copy
     if (index !== undefined) {
       this.editedStepIndex = index;
-      this.selectStep({ index: index - 1 });
+      const prevIndex = Math.max(index - 1, 0);
+      this.selectStep({ index: prevIndex });
     } else {
       this.editedStepIndex = -1;
     }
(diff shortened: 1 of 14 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@ node_modules
 /dist
 coverage
 playground/*.map
+playground/dist/*
```
adimascio: Please, remove this line. We do need to track some of the files inside `playground/dist` (e.g. `index.html`). Besides, this change should be made in another PR.

src/components/Vqb.vue
```diff
@@ -70,7 +71,8 @@ export default class Vqb extends Vue {
     this.initialValue = { ...params }; // make a copy
     if (index !== undefined) {
       this.editedStepIndex = index;
-      this.selectStep({ index: index - 1 });
+      const prevIndex = index === 0 ? index : index - 1;
```
adimascio: Another way would to do:

```suggestion
      const prevIndex = Math.max(index - 1, 0);
```

Don't know if it's more readable or not.

src/components/Vqb.vue
```diff
@@ -82,13 +84,18 @@ export default class Vqb extends Vue {
     // Reset value from DataViewer
     this.initialValue = undefined;
     const newPipeline: Pipeline = [...this.pipeline];
-    const index = this.computedActiveStepIndex + 1;
+    // FIXME: not sur about that specific implem to handle `domain` step
```
adimascio: Small typo in the comment

tests/unit/domain-step-form.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    emptyStore = setupStore({});
+  });
+
+  it('should instantiate', () => {
+    const wrapper = shallowMount(DomainStepForm, { store: emptyStore, localVue });
+
+    expect(wrapper.exists()).to.be.true;
+  });
+
+  it('should have exactly one widgetinputtext component', () => {
```
adimascio:
```suggestion
  it('should have exactly one widgetautocomplete component', () => {
```

.gitignore
Dalanir: Right ! My bad

PR #210, opened 2019-06-27
AggregateStepForm :: Use the new way to update property when selectedColumns change

Dalanir:
Use the new way to update property when selectedColumns change using the reactive property inherit of BaseStepForm


Dalanir:
You're right, I'll make the changes !

Dalanir:
@vdestraitt Changes done !

vdestraitt:
@Dalanir thank you, juste added a micro change: when clicking on a column to add it in the multi select, we want to keep focus on the last selected column (not the first as it was originally set).

PR #212, opened 2019-06-27
Feature :: Generic way to open forms

Dalanir:
In order to avoid writting special case for each steps. To do so, we now open form base on their names. The default case is defined in all forms and properties that we want to link to `selectedColumns` has already the logic to update itself when the getters was modified, I just add the fact is also done at the mounted cycle too.

### DIFF
--- a/src/components/ActionToolbarButton.vue
+++ b/src/components/ActionToolbarButton.vue
@@ -9,6 +9,7 @@
             v-for="(item, index) in items"
             :key="index"
             class="action-menu__option"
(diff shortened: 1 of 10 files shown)


src/components/ActionToolbarButton.vue
```diff
@@ -9,6 +9,7 @@
             v-for="(item, index) in items"
             :key="index"
             class="action-menu__option"
+            @click="actionClicked(index)"
```
Dalanir: Send directly `item.name`

PR #214, opened 2019-06-28
Feature :: Duplicate column step

vdestraitt:
Closes https://github.com/ToucanToco/vue-query-builder/issues/213

### DIFF
--- a/src/components/stepforms/DuplicateColumnStepForm.vue
+++ b/src/components/stepforms/DuplicateColumnStepForm.vue
@@ -1,0 +1,55 @@
+<template>
+  <div>
+    <step-form-title :title="title"></step-form-title>
+    <WidgetAutocomplete
+      id="columnInput"
+      v-model="editedStep.column"
+      name="Duplicate column..."
+      :options="columnNames"
+      @input="setSelectedColumns({ column: stepSelectedColumn })"
+      placeholder="Enter a column"
+    ></WidgetAutocomplete>
+    <WidgetInputText
+      id="newColumnNameInput"
+      v-model="editedStep.new_column_name"
+      name="New column name:"
+      placeholder="Enter a column name"
+    ></WidgetInputText>
+    <step-form-buttonbar :errors="errors" :cancel="cancelEdition" :submit="submit"></step-form-buttonbar>
+  </div>
+</template>
+
+<script lang="ts">
+import { Prop } from 'vue-property-decorator';
+import { StepFormComponent } from '@/components/formlib';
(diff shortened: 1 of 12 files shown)


src/lib/steps.ts
```diff
@@ -48,6 +48,11 @@ export type DomainStep = {
   domain: string;
 };
 
+export type DuplicateColumnStep = {
+  name: 'duplicate';
+  column: string;
```
adimascio: Shouldn't we, for consistency-sake with other steps, add a `new_column?: string;` field?

src/components/stepforms/DuplicateColumnStepForm.vue
```diff
@@ -0,0 +1,65 @@
...
+  @Prop({ type: Object, default: () => ({ name: 'duplicate', column: '' }) })
+  initialStepValue!: DuplicateColumnStep;
+
+  readonly title: string = 'Duplicate Column Step';
+
+  // Only manage the deletion of 1 column at once at this stage
+  editedStepModel = duplicateSchema;
+  column: string = this.initialStepValue.column;
+
+  stepToValidate() {
```
adimascio: Why is this needed? It looks like the default behaviour to me.

src/components/stepforms/DuplicateColumnStepForm.vue
```diff
@@ -0,0 +1,65 @@
...
+  set stepSelectedColumn(colname: string | null) {
+    if (colname === null) {
+      throw new Error('should not try to set null on duplicate "column" field');
+    }
+    if (colname !== null) {
+      this.column = colname;
+    }
+  }
+
+  rebuildStep() {
```
adimascio: Same as above, Is this needed at all?

adimascio: Shouldn't we check that we're not overriding an existing column in the dataset?

src/lib/steps.ts
vdestraitt: @adimasci it was on purpose: we wanted, as for the `delete` step, to get the column creation immediately effective after clicking, without having to fill in any form field. It is inspired from other softwares and looks lighter, offering a more direct feedback to the user.

But after reflexion, I think that you're right and that in our context we should ask for the new name in a form at creation, because a duplicated column will need a rename in 95% of the cases.
And it would be way heavier than other softwares I had in mind if we have to log an additional `rename` step in the pipeline every time we duplicate a column !

So I will amend this ! Thank you

PR #215, opened 2019-06-29
Fix :: Remove stepSelectedColumn getter and setter in AggregateStepForm

vdestraitt:
The stepSelectedColumn getter and setter functions were designed to manage some interactions with the table but had side effects.
We prefer no interaction rather than bad. I think that this interaction matter deserves a specific PR. In the meantime, forms work very well (or even better) without them !

**Exemple of issue:**

Context:

- An aggregate step has already been created in the pipeline and we want to modify it.
- A column which is not in the group by (`on`) parameter of the aggregate step form, is focused in the dataset

When clicking on edit: the focused column gets added to `editedStep.on` on opening.
When clicking on cancel: it does not cancel the undesired change (editedStep.on has been muted and cancel has the same effect as submit => there may be an issue to tackle in our BaseStepForm ?)

### DIFF
--- a/src/components/stepforms/AggregateStepForm.vue
+++ b/src/components/stepforms/AggregateStepForm.vue
@@ -23,7 +23,6 @@
 </template>
 
 <script lang="ts">
-import _ from 'lodash';
 import { Prop } from 'vue-property-decorator';
 import { AggFunctionStep, AggregationStep } from '@/lib/steps';
 import aggregateSchema from '@/assets/schemas/aggregate-step__schema.json';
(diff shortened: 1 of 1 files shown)


adimascio:
Thanks for spotting this bug! I have two problems, though:

- I would rather fix the behaviour rather than disabling it. I haven't put too many thoughts in it but I would say it's mainly a matter of checking wether we're editing the step or not?

- Doesn't the problem occur on all steps?

Dalanir:
I agree with you too that's why I think if we put an effort on improve column selection behavior, it will fix that without loose on the UX side

PR #222, opened 2019-07-01
Feature :: Improve forms diplay

vdestraitt:
This PR mainly consists in aesthetics:

- Clean duplicated CSS
- Improve styling: text sizing of labels, placeholders & buttons, text color, title alignment and  autocomplete/multiselect items highlight color
- Improve naming: labels & title

**Before:**

**After:**

**Before:**

**After:**

### DIFF
--- a/src/components/stepforms/DomainStepForm.vue
+++ b/src/components/stepforms/DomainStepForm.vue
@@ -4,7 +4,7 @@
     <WidgetAutocomplete
       id="domainInput"
       v-model="editedStep.domain"
-      name="Change your domain:"
+      name="Select domain..."
       :options="domains"
       placeholder="Choose a domain"
     ></WidgetAutocomplete>
@@ -40,7 +40,7 @@ export default class DomainStepForm extends BaseStepForm<DomainStep> {
 
   @State domains!: string[];
 
-  readonly title: string = 'Change your domain';
+  readonly title: string = 'Select a domain';
 
   editedStepModel = domainSchema;
 }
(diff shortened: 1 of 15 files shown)


src/components/stepforms/DomainStepForm.vue
```diff
@@ -40,7 +40,7 @@ export default class DomainStepForm extends BaseStepForm<DomainStep> {
 
   @State domains!: string[];
 
-  readonly title: string = 'Change your domain';
+  readonly title: string = 'Select a domain';
```
Dalanir: We open the DomainStep form only for the edition, I think for choosing it we've got a special CTA

vdestraitt: Actually "Change" sounds less natural English to my ears in such a context. To say "Change your domain" I would rather say "Select a new domain" but I am not an linguistic expert !!

Also, I may decide not to "change" the domain, I can leave it unchanged eventually. The 100% truth is in the selection, not the change :)

And my last answer will be: we are not searching for an important CTA like a payment button, so I think we can be quite flexible on synonyms ;)

vdestraitt:
> One comment plus to store #a7a7a7 hex code in var and we're good, thanks for that :)

Done thanks !

vdestraitt:
> Seems really good to me !
> Just one thing I'm not sure if the color of the placeholder text has enough contrast for accessibility :
> image
> 
> Update : I Check it on a web accessibility tool (source https://webaim.org/resources/contrastchecker/).
> Results : it's not ok
> image

vdestraitt:
> Seems really good to me !
> Just one thing I'm not sure if the color of the placeholder text has enough contrast for accessibility 
> Update : I Check it on a web accessibility tool

@zegonz thanks I replaced this color with a much darker grey !

src/components/stepforms/DomainStepForm.vue
Dalanir: 👍

PR #223, opened 2019-07-01
Implement dynamic jsonschema

adimascio:
- feed a `stepname` attribute to StepForm components

- provide schema factories indexed by step name instead of static jsonschemas.
  The factories accept a `form` parameter that is the form component
  instance. For intsance, it is therefore possible, inside the factory, to
  access the list of column names via the `columnNames` attribute.


src/components/stepforms/StepForm.vue
```diff
@@ -98,12 +99,16 @@ export default class BaseStepForm<StepType> extends Vue {
   }
 
   created() {
+    this.intiialize();
```
Dalanir: I guess there's a typo here: `this.initialize();`

tests/unit/percentage-step-form.spec.ts
```diff
@@ -91,7 +91,7 @@ describe('Percentage Step Form', () => {
         keyword: err.keyword,
         dataPath: err.dataPath,
       }));
-      expect(errors).to.eql([{ keyword: 'nameAlreadyUsed', dataPath: '.new_column' }]);
+      expect(errors).to.eql([{ keyword: 'not', dataPath: '.new_column' }]);
```
Dalanir: I like the approach but now we've got an error message less clear than before :/

adimascio: Indeed, that's something I would need to dig a bit more but I'm not sure this is something we can change or configure.

src/components/stepforms/schemas/delete.ts
```diff
@@ -0,0 +1,26 @@
...
+      attrs: {
+        placeholder: 'Enter a column',
+      },
+    },
+  },
+  required: ['name', 'column'],
+  additionalProperties: false,
+};
+
+export default function buildSchema(_form: any) {
```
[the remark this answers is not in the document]
adimascio: Just to make I understand your point, you would prefer writing:

```javascript
export default { ... }
```

rather than

```javascript
export default() => { ... }
```

I'm ok with it, although it would probabliy require a `typeof ` test in the caller code.

src/components/stepforms/StepForm.vue
```diff
@@ -98,12 +99,16 @@ export default class BaseStepForm<StepType> extends Vue {
...
+  }
+
+  intiialize() {
```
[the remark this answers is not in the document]
adimascio: Indeed, but on the bright side... error is consistent :)

src/components/stepforms/schemas/delete.ts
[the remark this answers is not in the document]
vdestraitt: Not sure it is the same question than @Dalanir but: do we need to go through buildSchema here ? It does nothing here right ? So I think the question was: can't we just export it directly like the argmin schema for exemple ?

src/components/stepforms/schemas/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+  properties: { [prop: string]: JsonPropSchema };
+};
+
+export type StepFormType = {
+  columnNames: string[];
+};
+
+export function addNotInColumnNamesConstraint(
+  model: JsonSchema,
+  stepname: string,
```
vdestraitt: Actually it is more a filed name that the step name that we use here right ? If so I find it confusing...

src/components/stepforms/schemas/delete.ts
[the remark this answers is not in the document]
Dalanir: That's exactly what I wanna say @vdestraitt

src/components/stepforms/StepForm.vue
```diff
@@ -90,6 +90,7 @@ export default class BaseStepForm<StepType> extends Vue {
   editedStep: Writable<StepType> = { ...this.initialStepValue };
   editedStepModel!: object;
   errors?: VqbError[] | null = null;
+  stepname!: string; // fed by @SstepFormComponent
```
Dalanir: Typo: `@StepFormComponent`

src/components/stepforms/StepForm.vue
```diff
@@ -18,7 +20,7 @@ type VqbError = Partial<ErrorObject>;
  */
 function componentProxyBoundOn(self: Vue) {
   return {
-    get: function(target: any, prop: string) {
+    get: function (target: any, prop: string) {
```
Dalanir: I'm gonna kill your linter one day

PR #233, opened 2019-07-03
Introduce a Top step

vdestraitt:
Closes https://github.com/ToucanToco/vue-query-builder/issues/232

### DIFF
--- a/src/components/stepforms/TopStepForm.vue
+++ b/src/components/stepforms/TopStepForm.vue
@@ -1,0 +1,62 @@
+<template>
+  <div>
+    <step-form-title :title="title"></step-form-title>
+    <WidgetInputText
+      id="limitInput"
+      v-model.number="editedStep.limit"
+      type="number"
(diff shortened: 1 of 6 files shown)


src/components/stepforms/TopStepForm.vue
```diff
@@ -0,0 +1,63 @@
...
+    WidgetInputText,
+    WidgetMultiselect,
+  },
+})
+export default class TopStepForm extends BaseStepForm<TopStep> {
+  @Prop({ type: Object, default: () => ({ name: 'top', rank_on: '', sort: 'asc' }) })
+  initialStepValue!: TopStep;
+
+  readonly title: string = 'Top N rows';
+  sortOrder: TopStep['sort'][] = ['asc', 'desc'];
```
adimascio: We probably don't need a data prop for `sortOrder`: we don't need to modify it. I think we should just hardcode these option values in the widget.

PR #238, opened 2019-07-05
Feature :: Infer type of DataSetColumn

### LINKED ISSUE #237: Infer types from column values
In order to cover the case when the backend don't return the types of colums and keep the availability to typecast (#44) on the fly some columns, we have to implement a method that can guess types from values.

Dalanir:
In order to eventually implement `cast` step and improve the steps suggestions in our ActionToolbar, we implement a function that guess type of columns according a nb of first rows. Of course, others backend can provide us the types directly but if it's not the case, we've got this little alternative.

That closes #237 

Now
Capture d’écran 2019-07-05 à 17 12 37

### DIFF
--- a/src/lib/dataset/mongo.ts
+++ b/src/lib/dataset/mongo.ts
@@ -67,3 +67,41 @@ export function _guessType(val: any, prevType: any = null): DataSetColumnType | 
 
   return null;
 }
+
+/**
+ * @descritpion extend a `DataSet` headers with inferred types, base on value of columns
+ * @param {DataSet} dataset - The dataset that we want to extend headers with inferred types
+ * @param {number} [maxRows=50] - Determine on how many rows we're goind to do our guess
+ *
+ */
+export function inferTypeFromDataset(dataset: DataSet, maxRows: number = 50): DataSet {
+  // For the moment, we infer arbitrarily column's types on maximum the 50 first rows
+  maxRows = Math.min(dataset.data.length, maxRows);
+  const newHeaders: DataSetColumn[] = [];
+
+  for (const [colidx, header] of dataset.headers.entries()) {
+    let prevType = null;
+
+    for (let j = 0; j < maxRows; j++) {
+      const currentValue = dataset.data[j][colidx];
+      const guessedType = _guessType(currentValue, prevType);
+
+      if (prevType === null) {
+        prevType = guessedType;
+      } else if (prevType !== guessedType) {
+        prevType = null;
+        break;
+      }
+    }
+
+    // We pass here if we find 2 different types in the same column
+    if (prevType === null) {
+      newHeaders.push(header);
+    } else {
+      newHeaders.push({ ...header, type: prevType as DataSetColumnType });
+    }
+  }
+
+  // We return a new `DataSet` with the updated headers
+  return { ...dataset, headers: newHeaders };
+}
--- a/src/lib/dataset/mongo.ts
+++ b/src/lib/dataset/mongo.ts
@@ -2,7 +2,7 @@
  * mongo specific helpers for dataset manipulation.
  */
 
-import { DataSet } from '@/lib/dataset';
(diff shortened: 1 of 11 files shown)


src/lib/dataset/mongo.ts
```diff
@@ -67,3 +67,38 @@ export function _guessType(val: any, prevType: any = null): DataSetColumnType |
 
   return null;
 }
+
+/**
+ * extend a `DataSet` headers with inferred types, base on value of columns
```
adimascio: Could we maybe add some docstring about the `maxRows` parameter?

src/lib/dataset/mongo.ts
```diff
@@ -67,3 +67,38 @@ export function _guessType(val: any, prevType: any = null): DataSetColumnType |
...
+ */
+export function inferTypeFromDataset(dataset: DataSet, maxRows: number = 50): DataSet {
+  // For the moment, we infer arbitrarily column's types on maximum the 50 first rows
+  maxRows = Math.min(dataset.data.length, maxRows);
+  const newHeaders: DataSetColumn[] = [];
+
+  dataset.headers.forEach((header, i) => {
```
adimascio: I would use of `for..of` loop on `header.entries()` although you might need to set `downlevelIteration` in `tscfongig.json` to be perform a `for of` loop on an iterator. Otherwise, `Object.entries()` should also do the trick:

```typescript
for (const [i, header] of dataset.headers.entries()) {
   // ...
} 
```

Also, it might be worth to rename `i` into `colidx` :-)

src/lib/dataset/mongo.ts
```diff
@@ -67,3 +67,38 @@ export function _guessType(val: any, prevType: any = null): DataSetColumnType |
...
+    let prevType = null;
+
+    for (let j = 0; j < maxRows; j++) {
```
adimascio:
```suggestion
for (const rowData of dataset.data) {
  const currentValue = rowData[colIdx];  // ← assuming that you've renamed `i` into `colIdx`
```

tests/unit/dataset.spec.ts
```diff
@@ -135,3 +140,141 @@ describe('_guessType', () => {
...
+    };
+    const datasetWithInferredType = inferTypeFromDataset(dataset);
+    expect(datasetWithInferredType.headers).to.eql([
+      { name: 'city', type: 'string' },
+      { name: 'population', type: 'integer' },
+      { name: 'isCapitalCity', type: 'boolean' },
+    ]);
+  });
+
+  it('should infer type even with mixed values if we not enough rows', () => {
```
adimascio:
```
  it('should infer type even with mixed values if we set maxRows low enough', () => {
```

tests/unit/dataset.spec.ts
```diff
@@ -135,3 +140,141 @@ describe('_guessType', () => {
...
+    const dataset: DataSet = {
+      headers: [{ name: 'city' }, { name: 'population' }, { name: 'isCapitalCity' }],
+      data: [['Paris', 10000000, null], ['Paris', 10000000, false]],
+    };
+
+    const datasetWithInferredType = inferTypeFromDataset(dataset, 1);
+    expect(datasetWithInferredType.headers).to.eql([
+      { name: 'city', type: 'string' },
+      { name: 'population', type: 'integer' },
+      { name: 'isCapitalCity' },
```
adimascio: Hmm, shouldn't it be `{ name: 'isCapitalCity', type: 'boolean' }`?

Dalanir: You got it right, the test pass because I let `maxRows` set to 1

src/lib/dataset/mongo.ts
Dalanir: How do you propose to handle `maxRows` now ?

adimascio: Ah yes, your're right! Calling `.slice(maxRows)` would be a solution I guess, although less optimal computationally speaking.

PR #245, opened 2019-07-11
[Step] Improve Filter step

vdestraitt:
Improve Filter step:

- The API now allows for complex conditions (imbricated and/or)
- The UI allows to configure several conditions, linked by a logical "and" (at this stage the UI does not allow to build "or" conditions)
- For every condition, the user input depends on the selected operator (multiple values for in/nin, single value for the other operators)

Related to https://github.com/ToucanToco/vue-query-builder/issues/56

### DIFF
--- a/src/components/stepforms/WidgetFilterSimpleCondition.vue
+++ b/src/components/stepforms/WidgetFilterSimpleCondition.vue
@@ -1,0 +1,124 @@
+<template>
+  <div class="filter-form-single-condition__container">
+    <WidgetAutocomplete
+      id="columnInput"
+      v-model="editedValue.column"
+      name="Values in..."
+      :options="columnNames"
+      @input="setSelectedColumns({ column: editedValue.column })"
+      placeholder="Column"
+    ></WidgetAutocomplete>
+    <WidgetAutocomplete
+      id="filterOperator"
+      :value="getShortOperator(editedValue.operator)"
+      @input="updateStepOperator"
+      name="Must..."
+      :options="operators"
+      placeholder="Filter operator"
+    ></WidgetAutocomplete>
+    <WidgetInputText id="valueInput" v-model="editedValue.value" placeholder="Value"></WidgetInputText>
+  </div>
+</template>
+
+<script lang="ts">
+import { Component, Prop, Vue, Watch } from 'vue-property-decorator';
+import { Getter, Mutation } from 'vuex-class';
+import { MutationCallbacks } from '@/store/mutations';
+import WidgetAutocomplete from '@/components/stepforms/WidgetAutocomplete.vue';
+import WidgetInputText from '@/components/stepforms/WidgetInputText.vue';
+import { FilterSimpleCondition } from '@/lib/steps';
+
+type LiteralOperator =
+  | 'equal'
+  | 'not equal'
+  | 'be greater than'
+  | 'be greater than or equal to'
+  | 'be less than'
+  | 'be less than or equal to'
+  | 'be among'
+  | 'not be among';
+
+type ShortOperator = FilterSimpleCondition['operator'];
+
+const OPERATORS = {
+  equal: 'eq',
+  'not equal': 'ne',
+  'be greater than': 'gt',
+  'be greater than or equal to': 'ge',
+  'be less than': 'lt',
+  'be less than or equal to': 'le',
+  'be among': 'in',
+  'not be among': 'nin',
+};
+
+const OPERATORS_REV = {
+  eq: 'equal',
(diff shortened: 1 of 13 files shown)


src/components/stepforms/WidgetFilterSimpleCondition.vue
```diff
@@ -0,0 +1,124 @@
...
+  })
+  value!: FilterSimpleCondition;
+
+  @Getter columnNames!: string[];
+
+  @Mutation setSelectedColumns!: MutationCallbacks['setSelectedColumns'];
+
+  editedValue: FilterSimpleCondition = { ...this.value };
+
+  readonly operators = [
```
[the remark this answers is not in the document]
adimascio: Better yet... you should probably do something like:

```typescript
const options = [
  {operator: 'eq', label: 'equal'},
  {operator: 'neq': label: 'not equal'},
  // ...
];
```

and then use the `object` mode as in https://vue-multiselect.js.org/#sub-single-select-object:

```html
  <multiselect :options="operators" track-by='operator' label='name' :allow-empty="false">
  </multiselect>
```

I don't think you need a custom `slot` as demonstrated by the example for such a simple case.

src/lib/steps.ts
```diff
@@ -60,13 +60,6 @@ export type FillnaStep = {
   value: PrimitiveType;
 };
 
-// export type FilterStep = {
```
adimascio: Not very important but if would great if you could directly update the first commit to remove these commented lines.

tests/unit/widget-filter-simple-condition.spec.ts
```diff
@@ -22,9 +23,17 @@ describe('Widget WidgetAggregation', () => {
...
     const inputtextWrappers = wrapper.findAll('widgetinputtext-stub');
-    expect(inputtextWrappers.length).to.eql(1);
+    expect(inputtextWrappers.length).to.equal(1);
+  });
+
+  it('should have exactly have a WidgetMultiInputText if operator is "in" or "nin"', async () => {
+    const wrapper = shallowMount(WidgetFilterSimpleCondition, { store: emptyStore, localVue });
+    wrapper.setData({ editedValue: { column: 'foo', value: [], operator: 'in' } });
+    await localVue.nextTick();
+    const autocompleteWrappers = wrapper.findAll('widgetmultiinputtext-stub');
```
adimascio: genuine question: is the `nextTick()` required here?

tests/unit/widget-filter-simple-condition.spec.ts
```diff
@@ -55,6 +64,16 @@ describe('Widget WidgetAggregation', () => {
     expect(widgetWrappers.at(1).props().value).to.equal('be among');
   });
 
+  it('should channge the type of value accordingly when switching the "operator"', async () => {
+    const wrapper = shallowMount(WidgetFilterSimpleCondition, { store: emptyStore, localVue });
+    expect((wrapper.vm.$data.editedValue.value = ''));
+    wrapper.setData({ editedValue: { column: 'foo', value: '', operator: 'in' } });
+    await localVue.nextTick();
```
adimascio: same here

src/components/stepforms/schemas/filter.ts
```diff
@@ -35,6 +29,63 @@ export default {
...
+
+const complexConditionSchema = {
+  id: 'complexCondition',
+  type: 'object',
+  oneOf: [
+    {
+      properties: {
+        and: {
+          type: 'array',
+          minItems: 1,
```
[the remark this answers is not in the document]
vdestraitt: @adimascio if you have a simple condition it is expressed as `and: [{ single_condition }]` so I think we should let `minItems` = 1.

tests/unit/widget-filter-simple-condition.spec.ts
vdestraitt: Tested and it does not work...

tests/unit/widget-filter-simple-condition.spec.ts
vdestraitt: This one works fine without `nextTick`...

src/components/stepforms/WidgetFilterSimpleCondition.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <div class="filter-form-single-condition__container">
+    <WidgetAutocomplete
```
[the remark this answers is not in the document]
vdestraitt: I don't think so because it is part of a repeatable fields set and we do not support more than one columnPicker at the moment

PR #250, opened 2019-07-12
Feature :: search bar new branch and finalization

zegonz:
So now we get the first step  of the search bar :
image
Wow it's amazing isn't it ?

- The search bar use the vue `multiselect` widget to display the results. 
- For now it's not displaying the operations grouped by category but with multiselect it's going to be simplier that we thought.  
- you can use your keyboard as well to navigate. 
- There are few css improvement for this specific multiselect.

### DIFF
--- a/src/components/SearchBar.vue
+++ b/src/components/SearchBar.vue
@@ -1,7 +1,6 @@
 <template>
   <div class="search-bar">
     <multiselect
-      v-model="editedValue"
       :options="actionOptions"
       label="label"
       track-by="label"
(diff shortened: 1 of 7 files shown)


src/components/SearchBar.vue
```diff
@@ -0,0 +1,111 @@
...
+  name: 'search-bar',
+  components: {
+    Multiselect,
+  },
+})
+export default class SearchBar extends Vue {
+  actionOptions: ActionCategory[] = SEARCH_ACTION;
+  editedValue: string[] = [];
+
+  @Watch('value', { immediate: true })
```
Dalanir: There's no `value` property

src/components/SearchBar.vue
```diff
@@ -0,0 +1,111 @@
...
+  updateEditedValue(newValue: string[]) {
+    this.editedValue = newValue;
+  }
+
+  @Watch('editedValue')
```
Dalanir: It doesn't seem necessary to listen on `editedValue` changes

zegonz:
@Dalanir : thx for you review ! 
I removed the useless watcher code and did few CSS optimization (cleaning and fix an ugly gap on focus).

PR #254, opened 2019-07-16
Infer types :: Infer float type when a column has a mix of integers and floats

vdestraitt:
Before that PR, when in a column there were numbers of mixed types between integer and float, we would not infer any type. But as floats can include integers, we are better off inferring a float type in such cases.

### DIFF
--- a/src/lib/dataset/mongo.ts
+++ b/src/lib/dataset/mongo.ts
@@ -89,8 +89,14 @@ export function inferTypeFromDataset(dataset: DataSet, maxRows: number = 50): Da
       if (prevType === null) {
         prevType = guessedType;
       } else if (prevType !== guessedType) {
-        prevType = null;
-        break;
+        // if integers and floats are mixed, guessed type should be 'float'
+        const typeSet = new Set([prevType, guessedType]);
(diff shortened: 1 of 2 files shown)


src/lib/dataset/mongo.ts
```diff
@@ -3,6 +3,7 @@
  */
 
 import { DataSet, DataSetColumn, DataSetColumnType } from '@/lib/dataset';
+import { type } from 'os';
```
adimascio: This seems like an automatically added import from VSCode :)

src/lib/dataset/mongo.ts
```diff
@@ -89,8 +90,13 @@ export function inferTypeFromDataset(dataset: DataSet, maxRows: number = 50): Da
       if (prevType === null) {
         prevType = guessedType;
       } else if (prevType !== guessedType) {
-        prevType = null;
-        break;
+        const typeSet = new Set([prevType, guessedType]);
```
adimascio: In this specific case, I think an explicit test like this one is easier to read:

```typescript
if ((prevType === 'float' && guessedType === 'integer') || (prevType === 'integer' && guesstedType === 'float')) {
    prevType = 'float';
}
```

PR #256, opened 2019-07-16
Fillna step :: Cast user input depending on column data type

vdestraitt:


adimascio:
Looks overall good but reading this, there's something I don't get: why do we need to modify the json schema _and_ cast before emitting the new step? Shouldn't we do it the other way around, i.e update the input type dynamically when selected column changes? In that case, there might simply be nothing to do on submit? We might get problems with booleans though.

adimascio:
> Looks overall good but reading this, there's something I don't get: why do we need to modify the json schema _and_ cast before emitting the new step? Shouldn't we do it the other way around, i.e update the input type dynamically when selected column changes? In that case, there might simply be nothing to do on submit? We might get problems with booleans though.

OK. This cast is actually done _before_ schema validaiton, everything's fine.

PR #257, opened 2019-07-17
Fix WidgetAggregation :: Watch 'aggregation' data rather than 'value' prop

vdestraitt:
The previous code worked but in a roundabout way:

- `value` was set as `undefined` because no default was defined for this prop
- We copied the `value` prop in the `aggregation` data
- We watched a change in `value` to emit `aggregation` (weird already :))
- As such, `value` only changed once, at created (when switching from `undefined` to `defaultAggregation`)
- We therefore emitted `aggregation` at this first and only watched change
- At this stage, any subsequent change in `aggregation` impacted the parent only by reference and not by way of the watcher

So we propose here to clean the code to make it lighter:
- we set a valid default value for `value` prop directly through it decorator and not from a getter called at created
- we watch `aggregation` data rather than the `value` prop

### DIFF
--- a/src/components/stepforms/WidgetAggregation.vue
+++ b/src/components/stepforms/WidgetAggregation.vue
@@ -23,12 +23,6 @@ import { Getter } from 'vuex-class';
 import { AggFunctionStep } from '@/lib/steps';
 import WidgetAutocomplete from './WidgetAutocomplete.vue';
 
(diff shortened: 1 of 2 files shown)


adimascio:
Could you elaborate a bit in your commit message on the rationale behind this change?

PR #261, opened 2019-07-23
Introduce a replace step ui component

vdestraitt:
Closes https://github.com/ToucanToco/vue-query-builder/issues/244

### DIFF
--- a/src/components/stepforms/ReplaceStepForm.vue
+++ b/src/components/stepforms/ReplaceStepForm.vue
@@ -1,0 +1,102 @@
+<template>
+  <div>
+    <step-form-title :title="title"></step-form-title>
+    <ColumnPicker
+      id="searchColumnInput"
+      v-model="editedStep.search_column"
+      name="Search in column..."
+      placeholder="Enter a column"
+    ></ColumnPicker>
+    <WidgetInputText
+      id="newColumnInput"
+      v-model="editedStep.new_column"
+      name="(Optional) Write output in..."
+      placeholder="Enter a new column name"
+    ></WidgetInputText>
+    <WidgetList
+      addFieldName="Add a value to replace"
+      id="toReplace"
+      name="Values to replace:"
+      v-model="toReplace"
+      :defaultItem="[]"
+      :widget="widgetToReplace"
+      :automatic-new-field="false"
+    ></WidgetList>
+    <step-form-buttonbar :errors="errors" :cancel="cancelEdition" :submit="submit"></step-form-buttonbar>
+  </div>
+</template>
+
+<script lang="ts">
+import { Prop } from 'vue-property-decorator';
+import { Getter } from 'vuex-class';
+import { StepFormComponent } from '@/components/formlib';
+import ColumnPicker from '@/components/stepforms/ColumnPicker.vue';
+import WidgetInputText from '@/components/stepforms/WidgetInputText.vue';
+import WidgetList from '@/components/stepforms/WidgetList.vue';
+import WidgetToReplace from '@/components/stepforms/WidgetToReplace.vue';
+import BaseStepForm from './StepForm.vue';
+import { ReplaceStep } from '@/lib/steps';
+import { DataSetColumn } from '@/lib/dataset';
+
+@StepFormComponent({
+  vqbstep: 'replace',
+  name: 'replace-step-form',
+  components: {
+    ColumnPicker,
+    WidgetInputText,
+    WidgetList,
+  },
+})
+export default class ReplaceStepForm extends BaseStepForm<ReplaceStep> {
+  @Prop({ type: Object, default: () => ({ name: 'replace', search_column: '', to_replace: [[]] }) })
+  initialStepValue!: ReplaceStep;
+
+  @Getter columnHeaders!: DataSetColumn[];
+
+  readonly title: string = 'Replace values';
+  widgetToReplace = WidgetToReplace;
+
+  get stepSelectedColumn() {
+    return this.editedStep.search_column;
+  }
+
+  set stepSelectedColumn(colname: string | null) {
+    if (colname === null) {
+      throw new Error('should not try to set null on search_column field');
+    }
+    this.editedStep.search_column = colname;
+  }
+
+  get toReplace() {
+    if (this.editedStep.to_replace.length) {
+      return this.editedStep.to_replace;
+    } else {
+      return [[]];
+    }
+  }
+
+  set toReplace(newval) {
+    this.editedStep.to_replace = [...newval];
+  }
+
+  submit() {
+    const type = this.columnHeaders.filter(h => h.name === this.editedStep.search_column)[0].type;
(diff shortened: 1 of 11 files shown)


src/components/stepforms/ReplaceStepForm.vue
```diff
@@ -0,0 +1,107 @@
...
+  },
+})
+export default class ReplaceStepForm extends BaseStepForm<ReplaceStep> {
+  @Prop({ type: Object, default: () => ({ name: 'replace', search_column: '', to_replace: [[]] }) })
+  initialStepValue!: ReplaceStep;
+
+  @Getter columnHeaders!: DataSetColumn[];
+
+  readonly title: string = 'Replace values';
+  widgetToReplace = WidgetToReplace;
```
adimascio: `widgetToReplace` seems like odd name. I know that we've taken the bad habit to prefix (instead of suffix) widget class names with `widget` but it feels awkward here :/

src/components/stepforms/ReplaceStepForm.vue
```diff
@@ -0,0 +1,107 @@
...
+
+  get defaultToReplace() {
```
adimascio: I would argue in favor of dropping this getter. It doesn't rely on any internal state (neither prop nor data) and is used only once.

src/components/stepforms/ReplaceStepForm.vue
```diff
@@ -0,0 +1,107 @@
...
+      throw new Error('should not try to set null on search_column field');
+    }
+    this.editedStep.search_column = colname;
+  }
+
+  get toReplace() {
+    if (this.editedStep.to_replace.length) {
+      return this.editedStep.to_replace;
+    } else {
+      return [this.defaultToReplace];
```
adimascio: If the getter is removed, this line would become `return [[]];`

src/components/stepforms/ReplaceStepForm.vue
```diff
@@ -0,0 +1,107 @@
...
+
+  set toReplace(newval) {
+    this.editedStep.to_replace = [...newval];
+  }
+
+  submit() {
+    const type = this.columnHeaders.filter(h => h.name === this.editedStep.search_column)[0].type;
+    if (type === 'integer' || type === 'float') {
+      for (const tuple of this.editedStep.to_replace) {
+        if (!isNaN(Number(tuple[0]))) {
```
adimascio: Shouldn't we start to carve out all this cast logic. It's already used in the `fillna` step and will be used in probably other places later.

src/components/stepforms/WidgetToReplace.vue
```diff
@@ -0,0 +1,49 @@
...
+@Component({
+  name: 'widget-to-replace',
+  components: {
+    WidgetInputText,
+  },
+})
+export default class WidgetToReplace extends Vue {
+  @Prop({
+    type: Array,
+    default: ['', ''],
```
adimascio:
```suggestion
    default: () => ['', ''],
```

tests/unit/replace-step-form.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+describe('Replace Step Form', () => {
+  let emptyStore: Store<VQBState>;
+  beforeEach(() => {
+    emptyStore = setupStore({});
+  });
+
+  it('should instantiate', () => {
+    const wrapper = shallowMount(ReplaceStepForm, { store: emptyStore, localVue });
+    expect(wrapper.exists()).to.be.true;
+    expect(wrapper.vm.$data.stepname).equal('replace');
```
adimascio:
```suggestion
    expect(wrapper.vm.$data.stepname).to.equal('replace');
```

tests/unit/widget-to-replace.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  it('should instantiate', () => {
+    const wrapper = shallowMount(WidgetToReplace, { store: emptyStore, localVue });
+    expect(wrapper.exists()).to.be.true;
+  });
+
+  it('should have exactly two WidgetInputText components', () => {
+    const wrapper = shallowMount(WidgetToReplace, { store: emptyStore, localVue });
+    const widgetWrappers = wrapper.findAll('widgetinputtext-stub');
+    expect(widgetWrappers.length).to.eql(2);
```
adimascio:
```
    expect(widgetWrappers.length).to.equal(2);
```

tests/unit/widget-to-replace.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      store: emptyStore,
+      localVue,
+      data: () => {
+        return { toReplace: ['foo', 'bar'] };
+      },
+    });
+    const widgetWrappers = wrapper.findAll('widgetinputtext-stub');
+    expect(widgetWrappers.at(0).props().value).to.eql('foo');
+    expect(widgetWrappers.at(1).props().value).to.eql('bar');
+  });
```
adimascio: Use `.to.equal` here

src/components/stepforms/ReplaceStepForm.vue
vdestraitt: Hum ok but it mirrors the aggregation and filter steps. Every component that we use to feed the WidgetList component is prefixed with `widget`...
If we decide it is not a good naming, I think we should rather open a specific PR to clean the naming everywhere to keep consistency.

adimascio: yes, I did not mean to change it now. It will go in a specific PR

src/lib/helpers.ts
```diff
@@ -0,0 +1,29 @@
+import { DataSetColumnType } from './dataset';
+
+function issStringBoolean(string: string): boolean {
```
adimascio: looks like a typo here. You probably meant `isStringBoolean`

src/lib/helpers.ts
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    string === 'true' ||
```
adimascio: just lowercase the string and then compare to `true` / `false`. It will catch strings like `trUe` but is that really a problem?

tests/unit/helpers.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(castFromString(string4, 'boolean')).toEqual(true);
+    expect(castFromString(string5, 'boolean')).toEqual(false);
+    expect(castFromString(string6, 'boolean')).toEqual(false);
+    expect(castFromString(string7, 'boolean')).toEqual(false);
+    expect(castFromString(string8, 'boolean')).toEqual(false);
+  });
+
+  it('should not cast a string that does not convert to boolean type', () => {
+    const string = 'FaLsE';
+    expect(castFromString(string, 'boolean')).toEqual('FaLsE');
```
adimascio: Ok, dismissing my previous comment, it's a choice then

PR #267, opened 2019-07-28
[pagination] provide a simple pagination component

adimascio:
Add extra information to state to monitor current pagination context and wrap [vuejs-paginate](https://www.npmjs.com/package/vuejs-paginate)  to build a visual navigation bar

**NOTE** this PR should be merged into https://github.com/ToucanToco/vue-query-builder/pull/263

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -21,14 +21,24 @@ class MongoService {
     return response.json();
   }
 
-  async executePipeline(pipeline, limit) {
+  async executePipeline(pipeline, limit, offset = 0) {
     const { domain, pipeline: subpipeline } = filterOutDomain(pipeline);
     const query = this.translator.translate(subpipeline);
-    if (limit) {
-      query.push({ '$limit': limit })
+    // first offset
+    if (offset) {
+      query.push({ $skip: offset });
     }
-    const rset = await this.executeQuery(query, domain);
+    // then limit
+    if (limit) {
+      query.push({ $limit: limit });
+    }
+    const [{ count, data: rset }] = await this.executeQuery(query, domain);
     const dataset = mongoResultsToDataset(rset);
+    dataset.paginationContext = {
+      totalCount: count,
+      pagesize: limit,
+      pageno: Math.floor(offset / limit),
+    }
     const datasetWithInferedType = inferTypeFromDataset(dataset);
     return datasetWithInferedType;
   }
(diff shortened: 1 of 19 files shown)


playground/app.js
```diff
@@ -89,6 +99,7 @@ async function buildVueApp() {
     {
       pipeline: initialPipeline,
       currentDomain: 'test-collection',
+      pagesize: 2,
```
vdestraitt: We can remove this row to fall back to the default `pagesize` (`50` as of today) ?

src/components/Pagination.vue
```diff
@@ -0,0 +1,81 @@
...
+    this.setCurrentPage({ pageno });
+  }
+}
+</script>
+<style lang="scss">
+.pagination {
+  text-align: center;
+}
+
+.pagination__list {
```
vdestraitt: Can we reduce the size of the pagination list, which is quite heavy compared to any dataviewer row ?
A simple `font-size: 10px;`seems to do the trick...

src/components/Pagination.vue
```diff
@@ -0,0 +1,82 @@
...
+}
+</script>
+<style lang="scss">
+.pagination {
+  text-align: center;
+  font-size: 10px;
+}
+
+.pagination__list {
+  display: inline-block;
```
zegonz: Here you should use `display flex` property and `justify-content: center` instead of display: inline-block that will horizontal align children element (see below my other comments)

src/components/Pagination.vue
```diff
@@ -0,0 +1,82 @@
...
+  margin: 20px 0;
+  border-radius: 4px;
+}
+
+.pagination__list li {
+  display: inline;
+}
+
+.pagination__list li a {
+  position: relative;
```
zegonz: `position: relative;` doesn't seems usefull on this element

src/components/Pagination.vue
```diff
@@ -0,0 +1,82 @@
...
+
+.pagination__list {
+  display: inline-block;
+  padding-left: 0;
+  margin: 20px 0;
+  border-radius: 4px;
+}
+
+.pagination__list li {
+  display: inline;
```
zegonz: with display flex you will not need display: inline but you have to use `list-style: none;` to hide the bullet list style

PR #272, opened 2019-07-30
Fix Filter Step :: Cast the value key of the filter conditions depending on column type

vdestraitt:


src/components/stepforms/FilterStepForm.vue
```diff
@@ -65,6 +70,24 @@ export default class FilterStepForm extends BaseStepForm<FilterStep> {
   set conditions(newval) {
     this.editedStep.condition.and = [...newval];
   }
+
+  submit() {
+    for (const cond of this.editedStep.condition.and as FilterSimpleCondition[]) {
+      const type = this.columnHeaders.filter(h => h.name === cond.column)[0].type;
```
adimascio: This is getting a more and more common operation. Shouldn't we have a helper somewhere or a getter on the store?

src/components/stepforms/FilterStepForm.vue
```diff
@@ -65,6 +70,24 @@ export default class FilterStepForm extends BaseStepForm<FilterStep> {
...
+      if (type !== undefined) {
+        if (Array.isArray(cond.value)) {
+          const castArray = [];
+          for (const item of cond.value) {
+            castArray.push(castFromString(item, type));
+            cond.value = [...castArray];
```
adimascio: I think this shoud be outside the for-loop.

What about:

```typescript
cond.value = cond.value.map(v => castFromString(v, type));
```
?

PR #274, opened 2019-07-31
[getters] introduce the columnTypes getter

adimascio:
`columnTypes` is the `columnName` → `columnType` map for the current dataset.

This will avoid lines as below to proliferate:
```
const type = this.columnHeaders.filter(h => h.name === this.editedStep.search_column)[0].type;
```

since we can now use:
```
const type = this.columnTypes[this.editedStep.search_column];
```

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -1,6 +1,7 @@
 /**
  * exports the list of store getters.
  */
+import _ from 'lodash';
 
 import { VQBState, activePipeline, inactivePipeline } from './state';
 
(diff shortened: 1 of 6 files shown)


src/store/getters.ts
```diff
@@ -1,6 +1,7 @@
 /**
  * exports the list of store getters.
  */
+import _ from 'lodash';
```
PrettyWood: Je te suggere de faire plutot
```
import fromPairs from 'lodash/fromPairs';
```
https://medium.com/@armno/til-importing-lodash-into-angular-the-better-way-aacbeaa40473#e748

PR #276, opened 2019-08-02
Feature :: catch backend error

rhuille:
# Descrition
**Before:** when a backend error occurred nothing happens. The user had no feedback or debug message.
**Now:** we catch backend error and store the last error message in the vuex store.

# What does this PR introduce ?
To do so, we introduce a new backend response format call `BackendResponse` and implemented in `lib/backend-response.ts`. 

**Before:** There was two possible backend responses: 
- list of string --> from `listCollection`
- Dataset --> from `executePipeline`

**Now:** All backend responses have the same "meta" format. It must be a object with two fields uncompatible fields:
   - `data` can be of type `DataSet` (from `executePipeline`) or list of string (from `executePipeline`)
   - `error` must be of type `ErrorResponse`

If the response has field `'error'` we now commit in the store a new state properties: `backendErrorMessage`.
NB : if no error, then `backendErrorMessage` is `null`

Finally we use this new properties to display the message in the template. Need design for the integration.

⚠️ this PR is not tested yet

# Screenshot and additional info

vue-query-builder

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -12,7 +12,6 @@ const {
 const mongo36translator = getTranslator('mongo36');
 
 class MongoService {
-
   async listCollections() {
     const response = await fetch('/collections');
     return response.json();
@@ -29,15 +28,25 @@ class MongoService {
     if (limit) {
       query.push({ $limit: limit });
     }
-    const [{ count, data: rset }] = await this.executeQuery(query, domain);
-    const dataset = mongoResultsToDataset(rset);
-    dataset.paginationContext = {
-      totalCount: count,
-      pagesize: limit,
-      pageno: Math.floor(offset / limit) + 1,
+    const { isResponseOk, responseContent } = await this.executeQuery(query, domain);
+
+    if (isResponseOk) {
+      const [{ count, data: rset }] = responseContent;
+      const dataset = mongoResultsToDataset(rset);
+      dataset.paginationContext = {
+        totalCount: count,
+        pagesize: limit,
+        pageno: Math.floor(offset / limit) + 1,
+      };
+      const datasetWithInferedType = inferTypeFromDataset(dataset);
+      return {
+        data: datasetWithInferedType,
+      };
+    } else {
+      return {
+        error: responseContent.errmsg,
+      };
     }
-    const datasetWithInferedType = inferTypeFromDataset(dataset);
-    return datasetWithInferedType;
   }
 
   async executeQuery(query, collection) {
@@ -51,7 +60,10 @@ class MongoService {
         'Content-Type': 'application/json',
       },
     });
-    return response.json();
+    return {
+      isResponseOk: response.ok,
+      responseContent: await response.json(),
+    };
   }
 
   async loadCSV(file) {
@@ -85,9 +97,14 @@ async function setupInitialData(store, domain = null) {
       currentDomain: domain,
     });
   } else {
-    const dataset = await mongoservice.executePipeline(store.state.pipeline, store.state.pagesize);
-    store.commit('setDataset', { dataset });
+    const response = await mongoservice.executePipeline(store.state.pipeline, store.state.pagesize);
+    if (response.error) {
+      store.commit('setBackendErrorMessage', { backendErrorMessage: response.error });
+    } else {
+      store.commit('setDataset', { dataset: response.data });
+    }
   }
+  store.commit('setDataset', { dataset: response.data });
 }
 
 async function buildVueApp() {
@@ -96,7 +113,6 @@ async function buildVueApp() {
     {
       pipeline: initialPipeline,
       currentDomain: 'test-collection',
-
     },
     [mongoBackendPlugin],
   );
--- a/playground/app.js
+++ b/playground/app.js
@@ -99,7 +99,7 @@ async function setupInitialData(store, domain = null) {
   } else {
     const response = await mongoservice.executePipeline(store.state.pipeline, store.state.pagesize);
     if (response.error) {
-      store.commit('setBackendErrorMessage', { backendErrorMessage: response.error });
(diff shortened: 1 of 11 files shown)


src/store/backend-plugin.ts
```diff
@@ -31,16 +31,29 @@ export interface BackendService {
...
-    pageOffset(store.state.pagesize, store.getters.pageno),
-  );
-  store.commit('setDataset', { dataset });
+  try {
+    const reponse = await service.executePipeline(
+      pipeline,
+      store.state.pagesize,
+      pageOffset(store.state.pagesize, store.getters.pageno),
+    );
+    if(reponse.meta == 'data'){
```
adimascio: missing space before curly braces (at several locations)

src/store/backend-plugin.ts
```diff
@@ -31,16 +31,29 @@ export interface BackendService {
...
+      const dataset = reponse.data
+      store.commit('setDataset', { dataset });
+      // restore message error to null:
+      store.commit('setbackendErrorMessage',  { backendErrorMessage: null });
+    }else{
+      const backendErrorMessage = reponse.error
```
adimascio: Don't bother creating an intermediate `backendErrorMessage` variable here.

src/store/backend-plugin.ts
```diff
@@ -31,16 +31,29 @@ export interface BackendService {
...
+      store.commit('setbackendErrorMessage', { backendErrorMessage });
+    }
+  } catch(error){
+    const backendErrorMessage = { type: 'error', message: error }
```
adimascio: same here

src/store/mutations.ts
```diff
@@ -6,6 +6,11 @@ import { DomainStep } from '@/lib/steps';
 import { VQBState } from './state';
 
 // provide types for each possible mutations' payloads
+type BackendErrorMutation = {
+  type: 'setbackendErrorMessage';
```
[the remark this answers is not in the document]
rhuille: Yes ! Done

src/store/backend-plugin.ts
rhuille: Corrected

src/store/getters.ts
```diff
@@ -57,4 +57,9 @@ export default {
    */
   pageno: (state: VQBState) =>
     state.dataset.paginationContext ? state.dataset.paginationContext.pageno : 1,
+  /**
+   * Return current page number
```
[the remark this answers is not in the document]
rhuille: Oups, sorry I corrected it

Dalanir:
@adimasci @vdestraitt @rhuille On avait pas décider de déléguer l'affichage au projet qui utilisent le VQB ?

rhuille:
Yes @Dalanir your are right... my bad... I juste corrected it :)

playground/app.js
```diff
@@ -88,9 +101,15 @@ async function setupInitialData(store, domain = null) {
...
-    const dataset = await mongoservice.executePipeline(store.state.pipeline, store.state.pagesize);
-    store.commit('setDataset', { dataset });
+    const response = await mongoservice.executePipeline(store.state.pipeline, store.state.pagesize);
+    if (response.error) {
+      store.commit('setBackendErrorMessage', { backendErrorMessage: response.error });
+    } else {
+      store.commit('setDataset', { dataset: response.data });
+    }
   }
+  const dataset = response.data;
```
Dalanir: to remvove ?

playground/app.js
```diff
@@ -141,6 +141,12 @@ async function buildVueApp() {
         const query = mongo36translator.translate(this.$store.getters.activePipeline);
         return JSON.stringify(query, null, 2);
       },
+      isABackendError: function() {
+        return this.$store.getters.isABackendError;
+      },
+      backendErrorMessage: function() {
+        return this.$store.state.backendErrorMessage.message;
```
Dalanir: A getter to get the message would be cool rather than access directly the state

src/components/Vqb.vue
```diff
@@ -49,8 +49,10 @@ export default class Vqb extends Vue {
   @State isEditingStep!: boolean;
   @State pagesize!: number;
   @State dataset!: DataSet;
+  @State backendErrorMessage!: Error | null;
```
Dalanir: It's not used here

src/components/Vqb.vue
```diff
@@ -49,8 +49,10 @@ export default class Vqb extends Vue {
...
 
   @Getter computedActiveStepIndex!: number;
+  @Getter isABackendError!: boolean;
```
Dalanir: It is not used here

src/lib/backend-response.ts
```diff
@@ -0,0 +1,14 @@
+type ErrorType = 'error' | 'warning';
+
+export type Error = {
```
Dalanir: It is really used outside ?

src/lib/backend-response.ts
```diff
@@ -0,0 +1,14 @@
...
+  type: ErrorType;
+  message: string;
+};
+
+export interface ErrorResponse {
```
Dalanir: same

playground/server.js
```diff
@@ -179,14 +182,12 @@ function setupApp(config) {
...
-      client,
-      req.body.collection,
-      req.body.query,
-      res.json.bind(res),
-      console.error,
-    );
+    executeQuery(config, client, req.body.collection, req.body.query, res.json.bind(res), function(
+      err,
+    ) {
+      res.status(400).send(err);
```
Dalanir: Don't we need a if statement to check if we have to send `err` ?

Dalanir: My bad I understand now !

PR #278, opened 2019-08-05
Feature :: pagination counter (clean branch)

zegonz:
add a counter to pagination component with three informations :
- page current min row => first row number which is displayed in the current page
- page current max row => last row number which is displayed in the current page
- row total rows 

At the last page `page current max row` is hidden.

image

### DIFF
--- a/src/lib/dataset/pagination.ts
+++ b/src/lib/dataset/pagination.ts
@@ -24,3 +24,24 @@ export function numberOfPages(paginationContext: PaginationContext) {
 export function pageOffset(pagesize: number, pageno: number) {
   return Math.max(pageno - 1, 0) * pagesize;
 }
+
+/**
+ * Get page min/max row
+ */
+export function pageMinMax(paginationContext: PaginationContext) {
+  const pageRowMin = pageOffset(
+    paginationContext.pagesize,
+    paginationContext.pageno,
+  );
+  let pageRowMax;
+  if (paginationContext.pageno === numberOfPages(paginationContext)) {
+    pageRowMax = paginationContext.totalCount
+  } else {
+    pageRowMax = pageRowMin + paginationContext.pagesize;
+  }
+  const pageRows = {
+    min: pageRowMin + 1,
+    max: pageRowMax,
+  }
+  return pageRows;
+}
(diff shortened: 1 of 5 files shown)


src/lib/dataset/pagination.ts
```diff
@@ -24,3 +24,19 @@ export function numberOfPages(paginationContext: PaginationContext) {
...
+
+/**
+ * Get page min/max row
+ */
+export function pageMinMax(paginationContext: PaginationContext) {
+  const pageRowMin = pageOffset(
+    paginationContext.pagesize,
+    paginationContext.pageno + 1,
+  ) + 1;
+  const pageRowMax = paginationContext.pageno === paginationContext.pagesize ? paginationContext.totalCount : pageRowMin + (paginationContext.pagesize - 1);
```
adimascio: This line feels a bit long-ish. I think a more traditional `if` / `else` would be more readable. Besides, I have serious doubts on the test being performed here: `pageno === pagesize`. What does `pageno` and `pagesize` have to do together?

tests/unit/pagination.spec.ts
```diff
@@ -24,4 +24,13 @@ describe('pagination tests', () => {
...
     expect(P.pageOffset(20, 2)).toEqual(20);
   });
+
+  it('should return corerctly the min and max row of a page offset', () => {
+    const context: P.PaginationContext = {
+      pageno: 1,
+      pagesize: 20,
+      totalCount: 208,
+    };
+    expect(P.pageMinMax(context)).toEqual({ max: 40, min: 21 });
```
adimascio: Could you add tests for borders (first page and last page)? Also, I believe this test should return `{min: 1, max: 20}` since `pageno` starts at `1`, not `0`.

zegonz: Ok for adding tests to borders page but pageno really starts at 1 ;)

src/lib/dataset/pagination.ts
zegonz: yeah you're right it was a mistake thx to spot it !
I will improve the readability as well

zegonz:
I added your feedback and clean the history, thank you for your review @adimasci

PR #279, opened 2019-08-06
Make formula step accessible from compute widget

vdestraitt:
Make formula step accessible from compute widget and from search bar
Therefore we get rid of the widget '+ column' which was less
intuitive and took some unnecessary space in the toolbar.
At the same time, we propose an enhancement of the SEARCH_BAR
constant definition to reflect its dependancy on the
ACTION_CATEGORIES constant definition

### DIFF
--- a/src/components/constants.ts
+++ b/src/components/constants.ts
@@ -23,7 +23,10 @@ export type groupActions = {
 };
 
 export const ACTION_CATEGORIES: ActionCategories = {
-  compute: [{ name: 'percentage', label: 'Percentage of total' }],
+  compute: [
+    { name: 'formula', label: 'Formula' },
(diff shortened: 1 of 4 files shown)


src/components/constants.ts
```diff
@@ -37,19 +40,12 @@ export const ACTION_CATEGORIES: ActionCategories = {
 
 export const SEARCH_ACTION: groupActions[] = [
   {
-    type: 'compute',
-    actions: [{ name: 'percentage', label: 'Percentage of total' }],
+    type: 'filter',
+    actions: [...ACTION_CATEGORIES.filter],
```
zegonz: Really nice improvement 👍

PR #281, opened 2019-08-07
Feature :: Improve display error message

### LINKED ISSUE #181: Better error handling on step edition
When errors are raised on step validation, we should highlight problematic fields (e.g. the `newname` field of a `rename`) and put the error message above or below the field

zegonz:
To avoid many modifications into each form step we use the mixins : `formWidget`. This mixins compute, the field message error and classError by matching right error with `dataPath` props pass by its step form parent component.

For `filterStepForm` we needed to change his schema to remove in and and conditions operator (for the moment) otherwise there was too many errors to display.

image

image

image

closes #181

### DIFF
--- a/src/components/stepforms/schemas/filter.ts
+++ b/src/components/stepforms/schemas/filter.ts
@@ -1,94 +1,3 @@
-const singleValueConditionSchema = {
-  type: ['string', 'number', 'boolean', 'null'],
-  minLength: 1,
-  title: 'Value',
-  description: 'The value to compare',
-  attrs: {
-    placeholder: 'Enter a value',
-  },
-};
-
-const multipleValueConditionSchema = {
-  type: 'array',
-  minItems: 1,
-  items: {
-    type: ['string', 'number', 'boolean', 'null'],
-  },
-  title: 'Value',
-  description: 'The value(s) to compare',
-  attrs: {
-    placeholder: 'Enter a value',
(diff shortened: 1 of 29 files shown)


src/components/stepforms/schemas/filter.ts
```diff
@@ -100,7 +100,52 @@ export default {
     },
     condition: {
       type: 'object',
-      oneOf: [simpleConditionSchema, complexConditionSchema],
+      properties: {
```
adimascio: I think we should remove, for now, the "dead" code (`multipleValueConditionSchema`, etc.)

tests/unit/form-widget.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { shallowMount } from '@vue/test-utils';
```
adimascio: You could move this squash this commit with the one introducing the `FormWidget` mixin

src/components/stepforms/schemas/filter.ts
zegonz: I agree with you but I thought it was better to keep the useless code for reuse it later, well ok it's quite dirty ;)

zegonz: I thought it was better to keep the useless code for reuse it later... well ok it's quite dirty ;)

PR #307, opened 2019-09-17
Fix :: change offset computing method to get the good align right position

zegonz:
closes this issue : https://github.com/ToucanToco/vue-query-builder/issues/303

now when the popover is out of screen :
image

### DIFF
--- a/src/components/domutil.ts
+++ b/src/components/domutil.ts
@@ -82,10 +82,10 @@ function completeClientRect(customRect: Partial<ClientRect> = {}): ClientRect {
  */
 function completeOffset(customOffset: Partial<ElementOffset> = {}): ElementOffset {
   const defaultOffset = {
-    offsetWidth: 0,
-    offsetHeight: 0,
+    offsetWidth: customOffset.offsetWidth || 0,
+    offsetHeight: customOffset.offsetHeight || 0,
   };
-  return { ...defaultOffset, ...customOffset };
+  return defaultOffset;
 }
 
 /**


zegonz:
> Could you explain a bit more where the exact problem is in:
> 
> * the commit message,
> * maybe a unit test exposing the failing behaviour.
> 
> I see subtle semantic differences in the two implementations (i.e. before / after), especially if completeOffset`contains`null`or`undefined` values but I'm not sure what the exact error was.


@adimasci well there was no error ! the problem was that we didn't get the correct position for `alignRight` function. the popover offsetWidth size was always equal to zero.

Before we  forced `offsetWidth: 0` but it wasn't the element size, that's why we use `offsetWidth: customOffset.offsetWidth || 0` so in `function alignRight` the variable `let left = parentRight - element.offsetWidth;` can compute the correct left position with the correct popover offset width. 
Actually i'm not sure that we need a test for that because in all popover test we pass directly the offset position.

I will change the commit name anyway and add more description to the PR

Is it clearer for you ?

adimascio:
> Is it clearer for you ?

Well, not that much. We did not force `offsetWidth` to 0, that's my point :)

adimascio:
> Well, not that much. We did not force `offsetWidth` to 0, that's my point :)

OK. The problem seems to that those properties are not enumerable and thus not destructrued in the `...` expression

adimascio:
> OK. The problem seems to that those properties are not enumerable and thus not destructrued in the `...` expression

I've fixed the commit and added a link to the issue closed in the commit message.

Issue #300, opened 2019-09-16
Preselect column on contextual transform operations


adimascio:
As a user, when I use a column's contextual dropdown, I expect the column to be preselected.

For intsance, if I click on the dropdown and choose "filter", I expect the "column" field in the edition form to be pre-filled with the column name.

Note : this issue is not about general edition / table click interaction.

PR #310, opened 2019-09-20
[FilterStepForm] Fix :: Set column value with current selected column

Dalanir:
closes # #300


adimascio:
@Dalanir I've pushed a commit as an update proposal to:

- fix the tests,
- add a comment to the `mounted()` logic

Dalanir:
@adimascio Thank you, seems good to me if you want to submit it to review or review it yourself too !

adimascio:
> @Dalanir @adimasci beware this implementation introduces a bug:
> => The filtered column is now `Column2` and not `Column1` which is not expected.

Should be fixed now, thanks!

Dalanir:
LGTM too !

PR #316, opened 2019-09-29
Variable interpolation in pipeline steps

### LINKED ISSUE #320: Allow parametrized queries
As a query conceptor, I would like to be able to use templated strings rather than hardcoded values. 

Using [lodash template syntax](https://lodash.com/docs/4.17.15#template), this
could look like:

```json
[
  {
    "name": "domain",
    "domain": "team_infos",
  },
  {
    "name": "filter",
    "condition": {
      "and": [
        "operator": "gt",
        "column": "age",
        "value": "<%= someobj.age %>"
      ]
    }
  }
]
```

adimascio:
Implement parametrized queries.

Closes #320

### DIFF
--- a/src/lib/steps.ts
+++ b/src/lib/steps.ts
@@ -75,7 +75,7 @@ export type FilterSimpleCondition =
 
 type FilterConditionComparison = {
   column: string;
-  value: number;
+  value: number | string;
   operator: 'gt' | 'ge' | 'lt' | 'le';
 };
 
--- a/src/lib/steps.ts
+++ b/src/lib/steps.ts
@@ -3,6 +3,7 @@
  */
 
 type PrimitiveType = number | boolean | string | Date;
+type Templatable<T> = T | string;
 
 export type AggFunctionStep = {
   /** Name of the output column */
@@ -149,7 +150,7 @@ export type TopStep = {
   groups?: string[];
   rank_on: string;
   sort: 'asc' | 'desc';
-  limit: number;
+  limit: Templatable<number>;
 };
 
 export type UnpivotStep = {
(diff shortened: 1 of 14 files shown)


src/lib/templating.ts
```diff
@@ -0,0 +1,160 @@
...
+}
+
+/**
+ * The `Interpolator` class provides an "step interpolator" implementation
+ * for each possible step in a pipeline.
+ * A step interpolator is a function that takes a step in parameter and returns
+ * the "interpolated" version of a step, given a specific interpolate function (e.g.
+ * `_.template`).
+ */
+export class PipelineInterpolator implements StepMatcher<S.PipelineStep> {
```
vdestraitt: Ok with the list of steps and parameters supporting templating for a first version! Thanks

tests/unit/top-step-form.spec.ts
```diff
@@ -74,6 +76,39 @@ describe('Top Step Form', () => {
     });
   });
 
+  it('should accept templatable values', async () => {
```
vdestraitt: Should we add this kind of test in every step test files accepting templating ?

adimascio: Maybe but I think all steps (except the `Top` one) already accept strings so there's nothing much to test.

src/lib/steps.ts
```diff
@@ -75,7 +76,7 @@ export type FilterSimpleCondition =
 
 type FilterConditionComparison = {
   column: string;
-  value: number;
+  value: number | string;
```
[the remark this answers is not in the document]
adimascio: My idea was more to authorize strings natively, not specifically templates. e.g. to filter values `> 'a'`. I even think that we might allow dates. What do you think?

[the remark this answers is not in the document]
vdestraitt: Ok agreed

tests/unit/top-step-form.spec.ts
vdestraitt: Oook understood

PR #318, opened 2019-09-30
Simplify duplicate step

vdestraitt:
Do not ask for a new column name. We choose it for the user, then
the user can apply a rename step if needed. This makes the
workflow more direct (no need to fill a form at step creation)

This PR proposes to introduce a helper function to enforce valid column names when naming is automatic.

Closes https://github.com/ToucanToco/vue-query-builder/issues/314

### DIFF
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -27,3 +27,21 @@ export function castFromString(value: string, type: DataSetColumnType) {
   }
   return value;
 }
+
+/**
+ * Takes a proposed new column name and returns a valid name that does not erase
+ * an existing column name. So if the proposed new name already exist in
+ * existing names, it appends an integer that increments until the new name
+ * becomes unique
+ *
+ * @param newName the proposed new column name
+ * @param existingNames the existing columns names
+ */
+export function newColumnName(newName: string, existingNames: string[]): string {
+  let i = 1;
+  let validNewName = newName;
+  while (existingNames.includes(validNewName)) {
+    validNewName = `${newName}${i++}`;
+  }
+  return validNewName;
+}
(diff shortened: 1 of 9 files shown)


src/lib/helpers.ts
```diff
@@ -27,3 +27,21 @@ export function castFromString(value: string, type: DataSetColumnType) {
   }
   return value;
 }
+
+/**
+ * Takes a proposed new column name and returns a valid name that does not erase
+ * an existing column name. So if the proposed new name already exist in
```
adimascio:
```suggestion
  * an existing column name. So if the proposed new name already exist in
```

adimascio:
I'm a bit skeptical about this. Make Forccing the user to perform 2 steps to duplicate + rename seems awkward

src/lib/helpers.ts
```diff
@@ -27,3 +27,21 @@ export function castFromString(value: string, type: DataSetColumnType) {
...
+ * existing names, it appends an integer that increments until the new name
+ * becomes unique
+ *
+ * @param newName the proposed new column name
+ * @param existingNames the existing columns names
+ */
+export function newColumnName(newName: string, existingNames: string[]): string {
```
Dalanir: I think a name with an action verb would be interesting, like `generateColumnName` or something like that

vdestraitt:
@adimascio it's inspired by other software but the thing is:
If we think that in most cases we need the rename afterward (in my opinion we do) I agree that we do not offer any simplification as our rename requires to fill a form!
For comparison, in other software products, the rename consists in double-clicking the column header and entering a new name so the UX "duplicate + rename" is simpler and quicker.

So... I close this PR as long as we do not offer a more straightforward interaction with the table for renaming a column :)

PR #321, opened 2019-09-30
Simplify unpivot step

vdestraitt:
Do not require to specify the value_column_name or category_column_name.
The meaning of those parameters was not clear.
We can enforce a name to make the UX more straightforward, and if needed the user can apply a rename step.

Closes https://github.com/ToucanToco/vue-query-builder/issues/304

### DIFF
--- a/tests/unit/helpers.spec.ts
+++ b/tests/unit/helpers.spec.ts
@@ -1,4 +1,4 @@
-import { castFromString } from '@/lib/helpers';
+import { castFromString, generateNewColumnName } from '@/lib/helpers';
 
 describe('castFromString', () => {
   it('should cast numeric string to number type', () => {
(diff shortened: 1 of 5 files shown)


tests/unit/unpivot-step-form.spec.ts
```diff
@@ -206,8 +98,8 @@ describe('Unpivot Step Form', () => {
           name: 'unpivot',
           keep: ['columnA', 'columnB'],
           unpivot: ['columnC'],
-          unpivot_column_name: 'foo',
-          value_column_name: 'bar',
+          unpivot_column_name: '',
+          value_column_name: '',
```
Dalanir: To remove completely no ?

tests/unit/helpers.spec.ts
```diff
@@ -37,4 +37,16 @@ describe('castFromString', () => {
     const string = 'FaLsE';
     expect(castFromString(string, 'boolean')).toEqual('FaLsE');
   });
+
+  describe('newColumName', () => {
```
Dalanir:
```suggestion
  describe('generateNewColumName', () => {
```

PR #322, opened 2019-10-01
Fix :: Firefox hack for dataviewer border display

zegonz:
with background-clip: padding-box property hack it resolves the hidden border issue due to border collapse on FF. It was due to  the combination of `border-collapse: collapse` and a background-color on `td` element, it's actually a old known bug on FF (not so easy to find this trick...).

Now on FF it's the same display as in Chrome:
image


Dalanir:
Oh yeah thanks @zegonz !

PR #326, opened 2019-10-02
Feature :: concatenate step

vdestraitt:
Introduce a `concatenate` step translated into mongo.
Allows concatenating several columns using a separator of the user's choice.

Closes https://github.com/ToucanToco/vue-query-builder/issues/340

### DIFF
--- a/src/components/stepforms/ColumnPicker.vue
+++ b/src/components/stepforms/ColumnPicker.vue
@@ -40,6 +40,11 @@ export default class ColumnPicker extends Vue {
   @Prop({ default: null })
   dataPath!: string;
 
+  // Whether the column data of ColumnPicker should react to a change of
+  // selected column
+  @Prop({ default: true })
+  syncWithSelectedColumn!: boolean;
+
   // Only manage the deletion of 1 column at once at this stage
   column: string | null = null;
 
@@ -69,7 +74,7 @@ export default class ColumnPicker extends Vue {
 
   @Watch('selectedColumns')
   onSelectedColumnsChanged(val: string[], oldVal: string[]) {
-    if (!_.isEqual(val, oldVal)) {
+    if (!_.isEqual(val, oldVal) && this.syncWithSelectedColumn) {
       this.column = val[0];
     }
   }
--- a/src/components/stepforms/ConcatenateStepForm.vue
+++ b/src/components/stepforms/ConcatenateStepForm.vue
@@ -1,0 +1,91 @@
+<template>
+  <div>
+    <step-form-title :title="title"></step-form-title>
+    <ListWidget
+      addFieldName="Add columns"
+      id="toConcatenate"
+      name="Columns to concatenate:"
+      v-model="toConcatenate"
+      :defaultItem="[]"
+      :widget="columnPicker"
+      :componentProps="{ syncWithSelectedColumn: false }"
+      :automatic-new-field="false"
+      data-path=".columns"
+      :errors="errors"
+    ></ListWidget>
+    <InputTextWidget
+      id="separator"
+      v-model="editedStep.separator"
+      name="Separator:"
+      placeholder="Enter string of any length"
+      data-path=".separator"
+      :errors="errors"
+    ></InputTextWidget>
+    <InputTextWidget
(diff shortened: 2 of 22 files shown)


src/components/stepforms/ColumnPicker.vue
```diff
@@ -41,6 +40,9 @@ export default class ColumnPicker extends Vue {
   @Prop({ default: null })
   dataPath!: string;
 
+  @Prop({ default: true })
+  sync!: boolean;
```
adimascio: Maybe rename to `syncWithSelectedColumn`? Anyhow, add a docstring to this attribute.

src/components/stepforms/ConcatenateStepForm.vue
```diff
@@ -0,0 +1,92 @@
...
+})
+export default class ConcatenateStepForm extends BaseStepForm<ConcatenateStep> {
+  @Prop({
+    type: Object,
+    default: () => ({ name: 'concatenate', columns: [''], separator: '', new_column_name: '' }),
+  })
+  initialStepValue!: ConcatenateStep;
+
+  readonly title: string = 'Concatenate columns';
+  columnPicker = ColumnPicker;
```
adimascio: Should that be a prop?

src/components/stepforms/ConcatenateStepForm.vue
```diff
@@ -0,0 +1,92 @@
...
+      // Otherwise, fallback on the default initial value
+      this.editedStep = { ...this.initialStepValue };
+    }
+  }
+
+  get toConcatenate() {
+    if (this.editedStep.columns.length) {
+      return this.editedStep.columns;
+    } else {
+      // return this.defaultColumns;
```
adimascio: Remove comment

src/components/stepforms/ConcatenateStepForm.vue
```diff
@@ -0,0 +1,92 @@
...
+      return ['', ''];
```
adimascio: Hmm, this sounds hackish, why can't we use `['']` or `[]` directly?

src/components/stepforms/widgets/ColumnSelector.vue
```diff
@@ -0,0 +1,69 @@
+<template>
```
adimascio: I think this whole file should be removed now that we have the `sync` property on `ColumnPicker`.

src/lib/templating.ts
```diff
@@ -83,6 +83,10 @@ export class PipelineInterpolator implements StepMatcher<S.PipelineStep> {
     return { ...step };
   }
 
+  concatenate(step: Readonly<S.ConcatenateStep>) {
+    return { ...step };
```
adimascio: Wouldn't it be worth to make the `separator` field templatable?

tests/unit/column-picker.spec.ts
```diff
@@ -70,4 +70,26 @@ describe('Column Picker', () => {
     expect(wrapper.vm.$data.column).toEqual('columnB');
     expect(store.state.vqb.selectedColumns).toEqual(['columnB']);
   });
+
+  it('should not update step when selectedColumn is changed if prop "sync" is false', async () => {
+    const store: Store<any> = setupMockStore({
```
adimascio: I don't think we need to add the `Store<any>` annotation here.

src/components/stepforms/ConcatenateStepForm.vue
vdestraitt: It conforms to how is built every other component with repeatable widget.
I understand it would change the pattern of how we instantiate components with such a property then?
Maybe it's a change that can e proposed in a specific PR?

src/lib/templating.ts
vdestraitt: No it's ok not to make it templatable. Never saw such a edge case

tests/unit/column-picker.spec.ts
vdestraitt: Need to make typescript happy :)
No quick fix in mind. Maybe we could try to propose something more elegant in another PR ?

PR #334, opened 2019-10-05
Feature :: substring step

vdestraitt:
Introduce a `substring` step translated into mongo.
Allows extracting a substring from a text column based on start index and a number of characters to keep.

Closes https://github.com/ToucanToco/vue-query-builder/issues/338


doc/steps.md
```diff
@@ -783,6 +783,89 @@ When sorting on several columns, order of columns specified in `columns` matters
...
 
+### `substring` step
+
+Extract a substring in a string `column`. The substring begins at index
+`start_index` (beginning at 0) and retains a number of characters specified in
+`number_characters` (-1 to get all characters until the end of the string)
+
+```javascript
+{
+  name: 'sustring',
```
adimascio:
```suggestion
  name: 'substring',
```

doc/steps.md
```diff
@@ -783,6 +783,89 @@ When sorting on several columns, order of columns specified in `columns` matters
...
+| some_text | 20    |
+| a_word    | 1     |
+| toucan    | 10    |
+| toco      | 5     |
+
+**Step configuration:**
+
+```javascript
+{
+  name: 'sustring',
```
adimascio:
```suggestion
  name: 'substring',
```

src/lib/steps.ts
```diff
@@ -144,6 +144,13 @@ export type SortStep = {
   columns: SortColumnType[];
 };
 
+export type SubstringStep = {
+  name: 'substring';
+  column: string;
+  start_index: number;
+  number_characters: number;
```
adimascio: I would use `length` instead of `number_characters`. Besides, I'm wondering if we shouldn't use the same semantic as in python and use `start_index`, `stop_index`, make `stop_index` optional and use `$strLenCP` to compute actual stop index if `stop_index` is negative.

What do you think?

tests/unit/substring-step-form.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+  it('should emit "cancel" event when edition is cancelled', async () => {
+    const wrapper = mount(SubstringStepForm, { store: emptyStore, localVue });
+    wrapper.find('.widget-form-action__button--cancel').trigger('click');
+    await localVue.nextTick();
+    expect(wrapper.emitted()).toEqual({
+      cancel: [[]],
+    });
+  });
+
+  it('should reset selectedStepIndex correctly on cancel depending on isStepCreation', () => {
```
adimascio: Is this a test specific to the `substring`step? It looks more like a `QueryBuilder` general test.

src/lib/translators/mongo.ts
```diff
@@ -331,6 +332,47 @@ function transformSplit(step: Readonly<SplitStep>): MongoStep {
   ];
 }
 
+/** transform a 'substring' step into corresponding mongo steps */
+function transformSubstring(step: Readonly<SubstringStep>): MongoStep {
+  // const startIndexMongo: PropMap<number> = {};
```
adimascio: Please remove this comment

src/lib/translators/mongo.ts
```diff
@@ -331,6 +332,47 @@ function transformSplit(step: Readonly<SplitStep>): MongoStep {
...
+            $strLenCP: $$(step.column),
+          },
+          step.end_index,
+        ],
+      };
+
+  const lengthToKeep = {
+    $add: [
+      {
+        $subtract: [posEndIndex, posStartIndex],
```
adimascio: This is probably a detail or a "small optimization" but I would have found it nice if, in case of 2 positive integers, we would have generated a simpler step, rather than nesting `$add / $substract` on known indices.

src/components/stepforms/schemas/substring.ts
```diff
@@ -0,0 +1,38 @@
...
+      title: 'Substring starts at index...',
+      description: 'The index at which starts the substring (begins at 0)',
+      attrs: {
+        placeholder: 'Enter an integer',
+      },
+    },
+    end_index: {
+      type: 'integer',
+      title: 'Number of characters to keep:',
+      description: 'The index at which starts the substring (-1 = last index)',
```
adimascio:
```suggestion
       description: 'The index at which ends the substring (-1 = last index)',
```

doc/steps.md
```diff
@@ -831,90 +831,6 @@ When sorting on several columns, order of columns specified in `columns` matters
 | Label 6 | Group 2 | 5     |
 | Label 4 | Group 2 | 1     |
 
-### `split` step
```
adimascio: Why do you remove the `split` step documentation?

src/lib/translators/mongo.ts
adimascio: OK. Decision → do this in another PR

PR #342, opened 2019-10-06
Make lowercase and uppercase steps more straightforward

vdestraitt:
If a column is selected before clicking on lowercase or uppercase
operations, the step is applied directly without opening the form

Closes https://github.com/ToucanToco/vue-query-builder/issues/341

### DIFF
--- a/tests/unit/action-toolbar-button.spec.ts
+++ b/tests/unit/action-toolbar-button.spec.ts
@@ -1,6 +1,7 @@
 import { mount, createLocalVue } from '@vue/test-utils';
 import ActionToolbarButton from '@/components/ActionToolbarButton.vue';
-import Vuex from 'vuex';
+import Vuex, { Store } from 'vuex';
+import { setupMockStore } from './utils';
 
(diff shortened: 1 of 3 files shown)


tests/unit/action-toolbar-button.spec.ts
```diff
@@ -71,4 +70,106 @@ describe('ActionToolbar', () => {
...
+      });
+      const actionsWrappers = wrapper.findAll('.action-menu__option');
+      await actionsWrappers.at(0).trigger('click');
+      expect(wrapper.emitted().closed).toBeTruthy();
+    });
+  });
+
+  describe('When clicking on the "To uppercase" operation', () => {
+    it('should close any open step form', async () => {
+      const store: Store<any> = setupMockStore({
```
adimascio: Do you really need to specify the `Store<any>` type annotation?

vdestraitt: yes to make typescript happy :)
not sure to find something more elegant quickly... should we merge? :)

PR #345, opened 2019-10-07
feature :: add a separator label to filters form and fix form input wrong sizing

zegonz:
- Fix the input width size issue by adding the border-box: box-sizing property to all element inside the classname `.data-viewer` and `.query-builder`scope.

- Add a label separator `and` between conditions with a specific style to have a better visual UI understanding : now when we add a condition we get an extra margin-left on all widget-list item.

results :
image


zegonz:
> Thank you! It's nice. It remains one integration issue spotted on the screenshot below: I would expect the trash icon to remain on the right of the input text, not on top of it :)
> 
> image

Can you check again by clearing your browser cache.

PR #353, opened 2019-10-09
Feature :: add date operations

vdestraitt:
Add `todate` and `fromdate`

Closes https://github.com/ToucanToco/vue-query-builder/issues/351
Closes https://github.com/ToucanToco/vue-query-builder/issues/352


src/lib/translators/mongo.ts
```diff
@@ -485,6 +485,14 @@ const mapper: StepMatcher<MongoStep> = {
   argmin: transformArgmaxArgmin,
   concatenate: transformConcatenate,
   custom: step => step.query,
+  todate: step => ({
```
adimascio: Let's try to keep transformation methods alphabetaically sorted!

Issue #374, opened 2019-11-25
[ui-step] UI for step append


vdestraitt:
# UI to create step my-step

Corresponding typescript interface: `append`

## Can your step be created from the column header's popover?
No

- is it specific to a column type?  No
- what is the exact list of interactions the user add to perform from the moment he clicks until the form appears ? Click on the widget "Combine" and select "Append datasets"
- which field(s) should be preset in the form? None

## Which category your step belong to?
Combine

## Make sure your step is accessible in the search menu
OK

## Form specificities
Only a checkbox allowing to select one or several datasets to append

PR #375, opened 2019-11-25
Add append step implementation

vdestraitt:
This step allows to append several datasets, referenced as pipelines names in the application store.

We introduce a PipelineDerefencer class that provides an implementation of a `dereference` function for every step, that allows to replace pipeline names referenced in steps, by their corresponding pipeline before translation can be applied.

Closes https://github.com/ToucanToco/weaverbird/issues/374


src/lib/dereferencing.ts
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * The `PipelineDereferencer` class provides a "dereferencing" implementation
+ * for each possible step in a pipeline. It replaces pipeline names (saved in
+ * store) referenced in steps configuration by their corresponding pipelines.
+ *
+ * The `dereference` method takes a step as parameter and returns a new version
+ * of this step, with pipeline names dereferenced.
+ */
+export class PipelineDereferencer implements StepMatcher<S.PipelineStep> {
```
adimascio: Do we really need a a whole class to compute deferencement? It looks like we're killing a fly with a sledgehammer. Wouldn't a simple function like this one be enough?

```typescript
const dereferenced = [];
for (const step of pipeline.steps) {
  let newStep;
  if (step.name === 'append') { newStep = ...; }
  else { newStep = { ...step } };
  dereferenced.push(newStep);
}
```

adimascio: If we do need a whole class, then maybe a "noop" class for the TemplatingClass and this one would be nice to avoid all the "empty" methods that just copy the input step.

PR #381, opened 2019-12-02
Allow formula result to overwrite existing column

vdestraitt:


src/components/stepforms/schemas/formula.ts
```diff
@@ -31,7 +29,3 @@ const schema = {
   required: ['name', 'formula', 'new_column'],
   additionalProperties: false,
 };
-
-export default function buildSchema(form: StepFormType) {
-  return addNotInColumnNamesConstraint(schema, 'new_column', form.columnNames);
```
adimascio: I'm wondering if we shouldn't make the `new_column` optional to overwrite current column and still forbid to overwrite columns otherwise? I fear we might have problem with large datasets with lots of columns: the user might not realize he's overwriting an existing column.

PR #386, opened 2019-12-04
Fix :: Remove usage of flat()

Dalanir:
In order to works under Edge browser, we prefer to use a different way to flat our arrays.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [0.5.1] - 2019-12-04
+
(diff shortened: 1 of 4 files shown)


Dalanir:
[ETQC sur Edge je peux utiliser le VQB](https://trello.com/c/ghuBYHBT/3747-etqc-sur-edge-je-peux-utiliser-le-vqb)

adimascio:
Closing the PR, the branch is still here, tagged and was used to build and upload the v0.5.1 package.

Issue #377, opened 2019-11-29
[ui-step] UI for step join


vdestraitt:
# UI to create step my-step

Corresponding typescript interface: `join`

## Can your step be created from the column header's popover?
No

- is it specific to a column type?  No
- what is the exact list of interactions the user add to perform from the moment he clicks until the form appears ? Click on the widget "Combine" and select "Join datasets"
- which field(s) should be preset in the form? None

## Which category your step belong to?
Combine

## Make sure your step is accessible in the search menu
OK

## Form specificities
Only a checkbox allowing to select one or several datasets to append

PR #390, opened 2019-12-06
Add join step (mongo translation + UI)

adimascio:
Implements join step to join 2 datasets (mongo translation + UI)

Closes  https://github.com/ToucanToco/weaverbird/issues/377


docs/_docs/tech/steps.md
```diff
@@ -623,6 +613,169 @@ Column names must not be escaped. Strings have to be escaped with quotes.
...
+}
+```
+
+**Output dataset:**
+
+| Label   | Value | Country   |
+| ------- | ----- | --------- |
+| Label 2 | 1     | Country 2 |
+| Label 3 | 1     | Country 1 |
+
```
adimascio: This is not a `left outer join`. A `left outer join` would be an union of all rows where unmatched rows would be filled with `null` values.

src/lib/translators/mongo.ts
```diff
@@ -598,6 +599,51 @@ export class Mongo36Translator extends BaseTranslator {
...
+      },
+      {
+        $project: { _vqbJoinKey: 0 },
+      },
+    );
+    return mongoPipeline;
+  }
 }
 mapper['append'] = Mongo36Translator.prototype.transformAppend;
+mapper['join'] = Mongo36Translator.prototype.transformJoin;
```
adimascio: Why don't we name `transformAppend` `append` directly in the class?

src/components/stepforms/JoinStepForm.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <div>
+    <step-form-title :title="title" />
+    <AutocompleteWidget
+      id="rightPipelineInput"
+      v-model="editedStep.right_pipeline"
+      name="Select a dataset to join (as right dataset):"
+      :options="Object.keys(pipelines).filter(p => p !== currentPipelineName)"
```
adimascio: side-note: this is a detail and there's no need to handle this now but there are valid cases where we want to join a dataset with itself.

PR #396, opened 2019-12-16
Enhancement :: Playground Scrolling Hell

Dalanir:
Some linting changes are in it too. Cause before padding/marging were badly handled, we've got scroll on playground, this PR resolves that.

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -11,14 +11,13 @@ const {
   registerModule,
 } = vqb;
 
-const TRANSLATOR = 'mongo40'
+const TRANSLATOR = 'mongo40';
 
 const mongo40translator = getTranslator(TRANSLATOR);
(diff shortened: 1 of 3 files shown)


playground/app.js
```diff
@@ -156,20 +142,11 @@ class MongoService {
...
-},
-{
-  name: 'filter',
-  condition: {
-    and: [{
-      column: 'Groups',
-      value: '<%= groupname %>',
-      operator: 'eq'
-    }]
+const initialPipeline = [
```
adimascio: I think this domain was used to test the scroll itself but that it should be commited. Am I right?

Dalanir: Ah yes my bad, I changed it for testing purpose, I'll modify it

PR #399, opened 2019-12-31
keep history of error messages

adimascio:
This PR introduces two new mutations:

- `logBackendError` to add a backend error to the logbook
- `resetErrorMessages` to empty the logbook

The former `setBackendError` mutation is removed.

Therefore, rather than having a single error message in the store, keep
track of all of them until the `resetBackendErrors` is called.

This is needed because otherwise the backend-plugin would reset a
previous error once a pipeline execution was successful. Since we
sometimes peform several pipeline executions in a row (e.g. if 2
mutations trigger a pipeline execution consecutively).

### DIFF
--- a/src/store/mutations.ts
+++ b/src/store/mutations.ts
@@ -2,14 +2,14 @@
  * exports the list of store mutations.
  */
 
+import { BackendError } from '@/lib/backend-response';
 import { DomainStep, PipelineStepName } from '@/lib/steps';
-
 import { VQBState } from './state';
 
 // provide types for each possible mutations' payloads
 type BackendErrorMutation = {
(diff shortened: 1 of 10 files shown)


src/store/mutations.ts
```diff
@@ -88,6 +88,14 @@ export default {
     state.currentStepFormName = stepName;
     state.stepFormInitialValue = undefined;
   },
+
+  /**
+   * log a backend error message.
+   */
+  logBackendError(state: VQBState, { backendError }: {backendError: BackendError}) {
```
Dalanir:
```suggestion
  logBackendError(state: VQBState, { backendError }: { backendError: BackendError }) {
```

PR #404, opened 2020-01-17
Filter simple condition form :: Visual improvement

ToucanTokar:
Multiselect dropdown enhancement to improve filter simple condition display.

Before the dropdown multiselect tag was pushing the content below : 
image

Now it's positionned in absolute with a scrollbar and max-height : 
image

It also contains other form improvement like better height size and change the old ugly arrow by a font awesome one :
image


zegonz:
[ETQC j'ai un composant multiInputText dans le VQB mieux intégré](https://trello.com/c/DXhuXUpW/12-etqc-jai-un-composant-multiinputtext-dans-le-vqb-mieux-int%C3%A9gr%C3%A9)

PR #405, opened 2020-01-17
tests :: introduce a runner class

adimascio:
Most of the step form tests use the same scenario:

- test that the component can instantiate properly,
- test that it has the expected sub components
- test that validation errors are raised as expeced if some values are missing or don't match
  the specified jsonschema,
- test that the cancel button work.

This PR introduces a new runner class to help factoring out all those common patterns as well as some small `mount` / `shallowMount` helpers with reasonable defaults. Since Jest doesn't provide a class-based architecture, this class is not meant to be subclassed in each test but rather instantiated and its test method called explicitly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
     "mongodb-prebuilt": "^6.5.0",
     "multer": "^1.4.1",
     "node-sass": "^4.11.0",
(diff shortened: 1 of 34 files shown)


Dalanir:
Thanks for this !

Issue #241, opened 2019-07-10
Custom Step Mongo


vdestraitt:
This step form consists inn a text editor allowing to write any custom mongo query (mongo aggregation pipeline)

PR #406, opened 2020-01-21
New Step :: custom step

rhuille:
# New Step: custom step
Please read commit by commit

weaverbird

weaverbird

weaverbird

related to https://github.com/ToucanToco/weaverbird/issues/241

### DIFF
--- a/src/components/stepforms/CustomStepForm.vue
+++ b/src/components/stepforms/CustomStepForm.vue
@@ -38,5 +38,14 @@ export default class CustomStepForm extends BaseStepForm<CustomStep> {
   get name() {
     return `Write a custom ${getTranslator(this.translator).constructor.label} query`;
   }
+
+  validate() {
+    const errors = this.$$super.validate();
+    if (errors !== null) {
+      return errors;
+    }
+    const translatorErrors = getTranslator(this.translator).validate({ ...this.editedStep });
+    return translatorErrors;
+  }
 }
 </script>
--- a/src/components/stepforms/CustomStepForm.vue
+++ b/src/components/stepforms/CustomStepForm.vue
@@ -1,0 +1,42 @@
+<template>
+  <div>
+    <step-form-header :cancel="cancelEdition" :title="title" :stepName="this.editedStep.name" />
+    <label v-if="name">{{ name }}</label>
+    <TextareaWidget
+      v-model="editedStep.query"
+      placeholder="Write your custom mongo here"
+      :errors="errors"
+      data-path=".query"
+    />
+    <step-form-buttonbar :cancel="cancelEdition" :submit="submit" />
+  </div>
+</template>
+
+<script lang="ts">
+import { Prop } from 'vue-property-decorator';
+import { StepFormComponent } from '@/components/formlib';
+import BaseStepForm from './StepForm.vue';
+import { CustomStep } from '@/lib/steps';
+import TextareaWidget from './widgets/TextareaWidget.vue';
+import { getTranslator } from '@/lib/translators';
+import { VQBModule } from '@/store';
+
+@StepFormComponent({
+  vqbstep: 'custom',
+  name: 'custom-step-form',
+  components: { TextareaWidget },
+})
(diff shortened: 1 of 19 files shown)


src/components/stepforms/CustomStepForm.vue
```diff
@@ -0,0 +1,31 @@
...
+@StepFormComponent({
+  vqbstep: 'custom',
+  name: 'custom-step-form',
+})
+export default class CustomStepForm extends BaseStepForm<CustomStep> {
+  @Prop({ type: Object, default: () => ({ name: 'custom', query: "[{\"$match\": {\"domain\": \"test\"}}]" }) })
+  initialStepValue!: CustomStep;
+
+  readonly title: string = 'Custom step';
+  // Will need to be dynnamic depending on backend type in the future
```
adimascio:
```suggestion
   // Will need to be dynamic depending on backend type in the future
```

src/lib/translators/base.ts
```diff
@@ -60,6 +60,11 @@ function unsupported(target: BaseTranslator, propertyKey: S.PipelineStepName, de
  * - a `supports` method that tests the support of a given step name.
  */
 export class BaseTranslator implements StepMatcher<OutputStep> {
+  /**
+   * `label` will be display to the suer
```
adimascio:
```suggestion
   * `label` will be displayed to the user
```

src/lib/translators/base.ts
```diff
@@ -60,6 +60,11 @@ function unsupported(target: BaseTranslator, propertyKey: S.PipelineStepName, de
...
+   */
+  get label(): string { return '' }
```
adimascio: I think we should use a `static` class member for labels.

src/components/stepforms/CustomStepForm.vue
```diff
@@ -35,5 +35,14 @@ export default class CustomStepForm extends BaseStepForm<CustomStep> {
...
     return `Write a custom ${getTranslator(this.translator).label} query`;
   }
+
+  validate() {
+    const errors = this.$$super.validate();
+    if (errors !== null) {
+      return errors;
+    }
+    const translatorErrors = getTranslator(this.translator).validate({ ...this.editedStep })
+    return translatorErrors
```
adimascio:
```suggestion 
   return translatorErrors;
```

adimascio: more generally, there a few missing semi-colons or extra spaces here and there, please use prettier :)

src/lib/translators/base.ts
rhuille: I agree! Done ✅

src/lib/translators/mongo4.ts
```diff
@@ -30,5 +30,7 @@ function transformConvert(step: Readonly<ConvertStep>): MongoStep {
   return { $addFields: mongoAddFields };
 }
 
-export class Mongo40Translator extends Mongo36Translator {}
+export class Mongo40Translator extends Mongo36Translator {
+  label = 'Mongo 4.0';
```
adimascio: NOTE: this will be coiped on all instances, I had imagined something like `static label = 'Mongo 4.0'` but then you would probably need to write `myTranslator.constructor.label` to access it from the instance

tests/unit/custom-step-form.spec.ts
```diff
@@ -0,0 +1,43 @@
+import CustomStepForm from '@/components/stepforms/CustomStepForm.vue';
+import { BasicStepFormTestRunner } from './utils';
+
+
+interface ValidationError {
```
adimascio: This interface is not used anymore :) This is strange that `yarn lint`doesnt detect it. I will try to fix the configuration

src/lib/translators/mongo4.ts
rhuille: I agree this is better ! https://github.com/ToucanToco/weaverbird/pull/406/files#diff-3480543ec07dddab995da489925679aeR72

PR #410, opened 2020-01-28
Fix unpivot translation to manage dropna option properly

vdestraitt:
Everything in the title


src/lib/translators/mongo.ts
```diff
@@ -355,12 +355,16 @@ function transformUnpivot(step: Readonly<S.UnpivotStep>): MongoStep[] {
   // projectCols to be included in Mongo $project steps
   const projectCols: PropMap<string> = _.fromPairs(step.keep.map(col => [col, `$${col}`]));
   // objectToArray to be included in the first Mongo $project step
-  const objectToArray: PropMap<string> = _.fromPairs(step.unpivot.map(col => [col, `$${col}`]));
+  const objectToArray: PropMap<object> = _.fromPairs(
+    step.unpivot.map(col => [col, { $ifNull: [`$${col}`, null] }]),
```
adimascio: What about importing `$$` from `@/lib/helpers` and use the code below instead:

```suggestion
    step.unpivot.map(col => [col, { $ifNull: [$$(col), null] }]),
```

Issue #413, opened 2020-01-31
Add DateExtract step


adimascio:
This step will extract `year`, `month`, `day`, etc. properties from a date columns.

vdestraitt:
closed by https://github.com/ToucanToco/weaverbird/pull/414

PR #414, opened 2020-01-31
add dateextract step definition

adimascio:
Introduce a new `dateextract` step to extract date properties
(e.g. `year`, `month`, `dayOfWeek`) from a date column.

For now, this is a mongo-only translatable step.

cf. #413


tests/unit/dateextract-step-form.spec.ts
```diff
@@ -0,0 +1,146 @@
+import ColumnPicker from '@/components/stepforms/ColumnPicker.vue';
+import DateExtractStepForm from '@/components/stepforms/DateExtractStepForm.vue';
+
+import { BasicStepFormTestRunner, setupMockStore } from './utils';
+
+describe('Concatenate Step Form', () => {
```
vdestraitt:
```suggestion
describe('Date Extract Step Form', () => {
```

src/components/stepforms/DateExtractStepForm.vue
```diff
@@ -0,0 +1,122 @@
...
+    InputTextWidget,
+  },
+})
+export default class DateExtractStepForm extends BaseStepForm<DateExtractPropertyStep> {
+  @Prop({ type: Object, default: () => ({ name: 'dateextract', column: '' }) })
+  initialStepValue!: DateExtractPropertyStep;
+
+  readonly title: string = 'Convert Column From Text to Date';
+
+  readonly operations: OperationOption[] = [
```
vdestraitt: There are too many operations to display in the UI. We need to keep a direct access to year/month/day/week, and have an "other" that will open the step form on click

src/components/constants.ts
```diff
@@ -57,6 +58,28 @@ export const ACTION_CATEGORIES: ActionCategories = {
   date: [
     { name: 'todate', label: 'Convert text to date' },
     { name: 'fromdate', label: 'Convert date to text' },
+    { name: 'dateextract', label: 'Extract year from date', defaults: { operation: 'year' } },
```
vdestraitt: In terms of labels, I would rather shorten them and say "Extract year", "Extract month" etc.
We are in the Date widget so no need to repeat date everywhere

PR #416, opened 2020-02-04
add basic support for regexp in filter steps

### LINKED ISSUE #415: Allow using regexp in filter steps
As as user, I'd like to be able to specify a regexp to match against instead of a raw value in the `filter` step form.

adimascio:
This commit introduces two new filter operators:

- `matches` to test a value against a regexp
- `notmatches` to make sure a value doesn't match a regexp

This is translated directly to the `$regex` operator in mongo.

UI wise, we chose to add two new operatoers in the operator dropdown and
use the standard input widget for the user to type its regexp.

Current known limitations:

- it is not possible to use regexp with `in` / `nin` operators
- it is not possible to use flags (such as `i` to ignore case) but this
  can be circumvented most of the time by adding a column with a
  normalized value (e.g. lowercased) before testing the regexp.

closes #415

### DIFF
--- a/src/components/stepforms/widgets/FilterSimpleCondition.vue
+++ b/src/components/stepforms/widgets/FilterSimpleCondition.vue
@@ -23,6 +23,7 @@
       />
     </div>
     <component
+      id="filterValue"
       v-if="inputWidget"
       :is="inputWidget"
       v-model="editedValue.value"
@@ -55,6 +56,8 @@ type LiteralOperator =
   | 'be less than or equal to'
   | 'be one of'
   | 'not be one of'
+  | 'matches pattern'
+  | "doesn't match pattern"
   | 'be null'
   | 'not be null';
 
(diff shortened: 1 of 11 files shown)


src/components/stepforms/widgets/FilterSimpleCondition.vue
```diff
@@ -55,6 +55,8 @@ type LiteralOperator =
   | 'be less than or equal to'
   | 'be one of'
   | 'not be one of'
+  | 'matches pattern'
```
vdestraitt: I think that we should be clear here that we expect a regex pattern.
Also, regex are very powerful but very geeky and not in everybody's reach.
Should we not implement, aside from this generic regex matching, more basic 'contains', 'does not contain', 'begins with', 'does not begin with', 'ends with', 'does not end with' which are the most frequent use cases that we should target here

src/lib/labeller.ts
```diff
@@ -65,6 +65,10 @@ function filterExpression(
         return `"${condition.column}" in (${condition.value.join(MULTIVALUE_SEP)})`;
       case 'nin':
         return `"${condition.column}" not in (${condition.value.join(MULTIVALUE_SEP)})`;
+      case 'matches':
+        return `"${condition.column}" matches "${condition.value}"`;
```
vdestraitt:
```suggestion
        return `"${condition.column}" matches regex "${condition.value}"`;
```

PR #418, opened 2020-02-05
Fix :: add form input text ellipsis and wrong height dataset cells

zegonz:
All in the title ⬆️ 

** Form multiselect add ellipsis **
before :
image

after :
image

** Dataset wrong height **
before :
image

Now:
image


Dalanir:
Neat !

PR #424, opened 2020-02-10
Condition editor component

davinov:
A new component to edit condition filter trees:
Screenshot from 2020-02-07 16-39-07

This recursive component accepts a slot to define the tree leave's form.

Here is the default case, with a simple input:
Screenshot from 2020-02-07 19-01-31

And a more advanced case with a three properties-object:
Screenshot from 2020-02-07 19-01-37

We intentionally removed the ability to create more than one level of nesting, to avoid unnecessary complexity. 

On the type-level, it defines a generic type: `AbstractConditionTree<T>` where T is the resulting type of the elements defined in the slot.

This component will then be used to enhance the FilterStep capabilities.

### DIFF
--- a/src/components/stepforms/ConditionsEditor/ConditionForm.vue
+++ b/src/components/stepforms/ConditionsEditor/ConditionForm.vue
@@ -1,0 +1,64 @@
+<template>
+  <div class="condition-form">
+    <input
+      :value="condition.column"
+      class="condition-form__input condition-form__input--column"
+      @input="updateInput('column', $event.target.value)"
+    >
+    <input
+      :value="condition.comparison"
+      class="condition-form__input condition-form__input--comparison"
+      @input="updateInput('comparison', $event.target.value)"
+    >
+    <input
+      :value="condition.value"
+      class="condition-form__input condition-form__input--value"
+      @input="updateInput('value', $event.target.value)"
+    >
+  </div>
+</template>
+
+<script lang="ts">
+import _ from 'lodash';
+
+import { Component, Prop, Vue } from 'vue-property-decorator';
+
+import { TempFilterCondition } from '@/lib/steps';
+
+@Component({
+  name: 'condition-form',
+})
+
+export default class ConditionForm extends Vue {
+  @Prop({
+    type: Object,
+    default: () => ({ column: '', comparison: 'eq', value: ''})
+  })
+  condition!: TempFilterCondition;
+
+  updateInput = _.debounce((type: string, newValue: any) => {
+    const newCondition = {
+      ...this.condition,
+      [type]: newValue,
+    };
+    this.$emit('conditionUpdated', newCondition);
+  }, 500);
+}
+</script>
+
+<style lang="scss">
+
+.condition-form {
+  display: flex;
+  flex-grow: 1;
+}
+
+.condition-form__input {
+  box-sizing: border-box;
+  height: 40px;
+  flex-grow: 1;
+  font-size: 14px;
+  margin: 3px;
+  padding: 8px 10px;
+}
+</style>
--- a/src/components/stepforms/ConditionsEditor/ConditionForm.vue
+++ b/src/components/stepforms/ConditionsEditor/ConditionForm.vue
@@ -19,7 +19,6 @@
 </template>
 
 <script lang="ts">
-import _ from 'lodash';
 import { Component, Prop, Vue } from 'vue-property-decorator';
 
 import { FilterSimpleCondition } from '@/lib/steps';
@@ -43,13 +42,13 @@ export default class ConditionForm extends Vue {
     }
   }
 
-  updateInput = _.debounce((type: string, newValue: any) => {
+  updateInput(type: string, newValue: any) {
     const newCondition = {
       ...this.condition,
       [type]: newValue,
     };
     this.emitUpdatedCondition(newCondition);
-  }, 500);
+  }
 
   emitUpdatedCondition(newCondition: FilterSimpleCondition) {
     this.$emit('conditionUpdated', newCondition);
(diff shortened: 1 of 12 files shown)


arimbr:
# [Codecov](https://codecov.io/gh/ToucanToco/weaverbird/pull/424?src=pr&el=h1) Report
> Merging [#424](https://codecov.io/gh/ToucanToco/weaverbird/pull/424?src=pr&el=desc) into [master](https://codecov.io/gh/ToucanToco/weaverbird/commit/547886d5a0a7c43c8e8ca362226b0f42ac0cc791&el=desc) will **increase** coverage by `0.01%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/ToucanToco/weaverbird/pull/424?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #424      +/-   ##
==========================================
+ Coverage   96.36%   96.38%   +0.01%     
==========================================
  Files         111      113       +2     
  Lines        2479     2542      +63     
  Branches      361      365       +4     
==========================================
+ Hits         2389     2450      +61     
- Misses         90       92       +2     
```


| [Impacted Files](https://codecov.io/gh/ToucanToco/weaverbird/pull/424?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [...nents/stepforms/ConditionsEditor/ConditionForm.vue](https://codecov.io/gh/ToucanToco/weaverbird/pull/424/diff?src=pr&el=tree#diff-c3JjL2NvbXBvbmVudHMvc3RlcGZvcm1zL0NvbmRpdGlvbnNFZGl0b3IvQ29uZGl0aW9uRm9ybS52dWU=) | `86.66% <0.00%> (ø)` | |
| [...nts/stepforms/ConditionsEditor/ConditionsGroup.vue](https://codecov.io/gh/ToucanToco/weaverbird/pull/424/diff?src=pr&el=tree#diff-c3JjL2NvbXBvbmVudHMvc3RlcGZvcm1zL0NvbmRpdGlvbnNFZGl0b3IvQ29uZGl0aW9uc0dyb3VwLnZ1ZQ==) | `100.00% <0.00%> (ø)` | |

------

[Continue to review full report at Codecov](https://codecov.io/gh/ToucanToco/weaverbird/pull/424?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/ToucanToco/weaverbird/pull/424?src=pr&el=footer). Last update [547886d...cc0791e](https://codecov.io/gh/ToucanToco/weaverbird/pull/424?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

src/components/stepforms/ConditionsEditor/ConditionForm.vue
```diff
@@ -0,0 +1,74 @@
...
+    <input
+      :value="condition.value"
+      class="condition-form__input condition-form__input--value"
+      @input="updateInput('value', $event.target.value)"
+    />
+  </div>
+</template>
+
+<script lang="ts">
+import _ from 'lodash';
```
adimascio:
```suggestion
import debounce from 'lodash/debounce';
```

and subsequent changes later in the component

src/components/stepforms/ConditionsEditor/ConditionsGroup.vue
```diff
@@ -0,0 +1,470 @@
...
+        <div
+          class="conditions-group__switch-button conditions-group__switch-button--or"
+          :class="{ 'conditions-group__switch-button--active': operator === 'or' }"
+          @click="switchOperator('or')"
+        >
+          or
+        </div>
+      </div>
+    </div>
+    <div v-for="(condition, rowIndex) in conditions" :key="'row' + rowIndex" class="condition-row">
```
adimascio:
```suggestion
    <div v-for="(condition, rowIndex) in conditions" :key="`row${rowIndex}`" class="condition-row">
```

but wouldn't `rowIndex` be enough in that case? Besides, isn't it a bit fragile to rely solely on the index?

src/components/stepforms/ConditionsEditor/ConditionsGroup.vue
```diff
@@ -0,0 +1,470 @@
...
+      <div
+        v-if="hasMultipleRows"
+        class="condition-row__link"
+        :class="{
+          'condition-row__link--last':
+            !(groups && groups.length) && rowIndex === conditions.length - 1,
```
adimascio: This is a matter of taste but I find the expression a bit too hard to read at first sight :)

src/components/stepforms/ConditionsEditor/ConditionForm.vue
davinov: Actually, reading this code again, I think it's not a good idea to introduce a `debounce` here.
Updating an object is not an expansive operation.
I prefer to let a higher component, which would trigger such operation at each form update, tocarry the debounce logic. (Then, the debounce delay would be chosen regarding the time cost of the operation)   

Thanks for noticing!

src/components/stepforms/ConditionsEditor/ConditionsGroup.vue
davinov: The underlaying structure of data is:
```
conditions: [x, y, z]
groups: [xg, yg, zg]
```

Therefore, the `v-for` need an id to identify x, y, z, xg, yg and zg.
- Adding `group` and `row` before the index distinguishes the conditions array elements from the `group` ones
- The index in the array seems enough because:
  - The order of the elements inside each array is not going to change (as operators are commutatives, there would be no need to reorder them)
  - As the content of a condition can be completly modified, and still stay the same row, I don't see any other possible identification for it.

src/components/stepforms/ConditionsEditor/ConditionForm.vue
davinov: See d387f33

src/components/stepforms/ConditionsEditor/ConditionsGroup.vue
davinov: Totally right, I refactored this into methods that explains the logic better.
I also added a few tests to ensure this logic won't break.
See fe0c1ed

PR #425, opened 2020-02-10
Fix :: remove css cursor pointer on data viewer cells / add fallback fontAwesome declaration

zegonz:
everything is explained in this F**** long title  ⬆️ 

Before in tucana :
image

Now you get the right font icon arrow:
image


zegonz:
[ETQC le CSS du VQB est OK dans Tucana](https://trello.com/c/HJSUxLPv/3834-etqc-le-css-du-vqb-est-ok-dans-tucana)

PR #426, opened 2020-02-11
Feature :: Get unique groups/values step

vdestraitt:
Allows to get unique values from a column or unique groups or value from the combination of several columns.

This PR includes:

- The translation in Mongo
- The implementation of a labeller
- The implementation of templater
- The UI compoents
- The documentation


src/lib/translators/mongo.ts
```diff
@@ -355,6 +355,14 @@ function transformTop(step: Readonly<S.TopStep>): MongoStep[] {
   ];
 }
 
+/** transform an 'uniquegroups' step into corresponding mongo steps */
+function transformUniqueGroups(step: Readonly<S.UniqueGroupsStep>): MongoStep[] {
+  const id = columnMap(step.on);
+  const project = _.fromPairs(Object.keys(id).map(col => [col, `$_id.${col}`]));
+  _.fromPairs(Object.keys(id).map(col => [col, `$_id.${col}`]));
```
adimascio: I think this line is just a leftover and should be removed :)

src/lib/translators/mongo.ts
```diff
@@ -355,6 +355,14 @@ function transformTop(step: Readonly<S.TopStep>): MongoStep[] {
   ];
 }
 
+/** transform an 'uniquegroups' step into corresponding mongo steps */
+function transformUniqueGroups(step: Readonly<S.UniqueGroupsStep>): MongoStep[] {
+  const id = columnMap(step.on);
+  const project = _.fromPairs(Object.keys(id).map(col => [col, `$_id.${col}`]));
```
adimascio:
```suggestion
  const project = Object.entries(Object.keys(id).map(col => [col, `$_id.${col}`]));
```

tests/unit/action-menu.spec.ts
```diff
@@ -205,4 +205,29 @@ describe('Action Menu', () => {
 
     expect(wrapper.emitted().closed).toBeTruthy();
   });
+
+  describe('when clicking on "Get unique values"', () => {
+    it('should add a valide uniquegroups step in the pipeline', async () => {
```
adimascio:
```suggestion
    it('should add a valie uniquegroups step in the pipeline', async () => {
```

PR #434, opened 2020-02-12
[mongo translator] :: allow to escape column name in FormulaStep with "[" and "]"

### LINKED ISSUE #295: Formula step should manage columns names including whitespaces properly
In particular spaces in the middle of the name (e.g. "Market Share").
Currently a string escaped by simple or double quotes is considered as a string and not as a column name.

rhuille:
[mongo translator] :: allow to escape column name in FormulaStep with "[" and "]"

Before parsing the formula into a tree by mathjs, we replace all "[...]" by valid pseudo ("valid" for mathjs) and then re-put it back after building the mongo formula.

closes #295

### DIFF
--- a/src/components/stepforms/FormulaStepForm.vue
+++ b/src/components/stepforms/FormulaStepForm.vue
@@ -65,7 +65,7 @@ export default class FormulaStepForm extends BaseStepForm<FormulaStep> {
     let ret = this.validator({ ...this.editedStep });
     let errors: ErrorObject[] = [];
     try {
-      parse(this.editedStep.formula);
+      parse(this.editedStep.formula.replace(/\[.*?\]/g, 'var'));
     } catch {
       ret = false;
       errors.push({
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -408,13 +408,47 @@ function getOperator(op: string) {
 }
 
 /**
+  Transform a formula expression into a MathNode
+  1. Replace in formula all column name between `[]` by a "pseudo"
+  2. Parse the formula into a MathNode
+  3. Replace all pseudo into MathNode by there original name
+*/
+function buildFormulaTree(formula: string): MathNode {
+  const ESCAPE_OPEN = '[';
+  const ESCAPE_CLOSE = ']';
+
+  // 1. Replace in formula all column name between `[]` by a "pseudo"
+  let formulaPseudotised = formula;
+  const pseudo: Record<string, string> = {};
+  let index = 0;
+  const regex = new RegExp(`\\${ESCAPE_OPEN}(.*?)\\${ESCAPE_CLOSE}`, 'g');
+  for (const match of formula.match(regex) || []) {
+    pseudo[`__vqb_col_${index}__`] = match;
+    formulaPseudotised = formulaPseudotised.replace(match, `__vqb_col_${index}__`);
+    index++;
+  }
+
+  // 2. Parse the formula into a MathNode
+  const mathjsTree: MathNode = math.parse(formulaPseudotised);
+
+  // 3. Replace all pseudo into MathNode by there original name
+  mathjsTree.traverse(function(node: MathNode): MathNode {
+    if (node.type === 'SymbolNode') {
+      if (pseudo[node.name]) {
+        node.name = pseudo[node.name].replace(ESCAPE_OPEN, '').replace(ESCAPE_CLOSE, '');
+      }
+    }
+    return node;
+  });
+  return mathjsTree;
+}
+
+/**
  * Translate a mathjs logical tree describing a formula into a Mongo step
  * @param node a mathjs node object (usually received after parsing an string expression)
  * This node is the root node of the logical tree describing the formula
  */
 function buildMongoFormulaTree(node: MathNode): MongoStep | string | number {
-  // For type checking in `case: 'OperatorNode'` in the`switch`clause below,
-  // do not let`args` and `op` be potentially`undefined`
   switch (node.type) {
     case 'OperatorNode':
       if (node.args.length === 1) {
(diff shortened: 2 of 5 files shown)


src/lib/translators/mongo.ts
```diff
@@ -545,11 +548,22 @@ const mapper: Partial<StepMatcher<MongoStep>> = {
...
-  formula: (step: Readonly<S.FormulaStep>) => ({
-    $addFields: {
-      [step.new_column]: buildMongoFormulaTree(math.parse(step.formula)),
-    },
-  }),
+  formula: (step: Readonly<S.FormulaStep>) => {
+    // replace all `[column name]` values by a pseudo and store it in `pseudos`
+    // `buildMongoFormulaTree` will re-put the name back
+    const pseudos: { [key: string]: string } = {};
+    const formulaPseudotised = step.formula.replace(/\[.*?\]/g, columName => {
```
adimascio: I think you could do something like:

```suggestion
    const formulaPseudotised = step.formula.replace(/\[(.*?)\]/g, (_match, columName) => {
```

and avoid calling`replace` inside the callback, cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#Specifying_a_function_as_a_parameter

src/lib/translators/mongo.ts
```diff
@@ -545,11 +548,22 @@ const mapper: Partial<StepMatcher<MongoStep>> = {
...
+      const pseudo = `column_n_${Object.keys(pseudos).length}`;
```
adimascio: We have a "tradition" of prefixing such vqb-specific variables with `vqb`, cf. https://github.com/ToucanToco/weaverbird/blob/master/src/lib/translators/mongo.ts#L188 for instance. Maybe we could do that here?

src/lib/translators/mongo.ts
```diff
@@ -545,11 +548,22 @@ const mapper: Partial<StepMatcher<MongoStep>> = {
...
   filter: filterstepToMatchstep,
-  formula: (step: Readonly<S.FormulaStep>) => ({
-    $addFields: {
-      [step.new_column]: buildMongoFormulaTree(math.parse(step.formula)),
-    },
-  }),
+  formula: (step: Readonly<S.FormulaStep>) => {
+    // replace all `[column name]` values by a pseudo and store it in `pseudos`
+    // `buildMongoFormulaTree` will re-put the name back
+    const pseudos: { [key: string]: string } = {};
```
adimascio: It's a matter of style but I'm getting used to `Record<string, string>` and now I think it's a bit clearer. What do you think?

src/lib/translators/mongo.ts
```diff
@@ -410,28 +410,31 @@ function getOperator(op: string) {
...
  * @param node a mathjs node object (usually received after parsing an string expression)
+ * @param pseudos an object which keys are pseudo and values are corresponding column name
  * This node is the root node of the logical tree describing the formula
  */
-function buildMongoFormulaTree(node: MathNode): MongoStep | string | number {
-  // For type checking in `case: 'OperatorNode'` in the`switch`clause below,
-  // do not let`args` and `op` be potentially`undefined`
+function buildMongoFormulaTree(
+  node: MathNode,
+  pseudos: { [key: string]: string },
```
adimascio: It's a matter of style but I'm getting used to `Record<string, string>` and now I think it's a bit clearer. What do you think?

src/lib/translators/mongo.ts
```diff
@@ -545,11 +548,22 @@ const mapper: Partial<StepMatcher<MongoStep>> = {
...
-      [step.new_column]: buildMongoFormulaTree(math.parse(step.formula)),
-    },
-  }),
+  formula: (step: Readonly<S.FormulaStep>) => {
+    // replace all `[column name]` values by a pseudo and store it in `pseudos`
+    // `buildMongoFormulaTree` will re-put the name back
+    const pseudos: { [key: string]: string } = {};
+    const formulaPseudotised = step.formula.replace(/\[.*?\]/g, columName => {
+      const pseudo = `column_n_${Object.keys(pseudos).length}`;
+      pseudos[pseudo] = columName.replace('[', '').replace(']', '');
```
adimascio: I might be overlooking but it seems that using twice the same "column" name in a single formula would generate two distinct pseudos and I think it would be nice to always use the same. What do you think?

Besides, is everything translated correctly if there are several formula steps in a single pipeline? Are there any risk of overwriting a given "column"?

tests/unit/mongo.spec.ts
```diff
@@ -1096,6 +1096,48 @@ describe('Pipeline to mongo translator', () => {
...
+        new_column: 'test',
+        formula: '[column with space and + and, oh a - and_also *] + [an other ^column]',
+      },
+    ];
+    const querySteps = mongo36translator.translate(pipeline);
+    expect(querySteps).toEqual([
+      {
+        $addFields: {
+          test: {
+            $add: ['$column with space and + and, oh a - and_also *', '$an other ^column'],
```
adimascio: You made sure that it was valid mongo-wise, right? :)

src/components/stepforms/FormulaStepForm.vue
```diff
@@ -65,7 +65,7 @@ export default class FormulaStepForm extends BaseStepForm<FormulaStep> {
     let ret = this.validator({ ...this.editedStep });
     let errors: ErrorObject[] = [];
     try {
-      parse(this.editedStep.formula);
+      parse(this.editedStep.formula.replace(/\[.*?\]/g, 'var'));
```
adimascio: What would you think of adding `[]` to a string constant and use it? It's a genuine question, I have no real strong opinion for this specific case.

PR #439, opened 2020-02-17
Tidy the store to use pipelines library and currentPipelineName as references 

davinov:
The state had two separate properties: `pipelines` and `pipeline`. This can lead to a confusing situation where pipelines are not set but pipeline is, and therefore the dereferencing can't happen correctly.
Now, `pipeline` is a getter depending on `pipelines` and `currentPipelineName`.
The action `setPipeline` is still available, and updates `pipelines[currentPipelineName]`, so big API change.

I took the opportunity to try a few typings improvements along the way.

### DIFF
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -3,7 +3,7 @@
  */
 import _ from 'lodash';
 
-import { activePipeline, inactivePipeline, VQBState } from './state';
+import { activePipeline, currentPipeline, inactivePipeline, VQBState } from './state';
 
 export default {
   /**
@@ -32,12 +32,12 @@ export default {
    * a direct "usable" index (i.e. convert "-1" to a positive one) of last active step.
    */
   computedActiveStepIndex: (state: VQBState) =>
-    state.selectedStepIndex === -1 ? state.pipeline.length - 1 : state.selectedStepIndex,
+    state.selectedStepIndex === -1 ? currentPipeline(state).length - 1 : state.selectedStepIndex,
   /**
    * the first step of the pipeline. Since it's handled differently in the UI,
    * it's useful to be able to access it directly.
    */
-  domainStep: (state: VQBState) => state.pipeline[0],
+  domainStep: (state: VQBState) => currentPipeline(state)[0],
   /**
    * the part of the pipeline that is currently disabled.
    */
@@ -53,7 +53,7 @@ export default {
   /**
    * helper that is True if pipeline is empty or only contain a domain step.
    */
-  isPipelineEmpty: (state: VQBState) => state.pipeline.length <= 1,
+  isPipelineEmpty: (state: VQBState) => currentPipeline(state).length <= 1,
   /**
    * helper that is True if this step is after the last currently active step.
    */
@@ -64,6 +64,10 @@ export default {
    */
   pageno: (state: VQBState) =>
     state.dataset.paginationContext ? state.dataset.paginationContext.pageno : 1,
+  /**
+   * Return current edited pipeline
+   */
+  pipeline: (state: VQBState) => currentPipeline(state),
   /**
    * Return pipelines save in store
    */
@@ -79,7 +83,7 @@ export default {
   /**
    * Get the step config of the pipeline based on its index
    */
-  stepConfig: (state: VQBState) => (index: number) => state.pipeline[index],
+  stepConfig: (state: VQBState) => (index: number) => currentPipeline(state)[index],
   /**
    * Return the app translator name
    */
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -2,10 +2,11 @@
  * exports the list of store getters.
  */
 import _ from 'lodash';
+import { GetterTree } from 'vuex';
 
 import { activePipeline, currentPipeline, inactivePipeline, VQBState } from './state';
 
-export default {
+const getters: GetterTree<VQBState, any> = {
   /**
    * the part of the pipeline that is currently selected.
    */
@@ -89,3 +90,5 @@ export default {
    */
   translator: (state: VQBState) => state.translator,
 };
+
+export default getters;
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -5,13 +5,13 @@
  * e.g. `registerModule(hostStore, initialState)` to register a VQB store
  * and `unregisterModule()` for the dual operation.
  *
- * A `vuex-class`'s namesapce is exported as the '`VQBModule` symbol. It can therefore
+ * A `vuex-class`'s namespace is exported as the '`VQBModule` symbol. It can therefore
  * be used as a decorator inside a `Vue` component, .e.g.:
  *
  * ```typescript
  * @Component
  * class MyComponent extends Vue {
- *    @VQBModule.State pipeline!: Pipeline;
+ *    @VQBModule.Getter pipeline!: Pipeline;
  *    // ...
  * }
  * ```
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -42,7 +42,7 @@ export const VQBModule = namespace(VQB_MODULE_NAME);
 export function buildStoreModule(initialState: Partial<VQBState> = {}) {
   const store = {
     namespaced: true,
-    state: { ...emptyState, ...initialState },
+    state: { ...emptyState(), ...initialState },
     getters,
     mutations,
   };
@@ -81,7 +81,7 @@ export function setupStore(
     Vue.use(Vuex);
   }
   const store: StoreOptions<VQBState> = {
-    state: { ...emptyState, ...initialState },
+    state: { ...emptyState(), ...initialState },
     getters,
     mutations,
     plugins,
--- a/src/store/mutations.ts
+++ b/src/store/mutations.ts
@@ -3,7 +3,7 @@
  */
 
 import { BackendError } from '@/lib/backend-response';
-import { DomainStep, PipelineStepName } from '@/lib/steps';
+import { DomainStep, Pipeline, PipelineStepName } from '@/lib/steps';
 
 import { VQBState } from './state';
 
@@ -30,7 +30,12 @@ type DomainsMutation = {
 
 type PipelineMutation = {
   type: 'setPipeline';
-  payload: Pick<VQBState, 'pipeline'>;
+  payload: { pipeline: Pipeline };
+};
+
+type SetCurrentPipelineNameMutation = {
+  type: 'setCurrentPipelineName';
+  payload: { name: string };
 };
 
 type SelectDomainMutation = {
@@ -64,6 +69,7 @@ export type StateMutation =
   | DeleteStepMutation
   | DomainsMutation
   | PipelineMutation
+  | SetCurrentPipelineNameMutation
   | SelectedColumnsMutation
   | SelectDomainMutation
   | SelectedStepMutation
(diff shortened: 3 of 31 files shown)


src/store/getters.ts
```diff
@@ -64,10 +64,14 @@ export default {
...
     state.dataset.paginationContext ? state.dataset.paginationContext.pageno : 1,
+  /**
+   * Return current edited pipeline
+   */
+  pipeline: (state: VQBState) => currentPipeline(state),
   /**
    * Return pipelines save in store
    */
-  pipelines: (state: VQBState) => state.pipelines,
+  pipelines: (state: VQBState) => currentPipeline(state),
```
adimascio: :thinking: ?

src/store/mutations.ts
```diff
@@ -30,7 +30,12 @@ type DomainsMutation = {
 
 type PipelineMutation = {
   type: 'setPipeline';
-  payload: Pick<VQBState, 'pipeline'>;
+  payload: Pipeline;
```
adimascio: Shouldn't this be `payload: { pipeline: Pipeline }` to be consistent with other mutations?

src/store/state.ts
```diff
@@ -112,9 +107,10 @@ export const emptyState: VQBState = {
   stepFormInitialValue: undefined,
   stepFormDefaults: undefined,
   selectedStepIndex: -1,
-  pipeline: [],
-  currentPipelineName: undefined,
-  pipelines: {},
+  currentPipelineName: 'my_pipeline',
+  pipelines: {
+    my_pipeline: [],
+  },
```
adimascio: I kind of dislike the idea of an hardcoded `my_pipeline` initial value. I don't like the `my_` part but this is admittedly a matter of taste :)
Wouldn't it be better to have `currentPipelineName` and `pipelines` set to `undefined` at first? And maybe handle those cases in the getters. After all, that's why there are getters, don't you think?

src/store/state.ts
```diff
@@ -124,15 +120,27 @@ export const emptyState: VQBState = {
   interpolateFunc: (x: string, _context: ScopeContext) => x,
 };
 
+/**
+ * @param state current application state
+ * @return the pipeline currently edited
+ */
+export function currentPipeline(state: VQBState) {
```
adimascio: Genuine question: any reason to have it as a function rather than directly being a getter on the store?

src/store/index.ts
```diff
@@ -42,7 +42,7 @@ export const VQBModule = namespace(VQB_MODULE_NAME);
 export function buildStoreModule(initialState: Partial<VQBState> = {}) {
   const store = {
     namespaced: true,
-    state: { ...emptyState, ...initialState },
+    state: { ...emptyState(), ...initialState },
```
adimascio: :+1:

tests/unit/store.spec.ts
```diff
@@ -3,9 +3,18 @@ import getters from '@/store/getters';
 import mutations from '@/store/mutations';
 import { emptyState, VQBState } from '@/store/state';
 
+function stateWithOnePipeline(pipeline: Pipeline) {
```
adimascio: What about
```suggestion
function stateWithOnePipeline(pipeline: Pipeline, customProperties = {}) {
```

and then merge initial pipeline with `customProperties`?

tests/unit/store.spec.ts
```diff
@@ -38,7 +47,7 @@ describe('getter tests', () => {
         { name: 'rename', oldname: 'foo', newname: 'bar' },
         { name: 'rename', oldname: 'baz', newname: 'spam' },
       ];
-      const state = buildState({ pipeline });
+      const state = buildState({ ...stateWithOnePipeline(pipeline) });
```
adimascio:
```suggestion
      const state = buildState(stateWithOnePipeline(pipeline));
```

tests/unit/store.spec.ts
```diff
@@ -60,7 +69,7 @@ describe('getter tests', () => {
...
```
adimascio:
```suggestion
      const state = buildState(stateWithOnePipeline(pipeline));
```

tests/unit/store.spec.ts
```diff
@@ -125,7 +134,7 @@ describe('getter tests', () => {
...
```
adimascio:
```suggestion
      const state = buildState(stateWithOnePipeline(pipeline));
```

tests/unit/store.spec.ts
```diff
@@ -155,7 +164,7 @@ describe('getter tests', () => {
   describe('pipeline empty tests', () => {
     it('should return true if pipeline is empty', () => {
       const pipeline: Pipeline = [];
-      const state = buildState({ pipeline });
+      const state = buildState({ ...stateWithOnePipeline(pipeline) });
```
adimascio:
```suggestion
      const state = buildState(stateWithOnePipeline(pipeline));
```

tests/unit/store.spec.ts
```diff
@@ -165,7 +174,7 @@ describe('getter tests', () => {
         { name: 'rename', oldname: 'foo', newname: 'bar' },
         { name: 'rename', oldname: 'baz', newname: 'spam' },
       ];
-      const state = buildState({ pipeline });
+      const state = buildState({ ...stateWithOnePipeline(pipeline) });
```
adimascio:
```suggestion
      const state = buildState(stateWithOnePipeline(pipeline));
```

tests/unit/store.spec.ts
```diff
@@ -246,11 +255,11 @@ describe('mutation tests', () => {
       { name: 'rename', oldname: 'foo', newname: 'bar' },
       { name: 'rename', oldname: 'baz', newname: 'spam' },
     ];
-    const state = buildState({ currentDomain: 'foo', pipeline });
+    const state = buildState({ currentDomain: 'foo', ...stateWithOnePipeline(pipeline) });
```
adimascio:
```suggestion
      const state = buildState(stateWithOnePipeline(pipeline));
```

tests/unit/store.spec.ts
```diff
@@ -28,7 +28,7 @@ describe('getter tests', () => {
         { name: 'rename', oldname: 'baz', newname: 'spam' },
       ];
       const state = buildState(stateWithOnePipeline(pipeline));
-      expect(getters.activePipeline(state)).toEqual(pipeline);
+      expect(getters.activePipeline(state, {}, {}, {})).toEqual(pipeline);
```
adimascio: Hmm, I did not take the time to check / dig but this doesn't feel right. We shouldn't need to pass those 3 extra empty objects.

davinov: I'm afraid we do.
For better type inference, I added the type `GetterTree` to the `getters` object:
https://github.com/ToucanToco/weaverbird/pull/439/commits/9eedeee50f8aab95a1d333751c1833f1e56e1179
That means that every function defined there is of type `Getter`.
A `Getter` is a function that takes as arguments `ModuleState`, `ModuleGetters`, `RootState`, `RootGetters`. All are required.

If we want to avoid this (which I won't recommend), the workaround would be to define the functions alone, and then to add them into the `GetterTree getters` object later.
This way, using their "standalone" version comes with the inferred type `VQBState => something` rather that the `Getter` type.
I personally prefer the first option, where it become kinda obvious we're having a getter.

tests/unit/store.spec.ts
davinov: Thanks for noticing these which have passed my bulk replace vigilance :see_no_evil:

src/store/state.ts
davinov: I followed the model of the other functions defined in this file, below.
This function is imported into the getters file and used as a getter there.

We could also import the getter here and use it, it seems equivalent to me. The important thing is that we have the "logic" of getter a pipeline by its name only in one plcae?

src/store/state.ts
davinov: That's a good point ! I gave a bit of thought into this:

I started as you said by having an `undefined` state for both `currentPipelineName` and `pipelines`, and therefore the current `pipeline`. This caused quite a massive amount of ts errors, indicating the case where pipeline is not defined should be handled.
I thought of implementing these, before realizing that may not be the scope of this PR:
- before, we had a `pipeline: []`, so no case where the pipeline is not defined
- we also do not have a UI case where no pipeline is selected, so it feels weird implementing one before knowing what to display in this case.
My conclusion: this should be also be a design issue and be another refactoring outside this PR.

I choosed the `my_pipeline` in accordance with what we choose for Toucan Toco's default dataset name `my_dataset`. I also didn't liked it at first, but it makes clear that it is a customizable name, that acts as a placeholder. The other possibilities that came to my mind ('default`, `defaultPipeline`, `pipeline`) didn't seem as clear to me. If you have any better suggestion, I'll be very pleased to change that!

src/store/mutations.ts
davinov: What I want to express here is: it should be the type of any of the properties values of pipelines.
We may be able to describe thas as `Pick<<Pick<VQBState, 'pipeline'>, any>` but:
- I'm unsure the `any` as second parameter will match any of the pipeline's keys
- it's really confusing to read
- if someone puts something of the wrong type in state.pipelines[...], it would authorize something wrong here.

So I really prefer to keep that type simple and readable: it should be a `Pipeline`

src/store/getters.ts
davinov: Oupsy ! thanks for noticing

tests/unit/store.spec.ts
davinov: I personally feel that brings nothing helpful compared to a simple splat. My policy would be: if a simple syntax feature, clear to read, is there, I'd prefer the language feature to a custom function. (It also seems to be our current policy towards other utilities like lodash).  
Could you develop on which errors it could prevent, or on how simpler to use if could be? Very open to challenge my policy regarding that :)

arimbr:
# [Codecov](https://codecov.io/gh/ToucanToco/weaverbird/pull/439?src=pr&el=h1) Report
> Merging [#439](https://codecov.io/gh/ToucanToco/weaverbird/pull/439?src=pr&el=desc) into [master](https://codecov.io/gh/ToucanToco/weaverbird/commit/602f9c90a9fd556583da4c4b1ebb49664237ab73&el=desc) will **decrease** coverage by `0.03%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/ToucanToco/weaverbird/pull/439?src=pr&el=tree)

```diff
@@            Coverage Diff             @@
##           master     #439      +/-   ##
==========================================
- Coverage   96.46%   96.43%   -0.04%     
==========================================
  Files         117      117              
  Lines        2600     2606       +6     
  Branches      369      370       +1     
==========================================
+ Hits         2508     2513       +5     
- Misses         92       93       +1     
```



------

[Continue to review full report at Codecov](https://codecov.io/gh/ToucanToco/weaverbird/pull/439?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/ToucanToco/weaverbird/pull/439?src=pr&el=footer). Last update [602f9c9...57594f5](https://codecov.io/gh/ToucanToco/weaverbird/pull/439?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

tests/unit/store.spec.ts
adimascio: No strong opinion about that, it's just that it's a classical pattern to have a parameter to override defaults. Besides, it would mimic the `setupStore()`

src/store/state.ts
adimascio: resolving in favor of https://github.com/ToucanToco/weaverbird/issues/441

tests/unit/store.spec.ts
adimascio: OK fine. the `@Getters` decorator will handle extra arguments.

PR #444, opened 2020-02-20
Pipeline selector and no pipeline state

davinov:
Add a component to choose which pipeline is currently displayed.
Introduce the possibility that no pipeline can be selected.

Screenshot from 2020-02-20 21-03-09

All the available pipelines are listed:
Screenshot from 2020-02-20 21-03-20

And changing the selected pipeline displays the newly selected pipeline at its last step:
Screenshot from 2020-02-20 21-03-26

Update: empty state when no pipeline is selected:
Screenshot from 2020-02-21 18-21-38
(To reproduce it, change the initial `currentPipelineName` in the playground's to `undefined`)

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -166,24 +166,6 @@ class MongoService {
 
 const mongoservice = new MongoService();
 const mongoBackendPlugin = servicePluginFactory(mongoservice);
-const initialPipeline = [
-  {
-    name: 'domain',
-    domain: 'test-collection',
-  },
-  {
-    name: 'filter',
-    condition: {
-      and: [
-        {
-          column: 'Groups',
-          value: '<%= groupname %>',
-          operator: 'eq',
-        },
-      ],
-    },
-  },
-];
 
 async function setupInitialData(store, domain = null) {
   const collections = await mongoservice.listCollections();
--- a/playground/app.js
+++ b/playground/app.js
@@ -328,6 +328,9 @@ async function buildVueApp() {
     computed: {
       code: function() {
         let activePipeline = this.$store.getters[VQBnamespace('activePipeline')];
+        if (!activePipeline) {
+          return '';
+        }
         const pipelines = this.$store.getters[VQBnamespace('pipelines')];
         if (pipelines) {
           activePipeline = dereferencePipelines(activePipeline, pipelines);
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -1,0 +1,14 @@
+import { ActionTree } from 'vuex';
+
+import { VQBState } from './state';
+
+const actions: ActionTree<VQBState, any> = {
+  selectPipeline({ commit }, { name }: { name: string }) {
+    commit('setCurrentPipelineName', { name: name });
+
+    // Reset selected step to last one
+    commit('selectStep', { index: -1 });
+  },
+};
+
+export default actions;
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -32,13 +32,24 @@ const getters: GetterTree<VQBState, any> = {
   /**
    * a direct "usable" index (i.e. convert "-1" to a positive one) of last active step.
    */
-  computedActiveStepIndex: (state: VQBState) =>
-    state.selectedStepIndex === -1 ? currentPipeline(state).length - 1 : state.selectedStepIndex,
+  computedActiveStepIndex(state: VQBState) {
+    const pipeline = currentPipeline(state);
+    if (!pipeline) {
+      return;
+    }
+    return state.selectedStepIndex === -1 ? pipeline.length - 1 : state.selectedStepIndex;
+  },
   /**
    * the first step of the pipeline. Since it's handled differently in the UI,
    * it's useful to be able to access it directly.
    */
-  domainStep: (state: VQBState) => currentPipeline(state)[0],
+  domainStep(state: VQBState) {
+    const pipeline = currentPipeline(state);
+    if (!pipeline) {
+      return;
+    }
+    return pipeline?.[0];
+  },
   /**
    * the part of the pipeline that is currently disabled.
    */
@@ -54,7 +65,13 @@ const getters: GetterTree<VQBState, any> = {
   /**
    * helper that is True if pipeline is empty or only contain a domain step.
    */
-  isPipelineEmpty: (state: VQBState) => currentPipeline(state).length <= 1,
+  isPipelineEmpty(state: VQBState) {
+    const pipeline = currentPipeline(state);
+    if (!pipeline) {
+      return;
+    }
+    return pipeline.length <= 1;
+  },
   /**
    * helper that is True if this step is after the last currently active step.
    */
@@ -84,7 +101,10 @@ const getters: GetterTree<VQBState, any> = {
   /**
    * Get the step config of the pipeline based on its index
    */
-  stepConfig: (state: VQBState) => (index: number) => currentPipeline(state)[index],
+  stepConfig: (state: VQBState) => (index: number) => {
+    const pipeline = currentPipeline(state);
+    return pipeline?.[index];
+  },
   /**
    * Return the app translator name
    */
--- a/src/store/getters.ts
+++ b/src/store/getters.ts
@@ -91,6 +91,10 @@ const getters: GetterTree<VQBState, any> = {
    */
   pipelines: (state: VQBState) => state.pipelines,
   /**
+   * Return the array of pipeline names
+   */
+  pipelinesNames: (state: VQBState) => Object.keys(state.pipelines),
+  /**
    * Return true if an error occured in the backend
    */
   thereIsABackendError: (state: VQBState) => state.backendErrors.length > 0,
--- a/src/store/mutations.ts
+++ b/src/store/mutations.ts
@@ -5,7 +5,7 @@
 import { BackendError } from '@/lib/backend-response';
 import { DomainStep, Pipeline, PipelineStepName } from '@/lib/steps';
 
-import { VQBState } from './state';
+import { currentPipeline, VQBState } from './state';
 
 // provide types for each possible mutations' payloads
 type BackendErrorMutation = {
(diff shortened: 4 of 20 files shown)


src/store/getters.ts
```diff
@@ -32,13 +32,24 @@ const getters: GetterTree<VQBState, any> = {
...
+    }
+    return state.selectedStepIndex === -1 ? pipeline.length - 1 : state.selectedStepIndex;
+  },
   /**
    * the first step of the pipeline. Since it's handled differently in the UI,
    * it's useful to be able to access it directly.
    */
-  domainStep: (state: VQBState) => currentPipeline(state)[0],
+  domainStep(state: VQBState) {
+    const pipeline = currentPipeline(state);
```
adimascio: I think you could use the conditional operator here, as in:

```typescript
return currentPipeline(state)?[0];
```

src/store/getters.ts
```diff
@@ -84,7 +101,13 @@ const getters: GetterTree<VQBState, any> = {
   /**
    * Get the step config of the pipeline based on its index
    */
-  stepConfig: (state: VQBState) => (index: number) => currentPipeline(state)[index],
+  stepConfig: (state: VQBState) => (index: number) => {
+    const pipeline = currentPipeline(state);
```
adimascio: same suggestion as above.

src/store/mutations.ts
```diff
@@ -191,6 +199,9 @@ export default {
    * update pipeline.
    */
   setPipeline(state: VQBState, { pipeline }: { pipeline: Pipeline }) {
+    if (state.currentPipelineName == null) {
```
adimascio: It seems that `currentPipelineName` can't be `null` from a type point of view so I guess 

```suggestion
    if (state.currentPipelineName === undefined) {
```

would be better or am I missing something?

src/store/mutations.ts
```diff
@@ -148,16 +149,23 @@ export default {
...
+    }
+    pipeline.splice(index, 1);
     state.selectedStepIndex = index - 1;
   },
   /**
    * change current selected domain and reset pipeline accordingly.
    */
   setCurrentDomain(state: VQBState, { currentDomain }: Pick<VQBState, 'currentDomain'>) {
+    const pipeline = currentPipeline(state);
+    if (state.currentPipelineName == null || !pipeline) {
```
adimascio: It seems that `currentPipelineName` can't be `null` from a type point of view so I guess 

```suggestion
    if (state.currentPipelineName === undefined || pipeline === undefined) {
```

src/store/state.ts
```diff
@@ -148,13 +149,21 @@ function firstNonSelectedIndex(state: VQBState) {
  * @return the subpart of the pipeline that is currently active.
  */
 export function activePipeline(state: VQBState) {
-  return currentPipeline(state).slice(0, firstNonSelectedIndex(state));
+  const pipeline = currentPipeline(state);
+  if (!pipeline) {
+    return;
+  }
+  return pipeline.slice(0, firstNonSelectedIndex(pipeline, state.selectedStepIndex));
```
adimascio: You could probably use the optional call construct and remove the `if` (untested):

```suggestion
  return pipeline?.slice(0, firstNonSelectedIndex(pipeline, state.selectedStepIndex));
```

src/store/state.ts
```diff
@@ -148,13 +149,21 @@ function firstNonSelectedIndex(state: VQBState) {
...
 }
 
 /**
  * @param state current application state
  * @return the subpart of the pipeline that is currently inactive.
  */
 export function inactivePipeline(state: VQBState) {
-  return currentPipeline(state).slice(firstNonSelectedIndex(state));
+  const pipeline = currentPipeline(state);
```
adimascio: same remark as above :)

playground/app.js
```diff
@@ -166,24 +166,6 @@ class MongoService {
 
 const mongoservice = new MongoService();
 const mongoBackendPlugin = servicePluginFactory(mongoservice);
-const initialPipeline = [
```
adimascio: general query for this commit: could you maybe add a small 3-liner explanation of what is done in this commit?

tests/unit/pipeline-selector.spec.ts
```diff
@@ -0,0 +1,77 @@
+import { createLocalVue, shallowMount } from '@vue/test-utils';
```
adimascio: General comment: I would prefer folding this commit into the previous one, I would find that more consistent.

tests/unit/pipeline-selector.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      localVue,
+    });
+    const select = wrapper.find('select');
+    expect(select.exists()).toBeTruthy();
+    const options = select.findAll('option');
+    expect(options).toHaveLength(3);
+    expect(options.at(0).text()).toEqual('pipeline1');
+    expect(options.at(1).text()).toEqual('pipeline2');
+    expect(options.at(2).text()).toEqual('pipeline3');
+  });
```
adimascio: What about:

```javascript
    const entries = options.wrappers.map(o => o.text());
    expect(entries).toEqual(['pipeline1', 'pipeline2', 'pipeline3']);
```

?

tests/unit/pipeline-selector.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      store,
+      localVue,
+    });
+    const select = wrapper.find('select');
+    select
+      .findAll('option')
+      .at(2)
+      .setSelected();
+    select.trigger('input');
+    expect(store.state.vqb.currentPipelineName).toStrictEqual('pipeline3');
```
adimascio:
```suggestion
    expect(store.state.vqb.currentPipelineName).toEqual('pipeline3');
```

src/store/actions.ts
```diff
@@ -0,0 +1,14 @@
+import { ActionTree } from 'vuex';
+
+import { VQBState } from './state';
+
+const actions: ActionTree<VQBState, any> = {
```
adimascio: genuine question: do we need an action here? I though actions where primarily used for asynchronous mutations.

playground/app.js
```diff
@@ -329,6 +329,9 @@ async function buildVueApp() {
       code: function() {
         let activePipeline = this.$store.getters[VQBnamespace('activePipeline')];
         const pipelines = this.$store.getters[VQBnamespace('pipelines')];
+        if (!activePipeline) {
```
adimascio: I think this test could be moved one statement up

src/store/getters.ts
davinov: Oh, I didn't know that landed in Ts ! Nice to know, this will greatly simplify this PR :D

tests/unit/pipeline-selector.spec.ts
davinov: Done in 9cafa39; I also squashed with the one where I experimented the pipeline selector inside the Vqb component

src/store/actions.ts
davinov: > Actions _can_ contain arbitrary asynchronous operations.
(https://vuex.vuejs.org/guide/actions.html)
Mainly yes, but they also serve the purpose to compose multiple mutations.
In my understanding, it's a way to a extract higher-order functions. The only other way to factor multiple mutations would be to extract them as functions, call these functions in their mutation definition, and calls them also int he higher-order mutation that composes them. I seems easier to me tp use vuex actions for this.

tests/unit/pipeline-selector.spec.ts
davinov: Thanks for pointing that! I thought was mislead into a rough translation of chai's `eql` and `equal`. Turns out that jest does more than just checking for `==` and `===`.

playground/app.js
davinov: GitHub does not indicate which commit you're commenting on
I assume it's about dc92f572f5fb2479dd37bb2540a3ce6d80354f84 ?
I'll add a commentary

davinov: GitHub does not indicate which commit you're commenting on
I assume it's about 9cafa39 ?
The commit message is "PipelineSelector component in Vqb", I can add a comment, I thought the message was self-explainatory :thinking:

davinov: See 9cd5725 :)

davinov:
[(xx) ETQD intégrant le VQB, je ne peux pas faire d'erreurs d'ordre entre les actions setPipelines et setPipeline](https://trello.com/c/ZaY8UaSX/116-xx-etqd-int%C3%A9grant-le-vqb-je-ne-peux-pas-faire-derreurs-dordre-entre-les-actions-setpipelines-et-setpipeline)

playground/app.js
adimascio: Hmm, I can't find the commit. Mostly, what I meant is "describe why and how, rather than what" :)

src/store/actions.ts
adimascio: LGTM, thanks for the explanation!

adimascio:
Imports don't seem to be sorted correctly in `mutations.ts`, cf. https://app.circleci.com/jobs/github/ToucanToco/weaverbird/3752

playground/app.js
davinov: Here it is: https://github.com/ToucanToco/weaverbird/pull/444/commits/5932e21d277ade338801fc783b20d5002d6bd4ec
Doing modifications and fixup(squashing) has this drawback, it's difficult to follow what actually happened during the review :/

davinov:
@adimasci Feedback applied :)

The import was a mistake using the array option chaining: `arr?[0]` instead of `arr?.[0]`.

> Am I right in thinking that the pipeline selector won't be visible in Tucana? (and I think it should not be)
Indeed, the PipelineSelector is replaced in Tucana by the tabs by which we select the different datasets. Charge to Tucana to synchronize its current selected dataset with Weaverbird store. This should stay rather simple using the action `selectPipeline`.

> Could you add some closes #XXX comments in the commit so that github will be able to link the issue and the commits / PR?
Sure! Done in the comment of this commit: https://github.com/ToucanToco/weaverbird/pull/444/commits/d287843d970af8f21214d037857f78e0163475a9

PR #445, opened 2020-02-23
Playground :: make it static

rhuille:
## Playground :: make it static

This make the onboarding on the project easier: no need for a mongo db and a server. 

Also, it will allow use to make the playground live :)

-------

- add the module 'mingo' to the playground: this module allow to execute mongo aggregation pipeline on javascript object in browser
(link to mingo: https://github.com/kofrasa/mingo)

- Remove server side of playground: mongo db an server

- Adapte client-side with mingo: create a 'in browser mongo-like db'

- Add 'http-server' to yarn lock to run the playground
Now it runs with `http-server playground/dist --port=3000` (or `yarn playground`)


playground/dist/mingo.min.js
```diff
@@ -0,0 +1 @@
+../../node_modules/mingo/dist/mingo.min.js
```
davinov: Hm.. that's weird for a js file, isn't it? Looks like a symlink

playground/dist/index.html
```diff
@@ -111,6 +111,7 @@ <h1 class="vqb__drop-zone__message">
     <!-- built files will be auto injected -->
     <script src="https://cdn.jsdelivr.net/npm/vue/dist/vue.js"></script>
     <script src="https://unpkg.com/vuex@3.1.1/dist/vuex.js"></script>
+    <script src="/mingo.min.js"></script>
```
davinov: We could do the same as above, i.e. using a CDN to deliver mingo.
I think that would be better that to add this in the main package.json (we don't want our users to have to download mingo to use the weaverbird).
Another option would be to have a dedicated package.json for the playground.

That second option seems to me like the best idea, I see other dependencies that would be nice to exclude from today's deps (mongo, mongo-prebuilt, express,... everything related to the playground)

PR #446, opened 2020-02-24
Rollup step

vdestraitt:
Add a `rollup` steps that allows to aggregate data by hierarchical levels and stack the aggregated data of every level, based on a hierarchical columns (e.g. CITY < COUNTRY < CONTINENT )


src/lib/steps.ts
```diff
@@ -183,6 +183,22 @@ export type ReplaceStep = {
   to_replace: any[][];
 };
 
+export type RollupStep = {
+  name: 'rollup';
+  /** the list of hierarchical columns from lowest to highest level */
+  hierarchy: string[];
+  /** the list of columnns to aggregate, with related aggregation function to use */
+  aggregations: AggFunctionStep[];
+  /** Groupby columns if rollup has to bbe performed by groups */
```
davinov:
```suggestion
  /** Groupby columns if rollup has to be performed by groups */
```

src/lib/steps.ts
```diff
@@ -183,6 +183,22 @@ export type ReplaceStep = {
...
+  groupby?: string[];
+  /** To give a custom name to the output label column */
+  labelCol?: string;
+  /** To give a custom name to the output level column */
+  levelCol?: string;
+  /** To give a custom name to the output parent column */
+  parentCol?: string;
```
davinov: Should we name that `parentLabelCol`? Reading the description, I was unsure which column of the parent will be used to fill this one.

src/lib/translators/mongo.ts
```diff
@@ -290,6 +290,53 @@ function transformReplace(step: Readonly<S.ReplaceStep>): MongoStep {
   };
 }
 
+/** transform an 'replace' step into corresponding mongo steps */
```
davinov:
```suggestion
/** transform a rollup step into corresponding mongo pipeline steps */
```

src/lib/translators/mongo.ts
```diff
@@ -290,6 +290,53 @@ function transformReplace(step: Readonly<S.ReplaceStep>): MongoStep {
...
+        },
+      },
+      {
+        $project: project,
+      },
+    ];
+  }
+  return [
+    { $facet: facet },
+    { $project: { _vqbLevels: { $concatArrays: Object.keys(facet).map(col => $$(col)) } } },
```
davinov: May I suggest to call this variable `_vqbRollupLevels` so it's easier to understand which pipeline step it comes from when reading the output query?

src/lib/translators/mongo.ts
```diff
@@ -290,6 +290,53 @@ function transformReplace(step: Readonly<S.ReplaceStep>): MongoStep {
...
+      _id: 0,
+      ...Object.fromEntries(Object.keys(id).map(col => [col, `$_id.${col}`])),
+      ...Object.fromEntries(Object.keys(aggs).map(col => [col, 1])),
+      [step.labelCol ?? 'label']: `$_id.${elem}`,
+      [step.levelCol ?? 'level']: elem,
+    };
+    if (idx < step.hierarchy.length - 1) {
+      project[step.parentCol ?? 'parent'] = `$_id.${step.hierarchy[idx + 1]}`;
+    }
+    facet[`level_${idx}`] = [
```
davinov: :exclamation: 
Usually, when describing a hierarchy, the order is from the root level to the leaves
level_0 -> root
level_1 -> children
level_2 -> grand children
and so on

We also generally observe this in tables:

continent | country | city
| -- | -- | -- |
europe | | |
europe | germany | |
europe | germany | berlin

It's easier to read and explore, ad if follows a standard "tree-like" pattern.

OK for you to reverse the order of the`hierarchy` parameter? 
It should also naturally set the order of the rows in the output table like above.

davinov: (You'll have to modify the `id =` logic above)

src/lib/translators/mongo.ts
```diff
@@ -290,6 +290,53 @@ function transformReplace(step: Readonly<S.ReplaceStep>): MongoStep {
...
+        aggs[aggfStep.newcolumn] = {
+          [$$(aggfStep.aggfunction)]: $$(aggfStep.column),
+        };
+      }
+    }
+    const project: { [id: string]: string | number } = {
+      _id: 0,
+      ...Object.fromEntries(Object.keys(id).map(col => [col, `$_id.${col}`])),
+      ...Object.fromEntries(Object.keys(aggs).map(col => [col, 1])),
+      [step.labelCol ?? 'label']: `$_id.${elem}`,
```
davinov: It's just a matter of style a discoverability, but I'd feel better seeing all the default values declared up in the function, with their default values:
```
const labelCol = step.labelCol ?? 'label';
const levelCol = step.levelCol ?? 'level';
[...]
```
This way, we only have the default values in one place and it's easy to see them all at once.

tests/unit/templating.spec.ts
```diff
@@ -1024,4 +1024,72 @@ describe('Pipeline interpolator', () => {
...
+            newcolumn: 'value1',
+            aggfunction: 'sum',
+            column: 'bar',
+          },
+        ],
+      },
+    ]);
+  });
+
+  it('should interpolate complexe rollup steps if needed', () => {
```
davinov:
```suggestion
  it('should interpolate more complex rollup steps if needed', () => {
```

src/components/stepforms/RollupStepForm.vue
```diff
@@ -0,0 +1,130 @@
+<template>
+  <div>
+    <StepFormHeader :title="title" :stepName="this.editedStep.name" />
+    <MultiselectWidget
+      id="hierarchyColumnsInput"
+      v-model="editedStep.hierarchy"
+      name="Hierarchal columns (from lowest to highest level):"
```
davinov:
```suggestion
      name="Hierarchical columns (from lowest to highest level):"
```
Also a change here if you agree to reverse the order

src/components/stepforms/schemas/rollup.ts
```diff
@@ -0,0 +1,95 @@
+import { addNotInColumnNamesConstraint, StepFormType } from './utils';
+
+const schema = {
+  $schema: 'http://json-schema.org/draft-07/schema#',
+  title: 'Rollup step',
+  type: 'object',
+  properties: {
+    name: {
+      type: 'string',
+      enum: ['rollup'],
```
davinov: I think there is a `const` property in json schema that fits this case better :) 
https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values
It's from draft 6 and we're using draft 7 here, so it shouldn't be a problem.

src/components/stepforms/schemas/rollup.ts
```diff
@@ -0,0 +1,95 @@
...
+      },
+    },
+    aggregations: {
+      type: 'array',
+      items: {
+        type: 'object',
+        properties: {
+          column: {
+            type: 'string',
+            minLength: 1,
```
davinov: Would it be possible to have no aggregations columns?
I imagine people wanted to create a referential, and therefore having only labels and no value columns.
It shouldn't be a problem in the query I think, and it would be a shame to have to create a dummy column with only 0 for this to work.

src/components/stepforms/schemas/rollup.ts
```diff
@@ -0,0 +1,95 @@
...
+        placeholder: 'parent',
+      },
+    },
+  },
+  required: ['name', 'hierarchy', 'aggregations'],
+  additionalProperties: false,
+};
+
+export default function buildSchema(form: StepFormType) {
+  return addNotInColumnNamesConstraint(schema, 'aggregations.items.newcolumn', form.columnNames);
```
davinov: :exclamation: 
Shouldn't we also do this operation with `labelCol`, `levelCol` and `parentCol`?

docs/_docs/tech/steps.md
```diff
@@ -1176,6 +1176,180 @@ A replace step has the following strucure:
...
+      column: 'revenues'
+    },
+    {
+      newcolumn: 'average_cost',
+      aggfunction: 'avg', // aggregate using average
+      column: 'cost'
+    }
+   ],
+   groupby: ['date'], // optional, if needing to segment the rollup computation by group
+   labelCol: 'label', // optional, name of the outuput label column, 'label' by default
```
davinov:
```suggestion
   labelCol: 'label', // optional, name of the output label column, 'label' by default
```
And same ion the below lines

src/components/stepforms/schemas/rollup.ts
vdestraitt: Actually, I will delete this part even for aggregations. In this step we recreate the whole data structure, so there is no risk to delete an existing column... The only risk is to give the same name for the label column and the level column for exemple, but well... If you think it is important to prevent this, then we should overload the validate method of the base class StepForm.vue, as the `addNotInColumnNamesConstraint` checks unicity only against existing columns, not against other columns to be created.

vdestraitt:
@davinov thanks for the review. I created fix commits that mirror my original commits for ease of reading and so that I can easily fixup my orginal commits when you're fine. I will rebase when you approved so be prepared to approve twice then ;)

docs/_docs/tech/steps.md
```diff
@@ -1179,14 +1179,14 @@ A replace step has the following strucure:
...
-`hierarchy`, specified as a series of columns from lowest to highest level.
+`hierarchy`, specified as a series of columns from top to bottom level.
 The output data structure stacks the data of every level of the hierarchy,
 specifying for every row the label, level and parent in dedicated columns.
 
 ```javascript
 {
    name: 'rollup',
-   hierarchy: ['city', 'country', 'continent'], // Hierarchical columns, from lowest to highest level
+   hierarchy: ['city', 'country', 'continent'], // Hierarchical columns, from top to bottom level
```
davinov: Reverse order

PR #447, opened 2020-02-24
Multiselect tags overflow issue

zegonz:
This PR fix an overflow issue when there is more than one row of tag into multiselect.

before : 
image

Now it's cropped to display only one line and we show everything on click:
image

This PR fix also another css visual issue a background color on filter step :
image

Now : 
image


zegonz:
[Design VQB](https://trello.com/c/l6Lnk4PX/3854-design-vqb)

PR #448, opened 2020-02-25
Fix :: WidgetList item deletion

Dalanir:
As we remove and update element by their index in v-for, values are well updated but the dom does not render the right elements.

To fix that, we add unique ids but they can't be random cause it could lead to call `updateChildValue`, reinstantiating items because `children` computed properties will compute a new id as values change.

### DIFF
--- a/src/components/stepforms/widgets/List.vue
+++ b/src/components/stepforms/widgets/List.vue
@@ -2,7 +2,7 @@
   <div class="widget-list__container" :class="toggleClassErrorWarning">
     <label :for="id">{{ name }}</label>
     <div class="widget-list__body">
-      <div class="widget-list__child" v-for="(child, index) in children" :key="index">
+      <div class="widget-list__child" v-for="(child, index) in children" :key="child.uid">
         <span class="widget-list__component-sep" v-if="index > 0 && separatorLabel">{{
           separatorLabel
         }}</span>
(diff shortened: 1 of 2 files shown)


Dalanir:
[ETQConcepteur le delete d'un agrégat dans le VQB (step agregate) est correct](https://trello.com/c/V7WpjD23/3853-etqconcepteur-le-delete-dun-agr%C3%A9gat-dans-le-vqb-step-agregate-est-correct)

src/components/stepforms/widgets/List.vue
```diff
@@ -96,6 +96,7 @@ export default class ListWidget extends Mixins(FormWidget) {
       valueCopy.push(this.defaultChildValue);
     }
     return valueCopy.map(value => ({
+      uid: this.getUniqueId(value),
```
davinov: I don't think it's a good idea to put that in a computed property: at each evaluation loop, the id will change, triggering a destruction of the corresponding element.

davinov: Seeing the implementation of the function below, I understand that this is not an id that will change each time (like lodash's `uniqueId` but a deterministic computation from the row's values.
This also has drawbacks : at each modification of the row's value by the user (e.g. typing a letter in a text field), the whole row will be destroyed and recreated.

PR #452, opened 2020-02-26
Escape mongo fieldnames

davinov:
I suggest that `MongoTranslator` have a settable `escapeFieldName` function that will be used to replace all unauthorized characters from field names.
The mongo documentation suggest not to use `$` or `.` and provides different workarounds, such as replacing these characters by their unicode equivalent, or by special reserved strings. Another option would be to throw an exception if these characters are found, preventing the query to be translated.

As this is a database-specific rule, it seems logic to me that it stays in the `MongoTranslator`.

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -588,9 +588,13 @@ const mapper: Partial<StepMatcher<MongoStep>> = {
       },
     },
   }),
-  delete: (step: Readonly<S.DeleteStep>) => ({
-    $project: _.fromPairs(step.columns.map(col => [col, 0])),
-  }),
+
+  delete(this: Mongo36Translator, step: Readonly<S.DeleteStep>) {
+    return {
+      $project: _.fromPairs(step.columns.map(col => [this.escapeFieldName(col), 0])),
+    };
+  },
+
   domain: (step: Readonly<S.DomainStep>) => ({ $match: { domain: step.domain } }),
   duplicate: (step: Readonly<S.DuplicateColumnStep>) => ({
     $addFields: { [step.new_column_name]: $$(step.column) },
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -65,6 +65,7 @@ export function _simplifyAndCondition(filterAndCond: FilterComboAndMongo): Mongo
 }
 
 function buildMatchTree(
+  this: Mongo36Translator,
   cond: S.FilterSimpleCondition | S.FilterComboAnd | S.FilterComboOr,
   parentComboOp: ComboOperator = 'and',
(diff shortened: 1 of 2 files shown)


src/lib/translators/mongo.ts
```diff
@@ -588,9 +588,13 @@ const mapper: Partial<StepMatcher<MongoStep>> = {
       },
     },
   }),
-  delete: (step: Readonly<S.DeleteStep>) => ({
-    $project: _.fromPairs(step.columns.map(col => [col, 0])),
-  }),
+
+  delete(this: Mongo36Translator, step: Readonly<S.DeleteStep>) {
+    return {
+      $project: _.fromPairs(step.columns.map(col => [this.escapeFieldName(col), 0])),
```
davinov: `this.escapeFieldName` is very clear but quite long to write.
Does someone have a better naming idea? I really like the `$$` shortcut.

tests/unit/mongo.spec.ts
```diff
@@ -37,6 +46,24 @@ describe('Pipeline to mongo translator', () => {
...
+      { $match: { domain: 'test_cube' } },
+      {
+        $project: {
+          Man__DOLLAR_SIGN__: 1,
+          Re__DOLLAR_SIGN____DOLLAR_SIGN__gion: 1,
+        },
+      },
+      { $project: { _id: 0 } },
+    ]);
+  });
```
davinov: Testing this feature is gonna cause quite a duplication in tests. Does that feel OK to everyone?
It seems to be that it's the only option to test if every field name is correctly escpaed, without changing the actual tests too much.

src/lib/translators/mongo.ts
adimascio: `$$` is already used in our codebase (but maybe that's what you meant). `escapeField`, `quoteField`, `$field`, `$quote`, `field$`, `quote$` `escapeIdentifier`, `quoteIdentifier` or `quoteId` seems fine to me :)

We could even do things a little differently with something like this:

```typescript
function Field(name) { this.name = name; }
function $field(name) { return new Field(name) }
function isField(obj: any) : obj is Field{ return obj instanceof Field; }
```

and afterwards use `$field` everywhere explicitly and have a general postprocess visitor that would do the appropriate escaping. 

General question: where exactly should these fields be escaped? Only when used as keys or also when used as values?

davinov: @adimasci That was only for keys
But after discussions I has with @vdestraitt and @rhuille, we thought it may be better not to do this at the moment. (See internal slack thread: https://toucantoco.slack.com/archives/C08DRSQ5U/p1582819058152100)

davinov:
Closing: I finally don't think it's a good idea to allow this.

PR #454, opened 2020-02-28
Map domains to collection with a translator method

davinov:
Previously, we expected a map of domains to collections for the mongo
translator. This supposed that we already know all the available
domains and their correspondong collections. This is however not always
the case, e.g. if domains needs asynchronous loading.
I decided to replace it by a function, which cover the previous
case easily, and provide more possibilities.

### DIFF
--- a/tests/unit/mongo.spec.ts
+++ b/tests/unit/mongo.spec.ts
@@ -1,6 +1,11 @@
 import { Pipeline } from '@/lib/steps';
 import { getTranslator } from '@/lib/translators';
-import { _simplifyAndCondition, _simplifyMongoPipeline, MongoStep } from '@/lib/translators/mongo';
+import {
+  _simplifyAndCondition,
+  _simplifyMongoPipeline,
+  Mongo36Translator,
+  MongoStep,
+} from '@/lib/translators/mongo';
 
 describe('Mongo translator support tests', () => {
   const mongo36translator = getTranslator('mongo36');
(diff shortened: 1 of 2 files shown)


tests/unit/mongo.spec.ts
```diff
@@ -2103,3 +2108,52 @@ describe('Pipeline to mongo translator', () => {
...
+      {
+        name: 'join',
+        right_pipeline: rightPipeline,
+        type: 'left',
+        on: [['id', 'id']],
+      },
+    ];
+    const querySteps = mongo36translator.translate(pipeline);
+    const lookupStage = querySteps.find(s => s.$lookup);
+    expect(lookupStage).not.toBeUndefined();
```
adimascio:
```suggestion
    expect(lookupStage).toBeDefined();
```

tests/unit/mongo.spec.ts
```diff
@@ -2103,3 +2108,52 @@ describe('Pipeline to mongo translator', () => {
...
+      {
+        name: 'append',
+        pipelines: [pipelineBis],
+      },
+    ];
+
+    const querySteps = mongo36translator.translate(pipeline);
+
+    const lookupStage = querySteps.find(s => s.$lookup);
+    expect(lookupStage).not.toBeUndefined();
```
adimascio:
```suggestion
    expect(lookupStage).toBeDefined();
```

PR #455, opened 2020-03-02
Continuous deployment on heroku

adimascio:
This PR will implement mainly 3 things:

- a default "sane" collection for the playground to work on,
- standard `package.json` targets (e.g. `build` and `start`) so that most CD platforms (including heroku) will work (nearly) out of the box.
- add a link from the README to the live playground :)


package.json
```diff
@@ -17,9 +17,11 @@
   "scripts": {
     "codecov": "codecov --token=8e0bf2da-c4b9-435f-958e-446849d0d60e --file=./coverage/coverage-final.json",
     "lint": "eslint src/{components,lib,store}/**/*.{ts,vue} tests/**/*.ts",
+    "build": "yarn build-bundle",
```
rhuille: 👍

playground/server.js
```diff
@@ -252,7 +252,7 @@ function _testConnection(client) {
 }
 
 function setupApp(config) {
-  const client = new MongoClient(config.dburi, { useNewUrlParser: true });
+  const client = new MongoClient(config.dburi, { useNewUrlParser: true, useUnifiedTopology: true });
```
rhuille: Thanks !

playground/server.js
```diff
@@ -397,7 +397,7 @@ function parseCommandLine() {
 }
 
 function start(config) {
-  setupApp(config).listen(config.httpPort, function() {
+  setupApp(config).listen(process.env.PORT || config.httpPort, function() {
```
rhuille: This is for live playground right ? It might be good to specify on top of this file, that it is meant to be run on our local computer, but also on heroku for the live playground ?

adimascio: Sorry, I'm not sure to understand your comment :thinking:

davinov:
Thanks for this work @adimasci !
If we continue a bit the work of @rhuille with Mingo, we could even deploy the playground totally client-side, directly in GitHub pages? That could be a simpler alternative, maybe also safer (maybe we don't want random people to add data to the Heroku database?)

adimascio:
> Thanks for this work @adimasci !
> If we continue a bit the work of @rhuille with Mingo, we could even deploy the playground totally client-side, directly in GitHub pages? That could be a simpler alternative, maybe also safer (maybe we don't want random people to add data to the Heroku database?)

I think the live playground should be a fully working example and therefore I'd prefer to use a real mongodb server. For the safetyness-sake, the user configured on the heroku platform as read access only to the database.

davinov:
Thanks for the answer !
So that'll probably won't work with drop feature of the playground, we should deactivate it :)

PR #456, opened 2020-03-02
Export component :: autocomplete and multiInputText

zegonz:
Allow projets that use weaverbird to reuse form components so the UI stays visually coherent for those components :
- `Autocomplete`
- `MultiInputText`


zegonz:
after discussion this PR is not legit anymore. We don't need to expose those component

PR #458, opened 2020-03-02
New step :: evolution step

vdestraitt:
Mongo translation + user interface for a "compute evolution" step allowing to compute the row-level evolution of a value column based on a date column.


docs/_docs/user-interface/evolution.md
```diff
@@ -0,0 +1,80 @@
...
+#### Where to find this step?
+
+- Widget `Compute`
+- Search bar
+
+#### Options reference
+
+<img src="../../img/docs/user-interface/evolution_step_form.jpg" width="350" />
+
+- `Date column`: the date column (must be of type date) that will bbe used as a
```
rhuille:
```suggestion
- `Date column`: the date column (must be of type date) that will be used as a
```

rhuille: Why don't we automatically parse the date in this step ? It will save the step to systematically "parse date" before.

src/components/stepforms/schemas/evolution.ts
```diff
@@ -0,0 +1,70 @@
...
+        placeholder: 'Select columns',
+      },
+    },
+    absEvolColName: {
+      type: 'string',
+      minLength: 1,
+      title: 'Column name for the evolution in absolute value',
+      description: 'Column name for the evolution in absolute value',
+    },
+    pctEvolColName: {
```
rhuille: I think it is weird to produce 2 output columns, when maybe only one is useful. What if we output only one column. It can be absolute value or percentage value depending on a other parameter like: "compute_in_percentage: true / false" (for instance).

docs/_docs/tech/steps.md
```diff
@@ -620,6 +620,169 @@ This step is meant to duplicate a column.
...
+### `evolution` step
+
+Use this step if you need to compute the row-by-row evolution of a value column,
+based on a date column. It will output 2 columns: one for the evolution in
+absolute value, the other for the evolution in percentage.
+
+You must be careful that the computation is scoped so that there are no dates
+duplicates (so that any date finds no more than one previous date). That means
+that you may need to specify index columns to make any date unique by index. You
+should use the `uniqueIndexCols` parameter to specifiy what series of columns
```
rhuille: Very clear ! Thanks :)

src/components/stepforms/schemas/evolution.ts
```diff
@@ -0,0 +1,70 @@
...
+    },
+    uniqueIndexCols: {
+      type: 'array',
+      minItems: 0,
+      items: {
+        type: 'string',
+        minLength: 1,
+      },
+      title: 'uniqueIndexCols',
+      description: 'Index columns',
```
rhuille: I am not sure of the naming: `uniqueIndexCols`. If we display "Index columns" why do we use `uniqueIndexCols` ? And also, please do not use abbreviation, here 'Col'. 
For these reasons I suggest to use `IndexColumns` instead of `uniqueIndexCols`.




disapprove
--

docs/_docs/tech/steps.md
```diff
@@ -620,6 +620,169 @@ This step is meant to duplicate a column.
...
+| 2019-07 | 81    | 2              | 0.02531645569620253   |
+| 2019-08 | 77    | -4             | -0.04938271604938271  |
+| 2019-09 | 75    | -2             | -0.025974025974025976 |
+| 2019-11 | 78    |                |                       |
+| 2019-12 | 88    | 10             | 0.1282051282051282    |
+
+#### Example 2: Error on duplicate dates
+
+If 'COUNTRY' is not specified as index column, the computation will not be
+scoped by country. Then there are duplicate dates in the "DATE" columns.
```
rhuille:
```suggestion
scoped by country. Then there are duplicate dates in the "DATE" columns which is prohibited.
```

PR #459, opened 2020-03-02
Change how changes are propagated from JoinColumns widget

davinov:
Previously, the initial prop was copied once at component's creation.
The component's internal state was hen modified by the component.
A watcher on the internal state propagated this changes upwards.

This approach had flaws:
- the component holds an internal state, but shouldn't (the source of
truth should be the form complete value object)
- if the form's value is modified outside the component, the
modification will never appear in the component
- additional deep watchers are added, which are quite useless

I restored the usual way of passing down props to components, and emit
up events to communicate changes requests.

Same logic than #457

### DIFF
--- a/src/components/stepforms/widgets/JoinColumns.vue
+++ b/src/components/stepforms/widgets/JoinColumns.vue
@@ -2,8 +2,7 @@
   <div class="widget-join-column__container">
     <AutocompleteWidget
       id="leftOn"
-      @input="setRightColumn()"
-      v-model="joinColumns[0]"
+      v-model="leftOnColumn"
       placeholder="Current dataset column"
       :options="columnNames"
       :data-path="`${dataPath}[0]`"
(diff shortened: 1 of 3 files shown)


src/components/stepforms/widgets/JoinColumns.vue
```diff
@@ -50,19 +48,26 @@ export default class JoinColumns extends Vue {
...
 
-  joinColumns: string[] = [...this.value];
+  get leftOnColumn() {
+    return this.value[0];
+  }
+
+  set leftOnColumn(newLeftOnColumn) {
+    // If no right column, set it to the same as left column (smart default)
+    const newRightColumn = this.rightOnColumn === '' ? newLeftOnColumn : this.rightOnColumn;
+    this.update([newLeftOnColumn, newRightColumn]);
```
bloodstorms: 👍

PR #463, opened 2020-03-03
formula step :: update doc (escaping column names with whitespaces)

vdestraitt:
Everything is in the title


docs/_docs/tech/steps.md
```diff
@@ -700,8 +700,10 @@ operator).
 
 ### `formula` step
 
-Add a computation based on other columns or on values.
-Column names must not be escaped. Strings have to be escaped with quotes.
+Add a computation based on a formula. Usually column names do not need to be
+escaped, unless they include whitespaces, in such a case you need to use
+brackets '[]' (e.g. \[myColumn]). Any characters string escaped with quotes
```
davinov:
```suggestion
brackets '[]' (e.g. `[myColumn]`). Any characters string escaped with quotes
```

docs/_docs/tech/steps.md
```diff
@@ -713,7 +715,7 @@ Column names must not be escaped. Strings have to be escaped with quotes.
 }
 ```
 
-#### Example
+#### Example 1
```
davinov: Not a big fan of numerating examples, better create a `#### Examples` section and give a meaningful title to each, like
`##### Basic usage` and 
`##### Column names with white-spaces`

docs/_docs/user-interface/formula.md
```diff
@@ -6,15 +6,11 @@ permalink: /docs/formula/
 ### Formula
 
 You can use this step to create a column as a formula based on other columns.
-A column can be referenced by its name without quotes or any other escaping
-character.
+A column can be referenced by its name without quotes unless they include
+whitespaces, in such a case you need to use brackets '[]' (e.g. \[myColumn]).
```
davinov:
```suggestion
white-spaces, in such a case you need to use brackets '[]' (e.g. `[myColumn]`).
```

docs/_docs/tech/steps.md
```diff
@@ -700,8 +700,10 @@ operator).
 
 ### `formula` step
 
-Add a computation based on other columns or on values.
-Column names must not be escaped. Strings have to be escaped with quotes.
+Add a computation based on a formula. Usually column names do not need to be
+escaped, unless they include whitespaces, in such a case you need to use
```
davinov:
```suggestion
escaped, unless they include white-spaces, in such a case you need to use
```

docs/_docs/user-interface/formula.md
```diff
@@ -6,15 +6,11 @@ permalink: /docs/formula/
...
 
-**Warning: we do not support column names including whitespaces at the moment.
-So please apply a rename step beforehand if needed.** (For example rename
-`my column` into `my_column`)
-
-You can also specify a text escaped by quotes. It will then write the specified
-text in every row of the new column.
+Any characters string escaped with quotes (simple or double) will be considered
+as a string.
```
davinov: That's too bad we have to repeat that in the tow docs :man_shrugging:

docs/_docs/tech/steps.md
vdestraitt: @davinov I think I need to escape the first '[' in markdown to avoid the pattern being considered as a link (my linter reminded me about this :))

docs/_docs/tech/steps.md
vdestraitt: Ok to name the examples but I will stick to the enumeration to remain homogeneous with the other docs pages. So I will write: `Example 1: Basic usage`
If it's something we really want to change consistently, we should do that in a dedicated PR

docs/_docs/user-interface/formula.md
vdestraitt: `white-spaces` is a CSS property ;) https://developer.mozilla.org/fr/docs/Web/CSS/white-space
`whitespace` is right if we choose to believe in Wikipedia: https://en.wikipedia.org/wiki/Whitespace_character

docs/_docs/tech/steps.md
vdestraitt: Also, the markdown seems fine when I preview it in my editor. Tell me if I miss something

davinov: Yes, it's more to display it in a code block like this: `[myColumn]`
Not a big deal though

docs/_docs/tech/steps.md
davinov: Not a big deal ;)

PR #468, opened 2020-03-05
reset pagination after some mutations

adimascio:
When navigating in or modifying the pipeline, the pagination needs
to be reset to avoid inconsistencies.

e.g consider a pipeline with 4 steps and a dataset on page 3. If the last step
is a filter step and the user removes it, the page `3` does not have the same
meaning. The only consistent behaviour is to go back to page 1 with the new
dataset.

### DIFF
--- a/src/store/mutations.ts
+++ b/src/store/mutations.ts
@@ -3,6 +3,7 @@
  */
 
 import Vue from 'vue';
+import { MutationTree } from 'vuex';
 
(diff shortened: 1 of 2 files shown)


Dalanir:
[ETQC si je cree un nouveau step, je reset l'offset](https://trello.com/c/5PAOjR3X/3881-etqc-si-je-cree-un-nouveau-step-je-reset-loffset)

Issue #462, opened 2020-03-03
Propagate props changes correctly in form widgets


davinov:
Actually, in widgets, he initial prop is copied once at component's creation.
The component's internal state is then modified by the component using `v-model`s.
A watcher on the internal state propagate this changes upwards.

This approach has flaws:
- the component holds an internal state, but shouldn't (the source of truth should be the form complete value object)
- if the form's value is modified outside the component, the modification will never appear in the component (this caused a list item deletion bug)
- additional deep watchers are added, which are quite useless

I think all widgets are affected, but here is the list of priority widgets (because they are used in lists, therefore are prone to the bug):
- [x] Aggregation -> Fixed in #457 
- [x] JoinColumn -> Fixed in #459 
- [x] columnPicker
- [x] widgetFilterSimpleCondition
- [x] replaceWidget
- [x] widgetSortColumn
- [x]  Autocomplete widget

davinov:
Note: we should pay attention to emit a default value when the component is created without any value.

PR #472, opened 2020-03-11
fix :: replaceWidget change props behavior

zegonz:
the problem is explain here => https://github.com/ToucanToco/weaverbird/issues/462

### DIFF
--- a/src/components/stepforms/widgets/Replace.vue
+++ b/src/components/stepforms/widgets/Replace.vue
@@ -17,8 +17,8 @@
   </div>
 </template>
 <script lang="ts">
-import _ from 'lodash';
 import { ErrorObject } from 'ajv';
+import isEqual from 'lodash/isEqual';
 import { Component, Prop, Vue } from 'vue-property-decorator';
 
 import InputTextWidget from './InputText.vue';
@@ -43,7 +43,7 @@ export default class ReplaceWidget extends Vue {
   errors!: ErrorObject[];
 
   created() {
-    if(_.isEqual(this.value, ['', ''])) {
+    if (isEqual(this.value, ['', ''])) {
       this.update(this.value);
     }
   }
--- a/src/components/stepforms/widgets/Replace.vue
+++ b/src/components/stepforms/widgets/Replace.vue
@@ -2,14 +2,14 @@
   <div class="widget-to-replace__container">
     <InputTextWidget
       id="valueToReplace"
-      v-model="toReplace[0]"
+      v-model="valueToReplace"
       placeholder="Value to replace"
       :data-path="`${dataPath}[0]`"
       :errors="errors"
     />
     <InputTextWidget
       id="newValue"
-      v-model="toReplace[1]"
+      v-model="newValueToReplace"
       placeholder="New value"
       :data-path="`${dataPath}[1]`"
       :errors="errors"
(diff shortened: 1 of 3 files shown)


src/components/stepforms/widgets/Replace.vue
```diff
@@ -42,13 +41,28 @@ export default class ReplaceWidget extends Vue {
   @Prop({ type: Array, default: () => [] })
   errors!: ErrorObject[];
 
-  toReplace: any[] = [...this.value];
+  created() {
+    this.update(this.value);
```
davinov: Only when the value is the default one?
Otherwise it's unnecessary I think
```suggestion
  if(isEqual(this.value, ['', ''])) {
    this.update(this.value);
  }
```

tests/unit/replace-widget.spec.ts
```diff
@@ -30,24 +30,46 @@ describe('Widget ReplaceWidget', () => {
...
+  it('should emit "input" event with correct updated values when input valueToReplace is updated', async () => {
     const wrapper = shallowMount(ReplaceWidget, {
+      propsData: {
+        value: ['yolo', 'bim'],
+      },
+      store: emptyStore,
+      localVue,
+      sync: false,
+    });
+    await localVue.nextTick();
```
davinov: Not sure this nextTick would be necessary if we avoid sending the first unnecessary input

tests/unit/replace-widget.spec.ts
```diff
@@ -30,24 +30,46 @@ describe('Widget ReplaceWidget', () => {
...
+    wrapper
+      .findAll('InputTextWidget-stub')
+      .at(0)
+      .vm.$emit('input', 'foo');
+    expect(wrapper.emitted().input[1][0]).toEqual(['foo', 'bim']);
```
davinov: And the it will be the first, and not the second event emission

tests/unit/replace-widget.spec.ts
```diff
@@ -30,24 +30,46 @@ describe('Widget ReplaceWidget', () => {
...
+  });
+
+  it('should emit "input" event with correct updated values when input newValueToReplace is updated', async () => {
+    const wrapper = shallowMount(ReplaceWidget, {
+      propsData: {
+        value: ['yolo', 'bim'],
```
davinov: Same than the test before

tests/unit/replace-widget.spec.ts
```diff
@@ -30,24 +30,46 @@ describe('Widget ReplaceWidget', () => {
       store: emptyStore,
       localVue,
       sync: false,
-      data: () => {
-        return { toReplace: ['foo', 'bar'] };
+      propsData: {
+        value: ['foo', 'bar'],
```
davinov: So much cleaner :tada: :clap:

tests/unit/replace-widget.spec.ts
```diff
@@ -30,24 +30,46 @@ describe('Widget ReplaceWidget', () => {
...
     await localVue.nextTick();
-    expect(wrapper.emitted().input).toBeDefined();
-    expect(wrapper.emitted().input[1]).toEqual([['foo', 'bar']]);
+    wrapper
+      .findAll('InputTextWidget-stub')
+      .at(1)
+      .vm.$emit('input', 'bar');
+    expect(wrapper.emitted().input[1][0]).toEqual(['yolo', 'bar']);
   });
 });
```
davinov: I think that file misses a test where we don't set any prop (default case), and test the initial default value emission.

zegonz:
@davinov thx to your review I pushed all modifications if you can have a look it will be perfect !

davinov:
Thanks @zegonz 
Can you just import the isEqual util ?
`import isEqual from 'lodash/isEqual``
It avoids importing all of lodash into the bundle. We want weaverbird to stay slim :)

Thanks!

PR #476, opened 2020-03-18
Fix :: sort step change props propagation

zegonz:
the problem is described here => #462

### DIFF
--- a/src/components/stepforms/widgets/SortColumn.vue
+++ b/src/components/stepforms/widgets/SortColumn.vue
@@ -3,16 +3,15 @@
     <AutocompleteWidget
       id="columnInput"
       :options="columnNames"
-      v-model="sort.column"
+      v-model="sortColumn"
       name="Column"
       placeholder="Enter a column"
-      @input="setSelectedColumns({ column: sort.column })"
       :data-path="`${dataPath}[0]`"
       :errors="errors"
     />
     <AutocompleteWidget
       id="sortOrderInput"
-      v-model="sort.order"
+      v-model="sortOrder"
       name="Order"
       :options="['asc', 'desc']"
       placeholder="Order by"
@@ -23,8 +22,7 @@
 </template>
 <script lang="ts">
 import { ErrorObject } from 'ajv';
-import _ from 'lodash';
-import { Component, Prop, Vue, Watch } from 'vue-property-decorator';
+import { Component, Prop, Vue } from 'vue-property-decorator';
 
 import { SortColumnType } from '@/lib/steps';
 import { VQBModule } from '@/store';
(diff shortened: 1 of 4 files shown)


tests/unit/sort-column-widget.spec.ts
```diff
@@ -57,10 +57,16 @@ describe('Widget sort column', () => {
     const wrapper = shallowMount(SortColumnWidget, {
       store: emptyStore,
       localVue,
+      propsData: {
+        value: { column: 'bar', order: 'desc' },
+      },
```
davinov: It would be nice to add `sync: false` to these tests.

tests/unit/sort-column-widget.spec.ts
```diff
@@ -57,10 +57,16 @@ describe('Widget sort column', () => {
...
     await localVue.nextTick();
+    wrapper
+      .findAll('AutoCompleteWidget-stub')
+      .at(0)
+      .vm.$emit('input', 'year');
     expect(wrapper.emitted().input).toBeDefined();
-    expect(wrapper.emitted().input[1]).toEqual([{ column: 'bar', order: 'desc' }]);
+    expect(wrapper.emitted().input[0]).toEqual([{ column: 'year', order: 'desc' }]);
   });
 });
```
davinov: I think it misses a test for the update of the `sort` property.

zegonz:
@david I squash the commits with a better message and two more test as well.

I also add emit initial value on created to fix an issue => the editedStep was empty

src/components/stepforms/widgets/SortColumn.vue
```diff
@@ -55,17 +53,37 @@ export default class SortColumnWidget extends Vue {
   errors!: ErrorObject[];
 
   @VQBModule.Getter columnNames!: string[];
-
   @VQBModule.Mutation setSelectedColumns!: MutationCallbacks['setSelectedColumns'];
 
-  sort: SortColumnType = { ...this.value };
+  created() {
+    this.update(this.value);
```
davinov: I have the exact same concern than https://github.com/ToucanToco/weaverbird/pull/472

zegonz: not ok with that because in that case will always have a value to emit on created, the column proprety that was selected

zegonz:
[sortColumn fix propagation props](https://trello.com/c/0q70VMfY/240-sortcolumn-fix-propagation-props)

src/components/stepforms/widgets/SortColumn.vue
zegonz: @davinov What do you think ?

zegonz: otherwise I could merge that PR

davinov: Ok I see what you mean. I think the logic should be this one:
- if we do have a selected in column in the data table, then I expect the component to emit a new value with the selected column filled
- if we do NOT have a selected column in the data table, then:
  - if the value of the component is the default one, I expect the component to emit it
  - if the value of the component is not the default one, I expect the component NOT to emit anything, to avoid overriding this value

zegonz: After discussion the sortColumn widget has no need to emit this event at created to his parent because sortStepForm has already this data

PR #478, opened 2020-03-23
In-app doc :: add link to supported date formats in todate and fromdate steps UI

vdestraitt:
image

### DIFF
--- a/src/components/stepforms/widgets/InputText.vue
+++ b/src/components/stepforms/widgets/InputText.vue
@@ -1,6 +1,11 @@
 <template>
   <div class="widget-input-text__container" :class="toggleClassErrorWarning">
-    <label v-if="name" :for="id">{{ name }}</label>
+    <div class="widget-input-text__label">
+      <label v-if="name" :for="id">{{ name }}</label>
+      <a v-if="docUrl" :href="docUrl" target="_blank" rel="noopener">
+        <i class="fas fa-question-circle" />
+      </a>
+    </div>
     <input
       :id="id"
       :class="elementClass"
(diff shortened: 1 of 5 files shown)


src/components/stepforms/widgets/InputText.vue
```diff
@@ -1,6 +1,11 @@
 <template>
   <div class="widget-input-text__container" :class="toggleClassErrorWarning">
-    <label v-if="name" :for="id">{{ name }}</label>
+    <div class="widget-input-text__header">
+      <label v-if="name" :for="id">{{ name }}</label>
+      <a v-if="docUrl" :href="docUrl" target="_blank">
```
davinov: I think we should add [rel="noopener"](https://mathiasbynens.github.io/rel-noopener/) to links with `target="_blank"`.

src/components/stepforms/widgets/InputText.vue
```diff
@@ -1,6 +1,11 @@
 <template>
   <div class="widget-input-text__container" :class="toggleClassErrorWarning">
-    <label v-if="name" :for="id">{{ name }}</label>
+    <div class="widget-input-text__header">
```
davinov: `header` is maybe not the best identifier here. It could be below, or aside. Would you be OK with `label` instead?

tests/unit/input-text-widget.spec.ts
```diff
@@ -27,6 +27,18 @@ describe('Widget Input Text', () => {
     expect(labelWrapper.text()).toEqual('Stark');
   });
 
+  it('should not have a hyperlink element if prop "docUrl" is undefined', () => {
+    const wrapper = shallowMount(InputTextWidget);
+    expect(wrapper.find('a').exists()).toBeFalsy();
+  });
+
+  it('should have a label a hyperlink element if prop "docUrl" is defined', () => {
```
davinov:
```suggestion
  it('should have an hyperlink element if prop "docUrl" is defined', () => {
```

PR #488, opened 2020-03-26
sort step :: fix initialisation

vdestraitt:
Before this PR, at creation of the sort step form component, editedStep was
always overwritten with a default value.

We only want to suggest a default value when the step has never been edited yet

### DIFF
--- a/tests/unit/sort-step-form.spec.ts
+++ b/tests/unit/sort-step-form.spec.ts
@@ -51,12 +51,34 @@ describe('Sort Step Form', () => {
 
   runner.testResetSelectedIndex();
 
-  it('should suggest ordering by the column selected in the data table', function() {
+  it('should suggest the selected column as the column to be sorted', function() {
     const wrapper = runner.shallowMount({
       selectedColumns: ['selectedColumn'],
     });
     expect(wrapper.find('listwidget-stub').props().value).toEqual([
       { column: 'selectedColumn', order: 'asc' },
+    ]);
+  });
+
+  it('should not suggest a column to be sorted if no column is selected', function() {
+    const wrapper = runner.shallowMount({});
+    expect(wrapper.find('listwidget-stub').props().value).toEqual([{ column: '', order: 'asc' }]);
+  });
+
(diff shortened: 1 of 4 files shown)


tests/unit/sort-step-form.spec.ts
```diff
@@ -60,6 +60,28 @@ describe('Sort Step Form', () => {
     ]);
   });
 
+  it('should not suggest a column to be sorted if no column is selected', function() {
+    const wrapper = runner.shallowMount({});
+    expect(wrapper.find('listwidget-stub').props().value).toEqual([{ column: '', order: 'asc' }]);
```
rhuille:
```suggestion
    expect(wrapper.find(Listwidget).props().value).toEqual([{ column: '', order: 'asc' }]);
```
If you import `Listwidget` in the test you can use this cool syntax.
(This syntax is used on the "common tips" of Vue test utils: https://vue-test-utils.vuejs.org/guides/common-tips.html)

tests/unit/sort-step-form.spec.ts
```diff
@@ -60,6 +60,28 @@ describe('Sort Step Form', () => {
...
+      {
+        propsData: {
+          initialStepValue: {
+            name: 'sort',
+            columns: [{ column: 'amazing', order: 'desc' }],
+          },
+        },
+      },
+    );
+    expect(wrapper.find('listwidget-stub').props().value).toEqual([
```
rhuille:
```suggestion
    expect(wrapper.find(Listwidget).props().value).toEqual([
```

tests/unit/sort-step-form.spec.ts
vdestraitt: Thank you for the comment.

I'm not very convinced to implement that for several reasons:

- I find it kind of a waste to import a full component just to make the naming a bit fancier (but not that much more understandable than the same name with just a '-stub' suffix
- In my opinion we would be less explicit using the component name, as we do manipulate stubs, not real components
- At the end we are talking about details and I would be OK to change, but in that case I would prefer to do it in a separate PR, in all test files to keep a coherence

PR #489, opened 2020-03-26
Replace id of widgets by class

### LINKED ISSUE #487: Widget: remove id
Remove `id` of widgets. They are used several times in the project.

image

solution: passe these id into class.

rhuille:
Before: id of widget are not unique
Now: we replace id by class so there is not unicity problem anymore

- replace id from widgets template by class
- remove id from all widgets props
- replace reference to id in spec files by class
- remove reference to id in label (attribute "for")
   In the input wigget I preserved the focus behavior by triggering the focus of input in js.

closes #487

### DIFF
--- a/src/components/stepforms/AggregateStepForm.vue
+++ b/src/components/stepforms/AggregateStepForm.vue
@@ -2,7 +2,7 @@
   <div>
     <StepFormHeader :title="title" :stepName="this.editedStep.name" />
     <MultiselectWidget
-      id="groupbyColumnsInput"
+      class="groupbyColumnsInput"
       v-model="editedStep.on"
       name="Group rows by..."
       :options="columnNames"
@@ -13,7 +13,7 @@
     />
     <ListWidget
       addFieldName="Add aggregation"
-      id="toremove"
+      class="toremove"
       name="And aggregate..."
       v-model="aggregations"
       :defaultItem="defaultAggregation"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fixed bad sort step from initialization
 - Fixed columnPicker to prevent the mutation of the props
+- Fixed the no uniqueness of widgets ids by replacing them by class
 
 ## [0.15.0] - 2020-03-24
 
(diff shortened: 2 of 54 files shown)


src/components/stepforms/AggregateStepForm.vue
```diff
@@ -2,7 +2,7 @@
   <div>
     <StepFormHeader :title="title" :stepName="this.editedStep.name" />
     <MultiselectWidget
-      id="groupbyColumnsInput"
```
bloodstorms: Hum, In AutocompleteWidget, `id` is used as a props, i think it will break something to rename it `class`?

bloodstorms:
[ID in vqb](https://trello.com/c/jjbdkzq9/250-id-in-vqb)

src/components/stepforms/AggregateStepForm.vue
rhuille: Yes you are right: here is the line where I delete it: https://github.com/ToucanToco/weaverbird/pull/489/files#diff-0897a479bdd646438d8b6234e5d0b092L33

bloodstorms: Ok, it stay the `:for="id"` here https://github.com/ToucanToco/weaverbird/pull/489/files#diff-0897a479bdd646438d8b6234e5d0b092R3
But I don't know what it's used for.

rhuille: It's used to specifies which form element a label is bound to. I will find a way to overcome this.

vdestraitt:
@rhuille thank you !

It raises a lot of errors like those ones though, where `id` is used in `for` attributes of <label> balises: 
image

This pattern occurs in:

- `Autocomplete.vue`
- `CodeEditorWidget.vue`
- `InputText.vue`
- `List.vue`
- `Multiselect.vue`

Also:
 - you forgot 2 occurrences of id being passed to `InputTextWidget` in `AddTextColumnStepForm.vue`
 - if we do not use the id prop anymore, we should remove it in the compoents:
  - `ColumnPickert.vue`
  - `List.vue`
  - `MultiInputText.vue`
  - `Multiselect.vue`
- The doc `new-step.md` should be edited

src/components/stepforms/AggregateStepForm.vue
rhuille: Apparently the label is bind to nothing so I can remove it :)

bloodstorms:
897a549 => I just fix the conflit due to the PR on the columnPicker

PR #492, opened 2020-03-30
Improve Popover interface

rhuille:
**Before**: to open the popover you set active props to “true” (and to close, you set active to “false”)
This is an issue, because the component is alway mounted. As popover is used twice (action menu and datatype menu) in column menu this may cause severe performance issue.
**Now**: to open the popover you use v-if (to not mount the component when not visible) or v-show (if you are ok to mount the component when not visible)
Notes:
  - The v-if or v-show cannot be use directly on the popober.It should be set on a safe html parent. This parent is used to set its position.
  - Update interface of all components that uses popover:
    - `DataTypesMenu` (inside `DataViewer`)
    - `ActionMenu`  (inside `DataViewer`)
    - `ActionToolbarButton` (inside `ActionToolbar`)

### DIFF
--- a/src/components/Popover.vue
+++ b/src/components/Popover.vue
@@ -1,12 +1,14 @@
 <template>
-  <div :class="{ popover: true, 'popover--active': isActive }" :style="elementStyle">
-    <slot />
-  </div>
+  <span v-if="visible" class="popover-container">
+    <div class="popover" :style="elementStyle">
+      <slot />
+    </div>
+  </span>
 </template>
 <script lang="ts">
 import _ from 'lodash';
 import Vue from 'vue';
-import { Component, Prop, Watch } from 'vue-property-decorator';
+import { Component, Prop } from 'vue-property-decorator';
 
 import { Alignment } from '@/components/constants';
 import * as DOMUtil from '@/components/domutil';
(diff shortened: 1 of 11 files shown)


src/components/Popover.vue
```diff
@@ -29,7 +29,19 @@ interface ElementPosition {
 @description
 Implement a popover component
 
-@params {Boolean} [active = false] toggles visibility
+Use the popover inside a "safe" parent.
```
davinov, line 32: What do you mean by safe?

src/components/Popover.vue
```diff
@@ -29,7 +29,19 @@ interface ElementPosition {
...
+To toggle the visibility use a `v-if` or a `v-show` on it's parent of any of it's ancestor.
+Warning: do not use the v-if or v-show directly on the popover component
+
+For example:
+```
+<span class="popover-container">
+  <popover>
+    <div> This will be displayed insidthe popover<div>
+  </popover>
+</span>
```
davinov, line 42: This example does not showcase any v-if or v-show ; should we?

src/components/Popover.vue
```diff
@@ -29,7 +29,19 @@ interface ElementPosition {
...
-@params {Boolean} [active = false] toggles visibility
+Use the popover inside a "safe" parent.
+To toggle the visibility use a `v-if` or a `v-show` on it's parent of any of it's ancestor.
+Warning: do not use the v-if or v-show directly on the popover component
+
+For example:
+```
+<span class="popover-container">
+  <popover>
+    <div> This will be displayed insidthe popover<div>
```
davinov, line 40:
```suggestion
    <div>This will be displayed inside the popover<div>
```

src/components/Popover.vue
rhuille: A parent that should never be removed... I should specify that instead of the not clear word "safe" :)

src/components/Popover.vue
rhuille: I think so ! Thanks :)

PR #493, opened 2020-04-01
Components :: create component FilterEditor

bloodstorms:
Create an intermediate component (`FilterEditor`) between `FilterStepForm` and `ConditionsEditor`.

`FilterEditor` contains conditionsEditor + filterSimpleCondition + all data transformation logic
to permit the export of the interface with all the its logical, but without filterStep env.

PR #494, opened 2020-04-02
Filter editor :: fix missing error prop + add columnNames props

bloodstorms:
- Filter editor :: add missing error prop
   It was broken when added the new component FilterEditor here #493

- Filter editor :: add columnNames prop
   Add columnNames prop to permit to pass a list of column names from a dataset outside weaverbird

- Fix a CSS style not apply

### DIFF
--- a/src/components/stepforms/widgets/FilterSimpleCondition.vue
+++ b/src/components/stepforms/widgets/FilterSimpleCondition.vue
@@ -201,7 +201,7 @@ export default class FilterSimpleConditionWidget extends Vue {
   }
 }
 
-.multiselect__tags {
+.filter-form-simple-condition__container /deep/ .multiselect__tags {
   border-radius: 0;
   border: none;
 }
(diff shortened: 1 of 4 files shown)


src/components/stepforms/widgets/FilterSimpleCondition.vue
```diff
@@ -87,13 +87,19 @@ export default class FilterSimpleConditionWidget extends Vue {
...
+  columnNames!: string[];
+
   @Prop({ type: String, default: '' })
   dataPath!: string;
 
   @Prop({ type: Array, default: () => [] })
   errors!: ErrorObject[];
 
-  @VQBModule.Getter columnNames!: string[];
+  @VQBModule.Getter('columnNames') columnNamesGetter!: string[];
```
davinov, line 102: Hm...This is gonna cause an error in Tucana in the console (like `getter columnNames` does not exist).
Let's keep that way for now and ignore that.

Could you rather name it `columnNamesFromStore`? Then you can use a computed property that will use the prop is present and fallback to the store one.
In order to do that, avoid using any default value for the prop.

PR #495, opened 2020-04-03
Tech docs :: Documented exported UI components 

davinov:
Along with a few small others improvements, I added a "UI components" page to tech docs.
This will come handy to explain why components are exported, and how to use them.

It will be enriched, e.g. with this PR: #494  that exports a new component.

### DIFF
--- a/docs/_docs/tech/ui-components.md
+++ b/docs/_docs/tech/ui-components.md
@@ -46,6 +46,8 @@ The `Vqb` component consists is three sub-components, that you can use as well t
 
 ## Utility components
 
-Weaverbird also exports some of its internal components so they can be reused elsewhere:
+Weaverbird also exports some of its internal components so they can be reused elsewhere.
 
 - `FilterEditor` is the main component of the Filter step. It can display and edit trees of filtering conditions.
+
+> These components are not directly part of the public API. Changes can happen between versions.
--- a/docs/_docs/tech/ui-components.md
+++ b/docs/_docs/tech/ui-components.md
@@ -44,3 +44,8 @@ The `Vqb` component consists is three sub-components, that you can use as well t
 - `DataViewer` is the main right panel, where the data table resides and the associated actions.
 - `PipelineSelector` is the simple dropdown to select the currently displayed pipeline.
 
+## Utility components
+
+Weaverbird also exports some of its internal components so they can be reused elsewhere:
+
+- `FilterEditor` is the main component of the Filter step. It can display and edit trees of filtering conditions.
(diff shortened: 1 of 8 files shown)


docs/_docs/tech/ui-components.md
```diff
@@ -0,0 +1,46 @@
...
+> An example of such integration is provided in `playground/dist/app.js`.
+
+## Main sub-components
+
+The `Vqb` component consists is three sub-components, that you can use as well to fit it nicely into your app.
+
+- `QueryBuilder` is the main left panel, where user see the current pipeline and can edit its steps.
+- `DataViewer` is the main right panel, where the data table resides and the associated actions.
+- `PipelineSelector` is the simple dropdown to select the currently displayed pipeline.
+
```
rhuille, line 46: It misses a word about the `FilterEditor` (https://github.com/ToucanToco/weaverbird/blob/master/src/main.ts#L20)
```
`FilterEditor` is needed for internal purpose in ToucanToco. It must not be used has its api is uncertain.
```

rhuille: Also, this message could be added in a docstring in the `main.ts` abobe https://github.com/ToucanToco/weaverbird/blob/master/src/main.ts#L20 ?

davinov: I know :) I already prepared this, but as this PR may be merged before #494, I think it's better to add it in the PR that adds the component:

```
## Utility components

Weaverbird also exports some of its internal components so they can be reused elsewhere:

- `FilterEditor` is the main component of the Filter step. It can display and edit trees of filtering conditions.
```

davinov: Adding a doc string to separate clearly the components in main.ts is a great idea, doing thatright now

davinov: Reverting what I said, you're right it's already exported :joy: 
So I'm adding the doc abut FilerEditor in this PR

Thanks for noticing!

rhuille: I really think we should specify that these "utility components" (only the `FilterEditor` for now) `are for Toucan Toco internal purpose`. Maybe, we will decide in the futur that they are part of the weaverbird api, but maybe we will find an other way. It is too soon to be sure !

davinov: Not convinced there, it's clear the API as we've not reach the 1.0 version.
I can add a message though indicating that these components are not directly part of the external API and changes can happen on them.

PR #497, opened 2020-04-06
New component: ListUniqueValue

rhuille:
This component display a column's unique values. It allows to search among values and to select and unselect values.
Notes: This is a visual component, it does not trigger any action nor mutationhimself.
It is meant to be used inside "ActionMenu".

This commit introduce `ListUniqueValue` component, tests and a storybook. For now, the component is not used inside the project.

image

In the close futur: (spoiler alert)

image

### DIFF
--- a/src/components/ListUniqueValues.vue
+++ b/src/components/ListUniqueValues.vue
@@ -1,0 +1,208 @@
+<template>
+  <div class="list-unique-values">
+    <div class="list-unique-values__title">Filter values</div>
+    <input class="list-unique-values__search-box" placeholder="Search 50 items" v-model="search" />
+    <div class="list-unique-values__select-clear-all">
+      <span class="list-unique-values__select-all" @click="selectAll">Select all</span> &nbsp;
+      <span class="list-unique-values__clear-all" @click="clearAll">Clear all</span>
+    </div>
+    <div
+      class="list-unique-values__checkbox-container"
+      v-for="option in searchedOptions"
+      :key="option.value"
+    >
+      <CheckboxWidget
+        :label="`${option.value} (${option.count})`"
+        :value="isChecked(option)"
+        @input="toggleCheck(option)"
+      />
+    </div>
+    <div class="list-unique-values__load-all-values" v-if="loaded">
+      <div>List maybe incomplete</div>
+      <div @click="loadAllValues" class="list-unique-values__load-all-values-button">
+        Load all values
+      </div>
+    </div>
+  </div>
+</template>
+<script lang="ts">
+import _difference from 'lodash/difference';
+import _union from 'lodash/union';
+import { Component, Prop, Vue } from 'vue-property-decorator';
+
+import { FilterConditionInclusion } from '@/lib/steps.ts';
+
+// This type will be imported from @/lib/dataset/helpers.ts
+type ColumnValueStat = {
+  value: any;
+  count: number;
+};
+
+/**
+  Edit a `filter` object of the form:
+  filter = {
+    column: 'toto', // this key should not be updated by this component
+    operator: 'in' // or `nin` which means 'not in'
+    value: ['France', 'UK', 'Spain']
+  }
+ */
+import CheckboxWidget from './stepforms/widgets/Checkbox.vue';
+@Component({
+  name: 'list-unique-values',
+  components: { CheckboxWidget },
+})
+export default class ListUniqueValues extends Vue {
+  @Prop({
+    type: Array,
+    required: true,
+  })
+  options!: ColumnValueStat[];
+
+  @Prop({
+    type: Object,
+    required: true,
+  })
+  filter!: FilterConditionInclusion;
+
+  @Prop({
+    type: Boolean,
+    required: true,
+  })
+  loaded!: boolean;
+
+  search = '';
+
+  isChecked(option: ColumnValueStat): boolean {
+    if (this.filter.operator == 'in') {
+      return this.filter.value.includes(option.value);
+    } else {
+      return !this.filter.value.includes(option.value);
+    }
+  }
+
+  selectAll() {
+    if (this.search === '') {
+      this.$emit('input', {
+        column: this.filter.column,
+        operator: 'nin',
+        value: [],
+      } as FilterConditionInclusion);
+    } else {
+      this.$emit('input', {
+        column: this.filter.column,
+        operator: this.filter.operator,
+        value: (this.filter.operator === 'in' ? _union : _difference)(
+          this.filter.value,
+          this.searchedOptions.map(option => option.value),
+        ),
+      } as FilterConditionInclusion);
+    }
+  }
+
+  clearAll() {
+    if (this.search === '') {
+      this.$emit('input', {
+        column: this.filter.column,
+        operator: 'in',
+        value: [],
+      } as FilterConditionInclusion);
+    } else {
+      this.$emit('input', {
+        column: this.filter.column,
+        operator: this.filter.operator,
+        value: (this.filter.operator === 'in' ? _difference : _union)(
+          this.filter.value,
+          this.searchedOptions.map(option => option.value),
+        ),
+      } as FilterConditionInclusion);
+    }
+  }
+
+  toggleCheck(option: ColumnValueStat) {
+    const newFilter = { ...this.filter };
+    if (newFilter.value.includes(option.value)) {
+      newFilter.value.splice(newFilter.value.indexOf(option.value), 1);
+    } else {
+      newFilter.value.push(option.value);
+    }
+    this.$emit('input', newFilter);
+  }
+
+  get searchedOptions(): ColumnValueStat[] {
+    return this.options.filter(option => this.searchFunction(option.value, this.search));
+  }
+
+  // TODO: enhance `searchFunction`
+  searchFunction(value: string, search: string) {
+    return value.toLowerCase().startsWith(search.toLowerCase());
+  }
+
+  loadAllValues() {
+    throw new Error('Not implemented');
+  }
+}
+</script>
+<style lang="scss" scoped>
+@import '../styles/_variables.scss';
+
+.list-unique-values__title {
+  text-transform: uppercase;
+}
+
+.list-unique-values {
+  max-height: 300px;
+  overflow: auto;
+  background-color: $light-grey;
+  padding: 10px 12px;
+  font-size: 13px;
+  white-space: nowrap;
+}
+
+.list-unique-values__search-box {
+  box-sizing: border-box;
+  padding: 5px;
+  margin: 10px 0px;
+  width: 100%;
+  outline: none;
+}
+
+.list-unique-values__search-box::placeholder {
+  font-style: italic;
+  color: #cccccc;
+}
+
+.list-unique-values__select-clear-all {
+  margin: 0px 0px 15px;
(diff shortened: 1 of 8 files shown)


src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+  <div class="list-unique-values">
+    <div>FILTER VALUES</div>
```
davinov, line 3: I think the uppercase belongs to the styling part (css)

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,153 @@
...
+    </div>
+  </div>
+</template>
+<script lang="ts">
+import { Component, Prop, Vue } from 'vue-property-decorator';
+
+// This type will be imported from @/lib/dataset/helpers.ts
+type ColumnValueStat = {
+  value: any;
+  nbOcc: number;
```
davinov, line 36: `occurrences` or `occurrencesCount` or just `count`?
Better to stay explict

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,153 @@
...
+};
+
+import CheckboxWidget from './stepforms/widgets/Checkbox.vue';
+@Component({
+  name: 'filter-value',
+  components: { CheckboxWidget },
+})
+export default class ActionMenu extends Vue {
```
davinov, line 44: I'm confused about the name of this component:
- file name: ListUniqueValues
- name property: filter-value
- class name: ActionMenu (probably a forgotten copy-paste there :p)

If the mission of the component is to display and filter unique values of co column, its name should reflect that :)

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,153 @@
...
+  toggleCheck(value: string) {
+    this.isChecked[value] = !this.isChecked[value];
+    this.$emit(
+      'input',
+      Object.keys(this.isChecked).filter(k => {
+        return this.isChecked[k];
+      }),
+    );
+  }
+
```
davinov, line 80: Reading this means to me we have two sources of truth for the list of selected values.
One come from above this component and needs to be updates with the `$emit`
One come from inside the component

Keeping both in sync is tricky, and will eventually leads to error.
Could we use only the external one? In that case, the internal state become only a computed.

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,153 @@
...
+    });
+  }
+
+  // TODO: enhance searchFunction
+  searchFunction(value: string, search: string) {
+    return value.startsWith(search);
+  }
+
+  loadMore() {
+    // trigger store action to load more
```
davinov, line 93: Good idea to prototype lie this :)

You can do it python-like to ensure we log something when this gets merge:
`throw new Error('Not implemented');`

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+</script>
+<style lang="scss" scoped>
+.list-unique-values {
+  max-height: 300px;
+  overflow: auto;
+  background-color: #fafafa; //$grey
```
davinov, line 101: I feel we're gonna have a lot of different grays if we don't start using variables here ^^
May we add the color to `src/styles/_variables.scss` for further reuse?

stories/list-unique-values.js
```diff
@@ -0,0 +1,20 @@
+import { storiesOf } from '@storybook/vue';
+
+import { ListUniqueValues } from '../dist/storybook/components';
+const stories = storiesOf('ListUniqueValues', module);
+
+console.log('ListUniqueValues', ListUniqueValues)
```
davinov, line 6: Forgotten log?

stories/list-unique-values.js
```diff
@@ -0,0 +1,20 @@
...
+        values: [
+            {value: 'France', nbOcc: 12},
+            {value: 'UK', nbOcc: 1},
+            {value: 'China', nbOcc: 14},
+            {value: 'Italy', nbOcc: 23},
+        ]
+    }
+  },
+  template: `<div style="font-family: 'Montserrat', sans-serif;"><ListUniqueValues :values="values"/></div>`,
+}));
```
davinov, line 20: Thanks for enriching the storybook <3

tests/unit/list-unique-values.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+localVue.use(Vuex);
+
+describe('List Unique Value', () => {
+  const DUMMY_UNIQUE_VALUES = [
+    { value: 'A', nbOcc: 10 },
+    { value: 'B', nbOcc: 4 },
+    { value: 'C', nbOcc: 2 },
+  ];
+
+  it('should instantiate with the correct unique values', () => {
```
davinov, line 16:
```suggestion
  it('should display the list of unique values and how much time they are present in the whole dataset', () => {
```

tests/unit/list-unique-values.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    expect(wrapper.exists()).toBeTruthy();
+    const CheckboxWidgetArray = wrapper.findAll('CheckboxWidget-stub');
+    expect(CheckboxWidgetArray.length).toEqual(3);
+    expect(CheckboxWidgetArray.at(0).vm.$props.label).toEqual('A (10)');
+    expect(CheckboxWidgetArray.at(1).vm.$props.label).toEqual('B (4)');
+    expect(CheckboxWidgetArray.at(2).vm.$props.label).toEqual('C (2)');
+  });
+
+  it('should instantiate with all checkbox checked', () => {
```
davinov, line 29: IMO, this is the responsibility of the component that hosts the state "which unique values are selected".
Regarding my previous comment, I don't think it should be this one

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,153 @@
...
+  loaded!: boolean;
+
+  searchedValues: ColumnValueStat[] = this.values;
+
+  isChecked: Record<string, boolean> = Object.fromEntries(this.values.map(e => [e.value, true]));
+
+  selectAll() {
+    this.isChecked = Object.fromEntries(this.values.map(e => [e.value, true]));
+    this.$emit('input', Object.keys(this.isChecked));
+  }
```
davinov, line 64: In Toucan Toco, we have a clever way to say "all values" or "all values except", which avoids sending all the actual values. Indeed, this could be huge if the column has only unique values.
Also, it's not the same to say: "I want every values except this one" and "I want this this and this values".

The Toucan model for this:
```ts
{
  type: 'include' | 'exclude',
  values: any[]
}
```

src/components/ListUniqueValues.vue
rhuille: Oups sorry

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,176 @@
...
+    type: Array,
+    required: true,
+  })
+  options!: ColumnValueStat[];
+
+  @Prop({
+    type: Array,
+    required: true,
+  })
+  value!: boolean[];
```
davinov, line 53: What do you think about a docstring explaing what we intend by `value` here?
If I understand correctly, it's an array, indexed the same way as options, that say if a value is checked or not?

davinov: I still think it would be better to have a data structure saying if values are included or excluded.
By default, no value should be excluded.

I think this will prevent cases where the data table is updated with new values, which would be masked if the values are only included in a positive manner.
I'll explain:

Initial dataset:
displayedData: [1, 2, 3, 4, 5]

I decide to hide the value 2:
value: [true, false, true, true, true]
displayedData: [1, 3, 4, 5]

I change something in a step:
newData: [2, 3, 4, 5, 6]
What will be displayed ? 
using position of each value, it would be: [2, 4, 5, 6]
using positively included values, it would be: [3, 4, 5] because 6 didn't exist when we added all the values to the include filter.
Using such a data filter would force to reinitialize it at each request, to keep it in sync. It may be forgotten at some point by a dev, raising such issues.

That's why I advise using a filter like:
type: 'exclude', values: [2]
Even if the dataset changes, the intent of the user is correctly applied.
It's also more compact to store in the state :p specially when there is A LOT of unique values. An frequent use case for this would be to exclude 0 values from a numerical column. 

What are your opinion on this?

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,222 @@
...
+      } else {
+        this.$emit('input', {
+          column: this.filter.column,
+          operator: 'nin',
+          value: _difference(
+            this.filter.value,
+            this.searchedOptions.map(option => option.value),
+          ),
+        } as FilterConditionInclusion);
+      }
```
davinov, line 101: I think if would be clearer to declare a variable of type FilterConditionInclusion and emit it afterwards:

```suggestion
   const newFilter = {
        column: this.filter.column,
        operator: 'nin',
        value: [],
      };
    if (this.search !== '') {
      // Only select all _searched_ values
      if (this.filter.operator === 'in') {
        newFilter.operator = 'in';
        newFilter.value = _union(
            this.filter.value,
            this.searchedOptions.map(option => option.value),
          );
      } else {
        newFilter.value = _difference(
            this.filter.value,
            this.searchedOptions.map(option => option.value),
          )
       }
       this.$emit('input', newFilter);
      }
```

Same for clearAll I guess

src/components/ListUniqueValues.vue
```diff
@@ -0,0 +1,222 @@
...
+    type: Array,
+    required: true,
+  })
+  options!: ColumnValueStat[];
+
+  @Prop({
+    type: Object,
+    required: true,
+  })
+  filter!: FilterConditionInclusion;
```
davinov, line 57: I think there is no need for this component to know which column it is editing.
You can declare the sub type like this in steps.ts:
```
export type ConditionInclusion = {
  value: any[];
  operator: 'in' | 'nin';
};

FilterConditionInclusion = ConditionInclusion & {
  column: string;
};

File: stories/list-unique-values.js [hunk 15]
```diff
@@ -0,0 +1,22 @@
...
+            {value: 'UK', count: 1},
+            {value: 'China', count: 14},
+            {value: 'Framboise', count: 10},
+            {value: 'Italy', count: 23},
+        ],
+        filter: { column: "col1", operator: "in", value: []},
+        loaded: false,
+    }
+  },
+  template: `<div style="font-family: 'Montserrat', sans-serif;"><ListUniqueValues :loaded="loaded" :filter="filter" @input="filter = $event" :options="options"/></div>`,
```
davinov, line 21: It's weird to add the `font-family` there, isn't it?
Either we decide that this component should use the font of it's parent component (nothing to do, my preference go there) or we enforce Montserrat but it needs to be indicated in the component style.
If what we want is to change the default storybook font to Montserrat, let's do it in .storybook/preview-head.html

src/components/ListUniqueValues.vue
davinov: Summary of Slack discussion:
@rhuille prefer avoiding type multiplication, and keep a string link between this and the filter step, and the ability in the short future to turn this object into a real filter step condition
I still advise against keeping the `column` attribute, but it's not an important point so let's go with this :)

PR #498, opened 2020-04-07
Split `_updateDataset` and introduce `backendify`

rhuille:
- extract the pipeline preparation from `_updateDataset`, e.g. the
  dereferencing,

- introduce the `backendify` "decorator" (but can't be used as
  @backendify around functions unfortunately because JavaScript doesn't
  support it) that:

  - sets the `loading: true` property on the store at the beginning,
  - logs the error in the store if any,
  - sets the `loading: false` property on the store at the end.

### DIFF
--- a/src/store/backend-plugin.ts
+++ b/src/store/backend-plugin.ts
@@ -87,40 +87,67 @@ export function dereferencePipelines(
   return dereferencedPipeline;
 }
 
-async function _updateDataset(store: Store<any>, service: BackendService, pipeline: Pipeline) {
-  if (!store.getters[VQBnamespace('pipeline')].length) {
-    return;
+export let backendService: BackendService; // set at plugin instantiation
+
+/**
+ * `_preparePipeline` responsibility is to prepare the pipeline so as to be ready for direct translation.
+ * Specifically, this consists in 2 things:
+ *   - dereferencePipelines
+ *   - interpolate if an `interpolateFunc` has been set
+ */
+function _preparePipeline(pipeline: Pipeline, store: Store<any>) {
+  const { interpolateFunc, variables, pipelines } = store.state[VQB_MODULE_NAME];
+  if (pipelines && Object.keys(pipelines).length) {
+    pipeline = dereferencePipelines(pipeline, pipelines);
   }
-  try {
-    store.commit(VQBnamespace('setLoading'), { isLoading: true });
-    const { interpolateFunc, variables, pipelines } = store.state[VQB_MODULE_NAME];
-    if (pipelines && Object.keys(pipelines).length) {
-      pipeline = dereferencePipelines(pipeline, pipelines);
-    }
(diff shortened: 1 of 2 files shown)


playground/app.js
```diff
@@ -109,6 +109,7 @@ class MongoService {
   async executePipeline(_store, pipeline, limit, offset = 0) {
     const { domain, pipeline: subpipeline } = filterOutDomain(pipeline);
     const query = mongo40translator.translate(subpipeline);
+    console.log('in executePipeline', this)
```
adimascio, line 112: This should be removed :)

src/store/backend-plugin.ts
```diff
@@ -87,40 +87,69 @@ export function dereferencePipelines(
   return dereferencedPipeline;
 }
 
-async function _updateDataset(store: Store<any>, service: BackendService, pipeline: Pipeline) {
-  if (!store.getters[VQBnamespace('pipeline')].length) {
-    return;
+export let backendService: BackendService; // set at plugin instantiation
+
+/**
+  `_preparePipeline` do 2 stuffs:
```
adimascio, line 93:
```suggestion
  `_preparePipeline` does 2 stuffs:
```

adimascio: I would maybe phrase it like: 

```
`_preparePipeline` responsibility is to prepare the pipeline so as to be ready for direct translation. Specifically,
this consists in 2 things:

- ...

```

playground/app.js
rhuille: Oupsy sorry for this

src/store/backend-plugin.ts
rhuille: Thanks for the rephrase !

src/store/backend-plugin.ts
```diff
@@ -87,40 +87,67 @@ export function dereferencePipelines(
...
+  return pipeline;
+}
+
+/**
+ * `backendify` is a wrapper around backend service functions that:
+ *   - sets the `loading: true` property on the store at the beginning,
+ *   - logs the error in the store if any,
+ *   - sets the `loading: false` property on the store at the end.
+ */
+export function backendify(target: Function) {
```
davinov, line 117: executeBackendRequest?
withBackendRequest? (context-manager like)

src/store/backend-plugin.ts
```diff
@@ -87,40 +87,67 @@ export function dereferencePipelines(
...
+      const response = { error: { type: 'error', message: error.toString() } };
       store.commit(VQBnamespace('logBackendError'), {
-        backendError: { type: 'error', message: response.error },
+        backendError: response.error,
       });
-    } else {
-      store.commit(VQBnamespace('setDataset'), { dataset: response.data });
+      return response;
+    } finally {
+      store.commit(VQBnamespace('setLoading'), { isLoading: false });
```
davinov, line 135: callbacks avant requête / après succès / après erreur ?

PR #499, opened 2020-04-07
Add a `isUniqueValuesLoading` key to columns in the vuex state

zegonz:
`state.dataset.headers` contains elements like:
Before: `{name: 'col1', type: 'number'}`
Now: `{name: 'col1', type: 'number', isUniqueValuesLoading: false}`
Also, add the corresponding getter and mutation

Please read commit by commit.

ezgif com-optimize

NB: 
This PR also fix story of DataViewer and remove the broken story of `RenameStep`.  An issue has been written to create stories for the form steps: https://github.com/ToucanToco/weaverbird/issues/501


zegonz:
[(5) ETQDev le state "loading" n'est plus un boolean mais un "Enum(`None`, `DataSet`, `UniqueValues`)"](https://trello.com/c/sEHonI6R/348-5-etqdev-le-state-loading-nest-plus-un-boolean-mais-un-enumnone-dataset-uniquevalues)

davinov:
I'm not that sure it's such a good idea to introduce an enumeration here.
Why not simply add a "areUniqueValuesLoading" boolean?

rhuille:
The idea is to limit the number of state properties and to keep inside the name `loading` the loading state.

src/store/mutations.ts
```diff
@@ -284,6 +284,18 @@ class Mutations {
...
+   * Set unique values loading
+   */
+  setUniqueValuesLoading(
+    state: VQBState,
+    { isLoading, column }: { isLoading: boolean; column: string },
+  ) {
+    state.dataset.headers[
+      state.dataset.headers.map(col => col.name).indexOf(column)
+    ].isUniqueValuesLoading = isLoading;
+  }
```
adimascio, line 297:
```suggestion
    state.dataset.headers.find(hdr => hdr.name === column).isUniqueValuesLoading = isLoading;
```

src/lib/dataset/index.ts
```diff
@@ -12,6 +12,7 @@ export type ColumnTypeMapping = {
 export type DataSetColumn = {
   name: string;
   type?: DataSetColumnType;
+  isUniqueValuesLoading?: boolean; // It will be updated by backendify function
```
adimascio, line 15: Could you document the property more precisely. Maybe something like:

```typescript
    /**
      * indicates if there's an ongoing operation to fetch all unique values for this column from the database.
      * This value is updated by the `backendify` function.
      */
      isUniquesValueLoading?: boolean;
```

src/store/getters.ts
```diff
@@ -82,6 +82,12 @@ const getters: GetterTree<VQBState, any> = {
    */
   pageno: (state: VQBState) =>
     state.dataset.paginationContext ? state.dataset.paginationContext.pageno : 1,
+  /**
+   * helper that is True if unique values are loading
+   */
+  isUniqueValuesLoading: (state: VQBState) => (column: string) =>
+    state.dataset.headers[state.dataset.headers.map(col => col.name).indexOf(column)]
```
adimascio, line 89:
```suggestion
    state.dataset.headers.find(hdr => hdr.name === column).isUniqueValuesLoading
```

zegonz:
@adimasci I push changes you ask for, can you have a look please ?

src/store/backend-plugin.ts
```diff
@@ -114,32 +114,54 @@ function _preparePipeline(pipeline: Pipeline, store: Store<any>) {
  *   - logs the error in the store if any,
  *   - sets the `loading: false` property on the store at the end.
  */
-export function backendify(target: Function) {
+export function backendify(
+  target: Function,
+  callFront?: Function,
+  callBack?: Function,
+  reject?: Function,
+) {
```
adimascio, line 122: Two comments here:

- I would name those parameters `before`, `after`, `onError` or `before`, `onSuccess`, `onError`

But then… what does `backendify` become useful for? I mean if you specify those 3 parameters, it becomes more or less equivalent to:

```typescript
before()
try {
   target();
} cacth(err) {
   onError(err);
}
after();
```

but just (very) less clear. Am I missing something?

tests/unit/plugins.spec.ts
```diff
@@ -366,3 +435,176 @@ describe('backend service plugin tests', () => {
...
+      { name: 'domain', domain: 'GoT' },
+      { name: 'replace', search_column: 'characters', to_replace: [['Snow', 'Targaryen']] },
+      { name: 'sort', columns: [{ column: 'death', order: 'asc' }] },
+    ];
+    const store = setupMockStore(buildStateWithOnePipeline(pipeline), [
+      servicePluginFactory(new DummyBuggedService()),
+    ]);
+    const wrapper = mount(PipelineComponent, { store, localVue });
+    wrapper.find('.query-pipeline-queue__dot').trigger('click');
+    await flushPromises();
```
adimascio, line 451: Genuine question: is there a reason why `$nextTick()` would not work here?

src/store/backend-plugin.ts
rhuille: Hi tanks Adrien! This can be decided later... Thus I removed the modification on `backendify` from this PR

src/store/getters.ts
```diff
@@ -82,6 +84,11 @@ const getters: GetterTree<VQBState, any> = {
    */
   pageno: (state: VQBState) =>
     state.dataset.paginationContext ? state.dataset.paginationContext.pageno : 1,
+  /**
+   * helper that is True if unique values are loading
+   */
+  isUniqueValuesLoading: (state: VQBState) => (column: string) =>
+    (state.dataset.headers.find(hdr => hdr.name === column) as DataSetColumn).isUniqueValuesLoading,
```
adimascio, line 91: I'm surprised that the explicit casting is necessary :thinking:

tests/unit/list-unique-values.spec.ts
```diff
@@ -1,31 +1,62 @@
...
+  @param isUniqueValuesLoading the store parameter indicating if column unique value is loading
+
+  By default the wrapper is mounted with 4 options: "France", "Framboise", "Spain" and "UK"
+  - The checked values are "France" and "Spain"
+  - All uniques are loaded
+  */
+  const shallowMountWrapper = (
+    filterValue: string[] = ['France', 'Spain'],
+    operator: 'in' | 'nin' = 'in',
+    // eslint-disable-next-line @typescript-eslint/no-inferrable-types
```
adimascio, line 29: :thinking: ?

src/store/getters.ts
rhuille: The column names are not explicitly listed for typescript. Consequently, the `.find()` method may return an `undefined` value...

PR #500, opened 2020-04-10
FilterEditor: a condition can be deleted if groups are present

davinov:
Before:
Screenshot from 2020-04-10 15-06-13
Note that it's not possible to remove the first row, even if it's not alone.

After:
Screenshot from 2020-04-10 15-04-43


davinov:
I also added in this PR a fix concerning the WidgetJSON emitting too many events
It fixes the empty row created when adding a condition in a row

PR #506, opened 2020-04-29
addLocalUnique to dataset handles the pagination

rhuille:


src/lib/dataset/helpers.ts
```diff
@@ -80,11 +80,15 @@ function _statCompare(stat1: ColumnValueStat, stat2: ColumnValueStat) {
  */
 export function addLocalUniquesToDataset(dataset: DataSet) {
   const uniqueStats = localUniqueStats(dataset);
+  const totalCount: number = dataset.paginationContext
+    ? dataset.paginationContext.totalCount || 0
+    : 0;
+  const pagesize: number = dataset.paginationContext ? dataset.paginationContext.pagesize || 0 : 0;
```
davinov, line 86: Maybe we can use optional chaining for this case? https://toucantoco.atlassian.net/wiki/spaces/TTA/pages/1375666275/OKR+-+Product+Team+Q2+2020
```suggestion
  const pagesize: number = dataset.paginationContext?.pagesize || 0;
```

(If yes, same below)

PR #510, opened 2020-04-30
Add IF...THEN...ELSE step

vdestraitt:
image


src/lib/translators/mongo.ts
```diff
@@ -64,6 +81,155 @@ export function _simplifyAndCondition(filterAndCond: FilterComboAndMongo): Mongo
...
+): MongoStep {
+  const operatorMapping = {
+    eq: '$eq',
+    ne: '$ne',
+    lt: '$lt',
+    le: '$lte',
+    gt: '$gt',
+    ge: '$gte',
+    in: '$in',
+    nin: '$nin',
```
davinov, line 210: I don't find `$nin` among expression operators: https://docs.mongodb.com/manual/meta/aggregation-quick-reference/#index-of-expression-operators
I think we have no choice than to do a `$not: {$in: ...}}`
This breaks a bit the string map approach, we may have to use small functions instead

src/lib/translators/mongo.ts
```diff
@@ -64,6 +81,155 @@ export function _simplifyAndCondition(filterAndCond: FilterComboAndMongo): Mongo
...
+      if (pseudo[node.name]) {
+        node.name = pseudo[node.name].replace(ESCAPE_OPEN, '').replace(ESCAPE_CLOSE, '');
+      }
+    }
+    return node;
+  });
+  return mathjsTree;
+}
+
+function buildConditionTree(
```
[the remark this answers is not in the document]
davinov: To explain my point further: below we have `buildMatchTree(condition)` which looks awefully similar :p

src/lib/translators/mongo.ts
```diff
@@ -336,6 +524,27 @@ function transformEvolution(step: Readonly<S.EvolutionStep>): MongoStep {
...
+): MongoStep {
+  const ifBlock: MongoStep = buildConditionTree(step.if);
+  const thenBlock = buildMongoFormulaTree(buildFormulaTree(step.then));
+  let elseBlock: MongoStep | string | number;
+  if (typeof step.else === 'string') {
+    elseBlock = buildMongoFormulaTree(buildFormulaTree(step.else));
+  } else {
+    elseBlock = transformIfThenElseStep(step.else);
+  }
+  return { $cond: [ifBlock, thenBlock, elseBlock] };
```
davinov, line 545: :+1: 
Maybe it's easier to use the `{ $cond: { if: <boolean-expression>, then: <true-case>, else: <false-case> } }` syntax rather than the one that counts on the position inside the array. (To read more easily resulting queries)

src/lib/translators/mongo.ts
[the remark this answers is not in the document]
davinov: Reading again mongo's doc, maybe we should name that "buildExpression", as that's how this type of field is named in Mongo
https://docs.mongodb.com/manual/meta/aggregation-quick-reference/#aggregation-expressions

src/lib/translators/mongo.ts
```diff
@@ -20,10 +20,27 @@ export interface MongoStep {
...
+  month: '$month',
+  day: '$dayOfMonth',
+  hour: '$hour',
+  minutes: '$minute',
+  seconds: '$second',
+  milliseconds: '$millisecond',
+  dayOfYear: '$dayOfYear',
+  dayOfWeek: '$dayOfWeek',
+  week: '$week',
+};
```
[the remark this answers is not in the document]
vdestraitt: sorry about that... :(

src/lib/translators/mongo.ts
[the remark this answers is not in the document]
vdestraitt: You're right, and I'm going for............ `buildCondExpression` :)

src/lib/steps.ts
```diff
@@ -168,6 +168,14 @@ export type FromDateStep = {
   format: string;
 };
 
+export type IfThenElseStep = {
+  name: 'ifthenelse';
+  newColumn: string;
+  if: FilterSimpleCondition | FilterComboAnd | FilterComboOr;
```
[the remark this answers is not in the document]
vdestraitt: True, thanks

src/lib/translators/mongo.ts
```diff
@@ -64,6 +81,155 @@ export function _simplifyAndCondition(filterAndCond: FilterComboAndMongo): Mongo
...
+    notnull: '$ne',
+  };
+  if (S.isFilterComboAnd(cond)) {
+    if (cond.and.length == 1) {
+      return buildConditionTree(cond.and[0]);
+    } else {
+      // if cond.and.length > 1 we need to bind conditions in a $and operator,
+      // as we need a unnique document to be used as the first argument of the
+      // $cond operator
+      return { $and: cond.and.map(elem => buildConditionTree(elem)) };
```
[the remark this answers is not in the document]
vdestraitt: Indeed, thank you :)

src/lib/translators/mongo.ts
vdestraitt: 👍

src/lib/translators/mongo.ts
vdestraitt: Oooh thank you for your vigilance!

vdestraitt: @davinov actually it does exist https://docs.mongodb.com/v4.0/reference/operator/query/nin/index.html
Not sure why it's not referenced on the page you linked... I've tried it with a UAT it's all good :)

src/lib/steps.ts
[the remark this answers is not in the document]
vdestraitt: Specific commit [here](https://github.com/ToucanToco/weaverbird/pull/510/commits/acacf787034e978ad0416b3bda76b38e72b2f7f6)

src/lib/translators/mongo.ts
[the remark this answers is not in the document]
vdestraitt: Fixed [here](https://github.com/ToucanToco/weaverbird/pull/510/files#diff-b4949f396e00c36a178d54c27a25dcf3R221)

src/lib/translators/mongo.ts
vdestraitt: Fixed [here](https://github.com/ToucanToco/weaverbird/pull/510/files#diff-b4949f396e00c36a178d54c27a25dcf3R545)

src/components/stepforms/widgets/IfThenElseWidget.vue
```diff
@@ -0,0 +1,212 @@
...
+    default: () => [],
+  })
+  errors!: ErrorObject[];
+
+  @VQBModule.Getter columnTypes!: ColumnTypeMapping;
+
+  readonly title: string = 'Add a conditional column';
+  readonly elseModes = ['ELSE:', 'ELSE IF:'];
+  elseMode = typeof this.value.else === 'string' ? 'ELSE:' : 'ELSE IF:';
+  editedValue = { ...this.value };
```
rhuille, line 97: This component `emit` "immediately" the new value (it does not wait for a user action like a "save"), you should not need `editedValue`. Simply pass down `this.value` to child component, and emit an updated value using the `update{stuff}` methods. (In other word: this component does not need an internal state.)

Note: `editedValue` is useful when your component `emit` at a "save" action, like in step, because you cannot mutate the props, you need an internal state.

src/components/stepforms/IfThenElseStepForm.vue
```diff
@@ -0,0 +1,113 @@
...
+      else: '',
+    }),
+  })
+  initialStepValue!: IfThenElseStep;
+
+  @VQBModule.Getter columnTypes!: ColumnTypeMapping;
+
+  readonly title: string = 'Add a conditional column';
+
+  ifthenelse = _.omit({ ...this.initialStepValue }, ['name', 'newColumn']);
```
rhuille, line 57: To avoid importing all lodash object `_`, please import only `omit` function from lodash with `import _omit from 'lodash/omit'`

rhuille: Even better: avoid using lodash at all, with destructuring:
`const {name , newColumn , ...ifthenelse} = this.initialStepValue`

vdestraitt: I couldn't use the destructuring syntax inside the component. So I'm implementing your first suggestion

PR #514, opened 2020-05-03
Factorize buttons-menu-like component

rhuille:
This concerne: ActionMenu, DataTypeMenu and ActionToolbarButtons
The visual part of these three components is now factorise in one new component named "Menu"

Also update design of Apply Filter button:

drawing

### DIFF
--- a/src/components/Menu.vue
+++ b/src/components/Menu.vue
@@ -1,0 +1,178 @@
+<template>
+  <popover :visible="visible" :align="alignLeft" bottom @closed="$emit('closed')">
+    <div class="menu__body">
+      <transition name="slide-left" mode="out-in">
+        <div v-if="visiblePanel == 1">
+          <div class="menu__panel">
+            <div
+              v-for="button in buttons"
+              :key="button.label"
+              class="menu__option"
+              @click="button.onclick"
+              v-html="button.html"
+            />
+            <div
+              v-if="otherButtons.length > 0"
+              class="menu__option menu__option--top-bordered"
+              @click="visiblePanel = 2"
+            >
+              Other Operations
+            </div>
+            <div class="menu__option--top-bordered">
+              <slot name="first-panel-special" />
+            </div>
+          </div>
+        </div>
+      </transition>
+      <transition name="slide-right" mode="out-in">
+        <div v-if="visiblePanel == 2">
+          <div class="menu__panel">
+            <div class="menu__option--back" @click="visiblePanel = 1">
+              <i class="fas fa-angle-left" /> BACK
+            </div>
+            <div
+              class="menu__option"
+              v-for="button in otherButtons"
+              :key="button.label"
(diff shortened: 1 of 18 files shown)


src/components/Menu.vue
```diff
@@ -0,0 +1,178 @@
...
+    <div class="menu__body">
+      <transition name="slide-left" mode="out-in">
+        <div v-if="visiblePanel == 1">
+          <div class="menu__panel">
+            <div
+              v-for="button in buttons"
+              :key="button.label"
+              class="menu__option"
+              @click="button.onclick"
+              v-html="button.html"
```
davinov, line 12: I think it's best to avoid html if we can.
Here where labels seems enough :)

src/components/Menu.vue
```diff
@@ -0,0 +1,178 @@
...
+            />
+            <div
+              v-if="otherButtons.length > 0"
+              class="menu__option menu__option--top-bordered"
+              @click="visiblePanel = 2"
+            >
+              Other Operations
```
davinov, line 19:
```suggestion
              Other operations
```
Sorry for being a typo-freak :p

src/components/Menu.vue
davinov: Actually I realized you probably did that for data types icons.
I think it would even be better to have a slot here and a MenuButton component, so that the usage looks lioke:
```
<Menu>
  <MenuButton @click="...">Some operation</MenuButton>
  <MenuButton @click="...">Some operation</MenuButton>
  ...
</Menu>
```
It seems to me that this kind of declarative API, based on templates, is easier to read, understand and extend, that the array-based one.

src/components/Menu.vue
```diff
@@ -0,0 +1,178 @@
...
+            </div>
+            <div class="menu__option--top-bordered">
+              <slot name="first-panel-special" />
```
davinov, line 22: The fact that we need this makes me wonder if this abstraction is at the right level.
It seems that you had to add a whole bunch of stuff only for the ColumnMenu case.
 
What about just having a Menu component with a slot to let the inside customizable?
This way, DataTypeMenu and ActionMenu stays simple, and have another component, MultiLevelMenu, which can extend the simple Menu and handle the panel sliding.

src/components/Menu.vue
davinov: I took a minute to check on Vuetify's approach, and they also use this kind of template API. See https://vuetifyjs.com/en/components/lists/

davinov:
@rhuille I've started working on the component-based API, and took the opportunity to add stories for some undocumented components.
See 174e428  for the menu components and a920d34 008be15 for usage.
Still need to be done: tests update and multi-panel.

davinov:
After discussion with @rhuille, we don't agree on this componentization proposal. Let's get more opinions to know which approach to take in this project :)

PR #522, opened 2020-05-08
F/statistics step

rhuille:
New step "statistics" !

You can use this step to compute main statistics, like median or quintiles, of a numeric column.

With:
Screenshot 2020-05-18 at 10 57 41

You get:

Screenshot 2020-05-18 at 11 00 10

url to test: https://weaverbird-test.herokuapp.com/


src/lib/translators/mongo.ts
```diff
@@ -527,6 +527,91 @@ function transformSplit(step: Readonly<S.SplitStep>): MongoStep {
...
+    step.statistics.includes('variance') ||
+    step.statistics.includes('std');
+  // ---
+
+  const stepQuantileIndex: any = { $addFields: {} };
+  let statistic: string;
+  for (statistic of step.statistics) {
+    if (isQuintileStatistics(statistic)) {
+      stepQuantileIndex.$addFields[`${statistic}Index`] = {
+        $trunc: quintilesIndexFormula[`${statistic}Index`], //TODO: it would be more canonical to use a round version of index here. Migrating to Mongo 4.2 will allow use to use the $round step.
```
vdestraitt, line 572: Actually `$trunc` is good for what you want to do. I don't understand the comment ?

src/lib/translators/mongo.ts
```diff
@@ -527,6 +527,91 @@ function transformSplit(step: Readonly<S.SplitStep>): MongoStep {
...
+    firstQuartileIndex: { $divide: ['$count', 4] },
+    firstQuintileIndex: { $divide: ['$count', 10] },
+    minIndex: 0,
+  };
+  const varianceFormula: any = { $subtract: ['$average_sum_square', { $pow: ['$average', 2] }] };
+  const otherStatisticsFormula: any = {
+    count: 1,
+    average: 1,
+    variance: varianceFormula,
+    std: {
```
vdestraitt, line 547: what about [`$stdDevPop`](https://docs.mongodb.com/manual/reference/operator/aggregation/stdDevPop/) ? :)

src/lib/translators/mongo.ts
```diff
@@ -527,6 +527,91 @@ function transformSplit(step: Readonly<S.SplitStep>): MongoStep {
...
+    medianIndex: { $divide: ['$count', 2] },
+    firstQuartileIndex: { $divide: ['$count', 4] },
+    firstQuintileIndex: { $divide: ['$count', 10] },
+    minIndex: 0,
+  };
+  const varianceFormula: any = { $subtract: ['$average_sum_square', { $pow: ['$average', 2] }] };
+  const otherStatisticsFormula: any = {
+    count: 1,
+    average: 1,
+    variance: varianceFormula,
```
vdestraitt, line 546: what about a simple combination of `$pow` and [`$stdDevPop`](https://docs.mongodb.com/manual/reference/operator/aggregation/stdDevPop/) ? :)

src/lib/translators/mongo.ts
```diff
@@ -527,6 +527,91 @@ function transformSplit(step: Readonly<S.SplitStep>): MongoStep {
   ];
 }
 
+/** transform a 'statistics' step into corresponding mongo steps */
+function transformStatistics(step: Readonly<S.StatisticsStep>): MongoStep {
+  // Some useful formula:
+  const quintilesIndexFormula: any = {
+    maxIndex: { $subtract: ['$count', 1] },
```
vdestraitt, line 534: Oh I think here we should use `$max`, rather than complexifying the query...

src/lib/translators/mongo.ts
```diff
@@ -527,6 +527,91 @@ function transformSplit(step: Readonly<S.SplitStep>): MongoStep {
...
+    lastQuintileIndex: { $multiply: [{ $divide: ['$count', 10] }, 9] },
+    lastQuartileIndex: { $multiply: [{ $divide: ['$count', 4] }, 3] },
+    medianIndex: { $divide: ['$count', 2] },
+    firstQuartileIndex: { $divide: ['$count', 4] },
+    firstQuintileIndex: { $divide: ['$count', 10] },
+    minIndex: 0,
```
vdestraitt, line 540: Same comment as for max => `$min`

src/lib/steps.ts
```diff
@@ -93,6 +93,24 @@ export type DeleteStep = {
   columns: string[];
 };
 
+export type StatisticsStep = {
+  name: 'statistics';
+  column: string;
+  statistics: (
```
vdestraitt, line 99: After trying the step several times, I strongly believe it's a pain to have all statistics checked by default. It's quite overwhelming and in most cases I just want to have a look at 2 or 3 of them max, so I would prefer not to have anything checked by default and pick what I need

src/lib/translators/mongo.ts
```diff
@@ -527,6 +527,91 @@ function transformSplit(step: Readonly<S.SplitStep>): MongoStep {
...
+    {
+      $sort: { column: 1 },
+    },
+    {
+      $group: {
+        _id: 'whatever',
+        count: { $sum: 1 },
+        average_sum_square: computeColumnSquare(step) ? { $avg: '$column_square' } : undefined, // I am using this formula of the variance: avg(x^2) - avg(x)^2
+        average: computeAverage(step) ? { $avg: '$column' } : undefined, // No need to compute average if neither average or variance are requested
+        data: { $push: '$column' },
```
vdestraitt, line 603: Regarding the management of null values, a solution I suggest is to add an `$addField` stage before the next step to project a `dataNoNull` using the [`$filter`](https://docs.mongodb.com/manual/reference/operator/aggregation/filter/) operator, and to use this field as a basis for the quantiles computation

docs/_docs/user-interface/statistics.md
```diff
@@ -0,0 +1,34 @@
...
+- Mongo 3.6
+
+#### Where to find this step?
+
+- Widget `Compute`
+- Search bar
+
+#### Options reference
+
+<img src="../../img/docs/user-interface/statistics_step_form.jpg" width="350" />
```
vdestraitt, line 22:
```suggestion
<img src="../../img/docs/user-interface/statistics_step_form.png" width="350" />
```
(PNG, not JPG)

docs/_docs/user-interface/statistics.md
```diff
@@ -0,0 +1,34 @@
+---
```
vdestraitt, line 1: The technical doc is missing (in steps.md)

docs/_docs/user-interface/statistics.md
```diff
@@ -0,0 +1,34 @@
...
+
+- `Statistics`: the list of statistics to compute.
+
+#### Example
+
+<img src="../../img/docs/user-interface/statistics_example_conf.jpg" width="750" />
+
+This configuration results in:
+
+<img src="../../img/docs/user-interface/statistics_example_result.jpg" width="500" />
```
vdestraitt, line 34:
```suggestion
<img src="../../img/docs/user-interface/statistics_example_result.png" width="500" />
```
(PNG, not JPG)

docs/_docs/user-interface/statistics.md
```diff
@@ -0,0 +1,34 @@
...
+
+<img src="../../img/docs/user-interface/statistics_step_form.jpg" width="350" />
+
+- `Column`: the column to compute the statistics. WARNING: this column must be of type 'float' or 'integer'.
+
+- `Statistics`: the list of statistics to compute.
+
+#### Example
+
+<img src="../../img/docs/user-interface/statistics_example_conf.jpg" width="750" />
```
vdestraitt, line 30: This image is missing

docs/_docs/user-interface/statistics.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Column's Statistics
+permalink: /docs/statistics/
```
vdestraitt, line 3: A corresponding entry to the docs.yml file is missing if you want this nice page to be accessible by the public :)

src/components/stepforms/StatisticsStepForm.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div>
+    <StepFormHeader :title="title" :stepName="this.editedStep.name" />
+    <ColumnPicker
```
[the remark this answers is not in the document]
rhuille: I agree it would be nice to have this option ! Would it is possible to add it later in a v2 of this step ?

src/lib/translators/mongo.ts
rhuille: It's only available in Mongo 4.2, and we are using Mongo 4.0 Toucan backends for now...

src/lib/translators/mongo.ts
rhuille: It is actually related to your comment here:https://github.com/ToucanToco/weaverbird/pull/522#discussion_r422598341

I used a definition of quantile where their values are actual values in the data: `p-th quantile of order n is the value of the data which is above p/n% of other values and below (1-p/n)% of other values`.

 For instance, in the dataset [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11] the median is `5` (the value which "half of the data is smaller and at most and half is greater") not 5.5 (which is not a value inside the data).

With that definition, using `$trunc` or `$round` is the same for median, but for other quantiles it can end to a different result. With the dataset [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11] the 9th decile is `10` (with my definition of quantile: ) but with truncated index we get `9`...

src/lib/translators/mongo.ts
```diff
@@ -527,6 +527,91 @@ function transformSplit(step: Readonly<S.SplitStep>): MongoStep {
...
+      stepQuantileIndex.$addFields[`${statistic}Index`] = {
+        $trunc: quintilesIndexFormula[`${statistic}Index`], //TODO: it would be more canonical to use a round version of index here. Migrating to Mongo 4.2 will allow use to use the $round step.
+      };
+    }
+  }
+  const stepProjectStatistics: any = { $project: {} };
+  for (statistic of step.statistics) {
+    if (isQuintileStatistics(statistic)) {
+      stepProjectStatistics.$project[statistic] = {
+        $arrayElemAt: ['$data', `$${statistic}Index`],
```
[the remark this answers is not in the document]
rhuille: +1 for null values, I add a step to filter these null value.

For the median definition, I know the one you are proposing, but I was actually using an other one where their values are actual values in the data: `p-th quantile of order n is the value of the data which is greater *or equal* of p/n% of other values and smaller (1-p/n)% of other values`. (I was using it because it was easier to implement :p )

src/lib/translators/mongo.ts
rhuille: 👍

src/components/stepforms/StatisticsStepForm.vue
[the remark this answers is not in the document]
vdestraitt: Yep !

src/lib/translators/mongo.ts
vdestraitt: Oh yes alright!

docs/_docs/user-interface/statistics.md
vdestraitt: @rhuille an entry in the docs.yml file is still missing (your doc will not be indexed without that)

src/components/constants.ts
```diff
@@ -53,6 +53,7 @@ export const ACTION_CATEGORIES: ActionCategories = {
     { name: 'evolution', label: 'Compute evolution' },
     { name: 'cumsum', label: 'Cumulated sum' },
     { name: 'percentage', label: 'Percentage of total' },
+    { name: 'statistics', label: 'Statistics' },
```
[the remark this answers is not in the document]
vdestraitt: @rhuille the step is still not accessible from the "Other operations" of the columns menu. I think it would be nice though and does not cost much. What do you think?

src/components/stepforms/StatisticsStepForm.vue
```diff
@@ -0,0 +1,357 @@
...
+      order: 10,
+    },
+    {
+      label: 'first centile',
+      nth: 1,
+      order: 100,
+    },
+    {
+      label: 'last centile',
+      nth: 9,
```
vdestraitt, line 192:
```suggestion
      nth: 99,
```

rhuille:
Thanks a lot @vdestraitt for the review.
Here are the last missing changes:
 - Ive just added an entry in the docs.yml file: https://github.com/ToucanToco/weaverbird/pull/522/files#diff-c499ea545a22fba6d138353fded9bc34R33

src/components/stepforms/StatisticsStepForm.vue
rhuille: Done: https://github.com/ToucanToco/weaverbird/pull/522/files#diff-b5e903e6cb7cf07ee66e59ef9c10db7cR192

src/components/ActionMenu.vue
```diff
@@ -48,6 +48,9 @@
             <div class="action-menu__option" @click="openStep('replace')">Replace values</div>
             <div class="action-menu__option" @click="openStep('sort')">Sort values</div>
             <div class="action-menu__option" @click="createUniqueGroupsStep">Get unique values</div>
+            <div class="action-menu__option" @click="openStep('statistics')">
```
vdestraitt, line 51: Thank you ! I think that you may need to add a test though :)

PR #523, opened 2020-05-09
F/rank step

inesjoly:
- `rank` step to add a rank column to the dataset
rank_step
ranked_dataset


src/lib/translators/mongo.ts
```diff
@@ -432,7 +432,79 @@ function transformPivot(step: Readonly<S.PivotStep>): MongoStep[] {
...
+          newRoot: {
+            $mergeObjects: [
+              '$items.grouped_ranked_column',
+              { [step.rankColumnName || 'Rank']: '$items.rank' },
+            ],
+          },
+        },
+      },
+    ];
+  } else {
```
vdestraitt, line 482: You do not need this. It's strange to rank something based on nothing :)
Also, as already said, it will never be true because columnPicker always returns a column, and it's good!

src/components/constants.ts
```diff
@@ -33,6 +33,7 @@ export const ACTION_CATEGORIES: ActionCategories = {
   add: [
     { name: 'text', label: 'Add text column' },
     { name: 'formula', label: 'Add formula column' },
+    { name: 'rank', label: 'Add rank column' },
```
vdestraitt, line 36: I would rather add a `Rank` entry in the `Compute` widget

src/lib/translators/mongo.ts
```diff
@@ -432,7 +432,79 @@ function transformPivot(step: Readonly<S.PivotStep>): MongoStep[] {
...
+          _id: 'whatever',
+          items: {
+            $push: '$$ROOT',
+          },
+        },
+      },
+      {
+        $unwind: {
+          path: '$items',
+          includeArrayIndex: 'items.rank',
```
vdestraitt, line 463: Be careful: your rank starts at 0 here (you may want to add 1 afterwards)

src/lib/translators/mongo.ts
```diff
@@ -432,7 +432,79 @@ function transformPivot(step: Readonly<S.PivotStep>): MongoStep[] {
   ];
 }
 
-/** transform an 'replace' step into corresponding mongo steps */
+/** transform a 'rank' step into corresponding mongo steps */
+function transformRank(step: Readonly<S.RankStep>): MongoStep {
+  if (step.column) {
+    return [
```
vdestraitt, line 438: I think this solution is fine for dense ranking ([10, 15, 15, 20, 30] => [1, 2, 2, 3, 4]).
But in lots of cases it's not the expected ranking method, and you want to "skip" the rank of elements following duplicates ([10, 15, 15, 20, 30] => [1, 2, 2, 4, 5]) so that at the end the rank remains in line with the cumulated count of elements.
So we should make this an option in the step maybe (`dense: <true/false>`, `false` by default)

A nice inspiration of how to implement this: http://www.kamsky.org/stupid-tricks-with-mongodb/rank-and-dense-rank-in-aggregation

PR #526, opened 2020-05-11
Fix popover mount

rhuille:
Remove immediate watch on "visible" prop and handle the behaviour at mounted and destroy
Before: if popover is destroyed when visible is true, the slot content is never remove from DOM because destroyPositioning has not been called
Now: at popover destroy, the slot content is cleaned from DOM

### DIFF
--- a/src/components/Popover.vue
+++ b/src/components/Popover.vue
@@ -65,25 +65,24 @@ export default class Popover extends Vue {
   parent: HTMLElement | null = null;
   parents: HTMLElement[] = [];
 
-  get isBottom() {
-    return this.bottom;
-  }
-
-  hasBeenPositioning = false;
-
-  @Watch('visible', { immediate: true })
+  @Watch('visible')
   async onVisibleChange(visible: boolean) {
     if (!visible) {
-      // `destroyPositioning` must not be called if `setupPositioning` has not been called as well
-      if (this.hasBeenPositioning) {
-        this.hasBeenPositioning = false;
-        this.destroyPositioning();
-      }
+      this.destroyPositioning();
     } else {
-      // We need to wait for the "visible" element to become effectively visible before positioning it.
-      await this.$nextTick();
-      this.hasBeenPositioning = true;
       this.setupPositioning();
+    }
+  }
+
+  mounted() {
+    if (this.visible) {
+      this.setupPositioning();
+    }
+  }
+
+  beforeDestroy() {
+    if (this.visible) {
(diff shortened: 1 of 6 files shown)


src/components/Popover.vue
```diff
@@ -65,24 +65,23 @@ export default class Popover extends Vue {
   parent: HTMLElement | null = null;
   parents: HTMLElement[] = [];
 
-  get isBottom() {
-    return this.bottom;
+  beforeDestroy() {
+    if (this.visible) {
+      this.destroyPositioning();
+    }
```
davinov, line 71: :100: 
We don't enforce positioning rules of vue options in this repo. Nonetheless, it ease the read to order events following their order in the lifecycle of the component.
https://vuejs.org/v2/style-guide/#Component-instance-options-order-recommended

src/components/Popover.vue
```diff
@@ -108,10 +107,10 @@ export default class Popover extends Vue {
 
     // Attach listeners
     window.addEventListener('click', this.clickListener, { capture: true });
-    window.addEventListener('orientationchange', this.orientationchangeListener);
-    window.addEventListener('resize', this.resizeListener);
+    window.addEventListener('orientationchange', () => this.updatePosition());
+    window.addEventListener('resize', () => this.updatePosition());
     for (const p of parents) {
-      p.addEventListener('scroll', this.scrollListener);
+      p.addEventListener('scroll', () => this.updatePosition());
```
davinov, line 113: :+1: Cleaner and easier to read

tests/unit/popover.spec.ts
```diff
@@ -315,24 +314,40 @@ describe('Popover', function() {
...
       const destroyPositioningSpy: any = jest.spyOn(popoverWrapper.vm as any, 'destroyPositioning');
       popoverWrapper.setProps({ visible: false });
       expect(destroyPositioningSpy).toHaveBeenCalled();
     });
+
+    it('should call destroyPosition at destroy if visible', async function() {
+      createWrapper({ props: { visible: true } });
+
+      const destroyPositioningSpy: any = jest.spyOn(popoverWrapper.vm as any, 'destroyPositioning');
+      popoverWrapper.destroy();
```
davinov, line 341: Outside this PR scope: should we not destroy wrappers for other tests too to avoid memory leaks?
Maybe it's not necessary due to the way Jest runs tests, compared to mocha.
@JeremyPinhel I'd like your opinion on this, thanks! :)

PR #532, opened 2020-05-12
Sort unique values in alphanumeric order

rhuille:
Before: the values were sorted by count

Before:
</img>

After:
</img>

### DIFF
--- a/src/lib/dataset/helpers.ts
+++ b/src/lib/dataset/helpers.ts
@@ -76,7 +76,10 @@ function localUniqueStats(dataset: DataSet) {
  * item's rank is lesser than the other, any positive value otherwise.
  */
 function _statCompare(stat1: ColumnValueStat, stat2: ColumnValueStat) {
-  return stat2.count - stat1.count;
+  return JSON.stringify(stat1.value).localeCompare(JSON.stringify(stat2.value), 'en', {
+    numeric: true,
+    caseFirst: 'lower',
+  });
 }
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Changed
 
-- The way we handle backendMessages (error/warning) by `BackendService`
+- better handle of backend messages (error or warning) from `BackendService`
+- Unique values are now sorted in alphanumeric order
 - Parameters order in text and formula steps forms
 - Deactivate date conversion option depending on columnn datatype in `DataTypesMenu`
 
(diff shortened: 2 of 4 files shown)


src/lib/dataset/helpers.ts
```diff
@@ -76,7 +76,7 @@ function localUniqueStats(dataset: DataSet) {
  * item's rank is lesser than the other, any positive value otherwise.
  */
 function _statCompare(stat1: ColumnValueStat, stat2: ColumnValueStat) {
-  return stat2.count - stat1.count;
+  return JSON.stringify(stat1.value).localeCompare(JSON.stringify(stat2.value));
```
vdestraitt, line 79: IMO if we have `['Paris', 'berlin', 'paris', 'Berlin']`, we want the output order to be `['Berlin', 'berlin', 'Paris', 'paris']` (or `['berlin', 'Berlin', 'paris',  'Paris']` is also fine, whatever), and not `['Berlin', 'Paris', 'berlin', 'paris']`.

so maybe:

```suggestion
  return JSON.stringify(stat1.value).toLowerCase().localeCompare(JSON.stringify(stat2.value));
```

?

Could you please add a test for this one if you agree ?

rhuille: Aaaah right !! Good point !

rhuille: I added these test to ensure the sort is case insensitive: https://github.com/ToucanToco/weaverbird/pull/532/files#diff-072d2eb768b90ae7ae46f58661967f8aR535

davinov, line 79: Nice to use localeCompare there!
Some tips that can be useful to have a better sort:
- explicit the locale parameter to 'en', to avoid using the browser's locale which could lead to unreliable results
- try {sentityvity: 'base'}: it will collide upper and lower case letters, and diacritics (accents) a well, so it could be interesting (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator) 
- you can test the option {numeric: true} so numerical values would be correctly sorted (1 < 2 < 10 and not 1 < 10 < 2)  => test this against regular strings to see it doesn't break their sort (in that case better keep that off)

davinov: Actually, sensitivity: "base", as lowercasing results, could create unpredictable orders.
Better to explicitly set which comes first using `caseFirst` upper or lower.

PR #534, opened 2020-05-13
Fix popover position (again)

rhuille:
- add an "handleEvent" method added to listener and then removed when destroy positioning.
- remove v-if: the element is removed at mount. It is then added to body when setup positioning.
- add a `.stop` propogation event modifier on action toolbar

### DIFF
--- a/src/components/Popover.vue
+++ b/src/components/Popover.vue
@@ -1,5 +1,5 @@
 <template>
-  <span v-if="visible" class="popover" :style="elementStyle">
+  <span class="popover" :style="elementStyle">
     <slot />
   </span>
 </template>
@@ -63,7 +63,10 @@ export default class Popover extends Vue {
 
   elementStyle: ElementPosition = {};
   parent: HTMLElement | null = null;
+  element: HTMLElement | null = null;
   parents: HTMLElement[] = [];
+  updatePositionListener!: () => void;
+  clickListener!: (Event: any) => void;
 
   @Watch('visible')
   async onVisibleChange(visible: boolean) {
@@ -74,20 +77,34 @@ export default class Popover extends Vue {
     }
   }
 
-  mounted() {
+  async mounted() {
     // Save original parent before moving into body to use its position
     this.parent = this.$el.parentElement;
+    // Remove element from parent: it will be added to body when `setupPosition`
+    if (this.parent !== null) {
+      this.parent.removeChild(this.$el);
+    } else {
+      throw new Error('The popover has no parent!');
+    }
 
     // Get all scrollable parents
     const parents = [];
     let { parent } = this;
 
-    while (parent !== document.body && parent !== null) {
+    while (parent !== document.body) {
       parents.push(parent);
-      parent = parent.parentElement;
+      if (parent.parentElement) {
+        parent = parent.parentElement;
+      } else {
+        break;
+      }
     }
 
     this.parents = parents;
+
+    // instantiate the events listener
+    this.updatePositionListener = () => this.updatePosition();
+    this.clickListener = (event: Event) => this.closeWhenClickOutside(event);
 
     if (this.visible) {
       this.setupPositioning();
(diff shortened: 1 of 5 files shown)


src/components/Popover.vue
```diff
@@ -74,17 +75,23 @@ export default class Popover extends Vue {
...
+    }
 
     // Get all scrollable parents
     const parents = [];
     let { parent } = this;
 
     while (parent !== document.body && parent !== null) {
       parents.push(parent);
-      parent = parent.parentElement;
+      parent = parent.parentElement as any; //FIXME: type any
```
davinov, line 94: Will `Element` be a more appropriate type?

davinov: `Element` is more generic than `HTMLElement`.
For example, it would be a `SVGElement` also.
I think this is why TS could disagree that `HTMLElement` is the type of a `node.parent`

Here is the associated MDN doc: https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement

src/components/Popover.vue
```diff
@@ -100,26 +107,44 @@ export default class Popover extends Vue {
...
     // Attach listeners
-    window.addEventListener('click', e => this.clickListener(e), { capture: true });
-    window.addEventListener('orientationchange', () => this.updatePosition());
-    window.addEventListener('resize', () => this.updatePosition());
+    window.addEventListener('click', this);
+    window.addEventListener('orientationchange', this);
+    window.addEventListener('resize', this);
     for (const parent of this.parents) {
-      parent.addEventListener('scroll', () => this.updatePosition());
+      parent.addEventListener('scroll', this);
```
davinov, line 137: Passing the whole `this` is clever, as the component has indeed a method implementing `handleEvent`. Therefore, it does respect the [EventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventListener) interface.

However, this is a highly uncommon pattern, and it binds the entire vm reference to a DOMelement (idk if it can cause any problem of garbage collecting/reference counting).

I'm not against merging this, but I would find much simpler to simply bind the callbacks to `this` and passing them explicitely. that would look like:

```js
setupPositioning() {
  // These can also be defined only once in created or mounted hooks
  this.clickListener = () => {...}
  this.updatePosition = () => {...}

  window.addEventListener('click', this.clickListener);
  window.addEventListener('orientationchange', this.updatePosition);
  ...
}


destroyPositionning() {
  window.removeEventListener('click', this.clickListener);
  ...
}
```

tests/unit/popover.spec.ts
```diff
@@ -223,12 +223,12 @@ describe('Popover', function() {
 
   it('should instantiate a popover', function() {
     createWrapper({ props: { visible: true } });
-    expect(_.isElement(popoverWrapper.element)).toBeTruthy();
+    expect(document.body.contains(popoverWrapper.element)).toBeTruthy();
   });
 
   it('should instantiate a popover', function() {
```
davinov, line 229:
```suggestion
  it('should not add the popover to the body', function() {
```

tests/unit/popover.spec.ts
```diff
@@ -223,12 +223,12 @@ describe('Popover', function() {
...
```
davinov, line 224:
```suggestion
  it("should instantiate the popover in document's body", function() {
```

src/components/Popover.vue
rhuille: The issue comes not from `HTMLElement ` but from the fact parent can also be `null`... in the loop we ensure parent is not null, but parent.parentElement can be null and typescript does not like that...

src/components/Popover.vue
rhuille: Done: https://github.com/ToucanToco/weaverbird/pull/534/files#diff-46e1611babaddbb455a4f41bfb470d75R123

src/components/Popover.vue
rhuille: https://github.com/ToucanToco/weaverbird/pull/534/files#diff-46e1611babaddbb455a4f41bfb470d75R94

PR #535, opened 2020-05-13
todate/fromdate :: add date format presets

vdestraitt:
Before that PR, we only proposed the user to enter a custom date format for the steps allowing to convert a text into date and vice versa.

Now the user can choose directly one of the presets we propose, which are the most common date formats used. And rather than working with abstract format definitions, he can base his choice on concrete examples of what a format looks like:

image


src/lib/translators/mongo.ts
```diff
@@ -336,6 +336,231 @@ function transformEvolution(step: Readonly<S.EvolutionStep>): MongoStep {
...
+        { case: { $eq: ['$_vqbTempMonth', '08'] }, then: 'August' },
+        { case: { $eq: ['$_vqbTempMonth', '09'] }, then: 'September' },
+        { case: { $eq: ['$_vqbTempMonth', '10'] }, then: 'October' },
+        { case: { $eq: ['$_vqbTempMonth', '11'] }, then: 'December' },
+        { case: { $eq: ['$_vqbTempMonth', '12'] }, then: 'December' },
+      ],
+    },
+  };
+
+  switch (step.format) {
```
rhuille, line 379:
```suggestion
  /**
   * `%d %b %Y`, `%d-%b-%Y`, `%d %B %Y`, `%b %Y` are date format `$dateToString` cannot handle
   * Therefore we design special query to handle these formats
   */
  switch (step.format) {
```

tests/unit/fromdate-step-form.spec.ts
```diff
@@ -1,13 +1,32 @@
...
+const localVue = createLocalVue();
+localVue.use(Vuex);
 
 describe('Convert Date to String Step Form', () => {
   const runner = new BasicStepFormTestRunner(FromDateStepForm, 'fromdate');
   runner.testInstantiate();
   runner.testExpectedComponents({
     'columnpicker-stub': 1,
-    'inputtextwidget-stub': 1,
+    'autocompletewidget-stub': 1,
```
rhuille, line 16:
```suggestion
    'AutocompleteWidget-stub': 1,
```

and idem for `InputTextWidget ` 😬
Well... I am being picking :p

tests/unit/templating.spec.ts
```diff
@@ -185,7 +185,7 @@ describe('Pipeline interpolator', () => {
       {
         name: 'fromdate',
         column: '<%= foo %>',
-        format: '<%= bar %>',
+        format: '<%= foo %>',
```
rhuille, line 188: Is that it important ? 🤔 Otherwise no worries !

src/components/stepforms/ToDateStepForm.vue
```diff
@@ -6,15 +6,31 @@
...
+      placeholder="Add column"
+      data-path=".column"
       :errors="errors"
     />
-    <InputTextWidget
-      v-if="this.translator === 'mongo40'"
-      class="dateFormat"
-      v-model.number="editedStep.format"
+    <AutocompleteWidget
+      v-if="translator === 'mongo40'"
```
rhuille, line 14: I don't understand why you want to hide this if it's not mongo ?

src/components/stepforms/ToDateStepForm.vue
```diff
@@ -6,15 +6,31 @@
...
+      @input="updateStepFormat"
+      :options="formats"
+      placeholder="Date format"
+      :trackBy="`format`"
+      :label="`label`"
+      :withExample="true"
+    />
+    <InputTextWidget
+      v-if="
+        translator === 'mongo40' &&
```
rhuille, line 27: Same here

src/components/stepforms/ToDateStepForm.vue
```diff
@@ -63,5 +118,15 @@ export default class ToDateStepForm extends BaseStepForm<ToDateStep> {
...
+
+  updateStepFormat(newFormat: FormatOption) {
+    if (newFormat.format === 'guess') {
+      this.editedStep.format = undefined;
+    } else if (newFormat.format === 'custom') {
+      this.editedStep.format = '';
+    } else {
+      this.editedStep.format = newFormat.format;
+    }
+  }
```
rhuille, line 130: Same comment than for the dateToString step: I suggest to make the `editedStep` a computed of two data edited by `Autocomplete` and `InputText` and the component archi will be much more clearer.

src/components/stepforms/ToDateStepForm.vue
```diff
@@ -6,15 +6,31 @@
...
+    <InputTextWidget
+      v-if="
+        translator === 'mongo40' &&
+          editedStep.format !== undefined &&
+          !datePresets.includes(editedStep.format)
+      "
+      class="customFormat"
+      v-model="editedStep.format"
+      name="Custom date format:"
       placeholder="%Y-%m-%d"
```
rhuille, line 34: You can use the `translator` here for example (really not mandatory) in the placeholder to say: `enter a {translator} date specification`

src/components/stepforms/widgets/Autocomplete.vue
```diff
@@ -9,7 +9,19 @@
       :track-by="trackBy"
       :label="label"
       @input="$emit('input', $event)"
-    />
+    >
+      <!-- If you want to use those templates you should provide a 'label' and 
+      'example' key in the options-->
```
rhuille, line 14: 👍

src/components/stepforms/widgets/Autocomplete.vue
```diff
@@ -47,6 +59,9 @@ export default class AutocompleteWidget extends Mixins(FormWidget) {
 
   @Prop({ type: String, default: undefined })
   label!: string;
+
+  @Prop({ type: Boolean, default: false })
+  withExample!: boolean;
```
rhuille, line 64:
```suggestion
  get  withExample(): boolean {
    return Object.keys(this.options).includes('example')
  };
```

src/components/DataTypesMenu.vue
```diff
@@ -105,6 +105,11 @@ export default class DataTypesMenu extends Vue {
       this.close();
     }
   }
+
+  openStep(stepName: PipelineStepName) {
```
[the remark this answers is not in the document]
vdestraitt: True... `git rebase` clumsiness! Thanks :)
Fixed then. Also, it's tested now!!

src/components/stepforms/FromDateStepForm.vue
```diff
@@ -30,24 +42,59 @@ import { Prop } from 'vue-property-decorator';
...
     InputTextWidget,
   },
 })
 export default class FromDateStepForm extends BaseStepForm<FromDateStep> {
-  @Prop({ type: Object, default: () => ({ name: 'fromdate', column: '', format: '' }) })
+  @Prop({ type: Object, default: () => ({ name: 'fromdate', column: '', format: '%Y-%m-%d' }) })
   initialStepValue!: FromDateStep;
 
   readonly title: string = 'Convert Column From Date to Text';
+  readonly formats: FormatOption[] = [
```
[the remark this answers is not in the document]
vdestraitt: I'm going with the intermediate `formatOptions` :)

src/components/stepforms/FromDateStepForm.vue
```diff
@@ -30,24 +42,59 @@ import { Prop } from 'vue-property-decorator';
...
+    { format: '%Y-%m', label: '%Y-%m', example: '1970-12' },
+    { format: '%Y/%m', label: '%Y/%m', example: '1970/12' },
+    { format: '%m-%Y', label: '%m-%Y', example: '12-1970' },
+    { format: '%m/%Y', label: '%m/%Y', example: '12/1970' },
+  ];
+  readonly datePresets = this.formats
+    .filter(d => d.format !== 'guess' && d.format !== 'custom')
+    .map(d => d.format);
+
+  get format(): FormatOption {
```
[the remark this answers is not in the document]
vdestraitt: I'm going with `selectedFormat` :)

tests/unit/fromdate-step-form.spec.ts
vdestraitt: I pass this one, as this convention is not used elsewhere...

tests/unit/templating.spec.ts
vdestraitt: no it's not :) whatever we test that it does not try to innterpolate

src/components/stepforms/ToDateStepForm.vue
vdestraitt: Because mongo3.6 does not support for providing a format. It only tries to infer.
See https://docs.mongodb.com/manual/reference/operator/aggregation/dateFromString/
(the `format` parameter is only supported from version 4.0)

I will add a comment though (there is one in the steps technical doc but it's good to have it in the UI component too)

src/components/stepforms/ToDateStepForm.vue
vdestraitt: Oh yes it's nice!

src/lib/translators/mongo.ts
vdestraitt: :+1:

src/components/stepforms/widgets/Autocomplete.vue
vdestraitt: Hum, I'm not convinced because I prefer that we keep the option explicit (we may have cases where we want to use the key `example` in the options data structure, but not the templates)

src/components/stepforms/FromDateStepForm.vue
```diff
@@ -59,5 +106,13 @@ export default class FromDateStepForm extends BaseStepForm<FromDateStep> {
...
     this.editedStep.column = colname;
   }
+
+  updateStepFormat(newFormat: FormatOption) {
+    if (newFormat.format === 'custom') {
+      this.editedStep.format = '';
+    } else {
+      this.editedStep.format = newFormat.format;
+    }
+  }
```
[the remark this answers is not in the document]
vdestraitt: @rhuille I've tried to implement your suggestion which I agree looked more elegant, but actually it's not that easy because `editedStep` comes from the base class `BaseStepForm` so we cannot define it as a computed in this inherited class.

Also, about what you spotted as a complex computed, I do need to check the value of `editedStep` format in all cases so that my data is set correctly when I need to edit a step for example. So be it a data or a computed, I will not be able to get rid of the flow control implemented in the computed, which is not that complicated in my opinion though :)

Are you OK to let the code as it is or do you want to suggest something also?

src/components/stepforms/ToDateStepForm.vue
vdestraitt: Same problem as in `FromDateStepForm`... :(

vdestraitt:
@rhuille apart from the problem I faced to implement your main suggestion I fixed everything else so it's up for a new review! thank you

PR #538, opened 2020-05-13
DataTypesMenu :: Deactivate date conversion option for non string columns

vdestraitt:
If a column datatype is not a string, now the date conversion is deactivated. After several user feedback, I believe that this improves a bit the UX while avoiding impossible conversions that yield mongo errors difficult to understand for the users.

image

### DIFF
--- a/src/components/DataTypesMenu.vue
+++ b/src/components/DataTypesMenu.vue
@@ -2,25 +2,32 @@
   <popover :visible="visible" :align="alignLeft" bottom @closed="close">
     <div class="data-types-menu__body">
       <div class="data-types-menu__section">
-        <div class="data-types-menu__option" @click="createConvertStep('integer')">
+        <div class="data-types-menu__option--active" @click="createConvertStep('integer')">
           <span class="data-types-menu__icon">123</span>
           <span>Integer</span>
         </div>
-        <div class="data-types-menu__option" @click="createConvertStep('float')">
+        <div class="data-types-menu__option--active" @click="createConvertStep('float')">
           <span class="data-types-menu__icon">1.2</span>
           <span>Float</span>
         </div>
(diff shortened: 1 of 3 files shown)


src/components/DataTypesMenu.vue
```diff
@@ -140,6 +151,18 @@ export default class DataTypesMenu extends Vue {
   }
 }
 
+.data-types-menu__option--deactivated {
+  display: flex;
+  align-items: center;
+  cursor: pointer;
```
rhuille, line 157:
```suggestion
  cursor: not-allowed;
```

src/components/DataTypesMenu.vue
```diff
@@ -2,25 +2,31 @@
...
+        <div class="data-types-menu__option--active" @click="createConvertStep('text')">
           <span class="data-types-menu__icon">ABC</span>
           <span>Text</span>
         </div>
-        <div class="data-types-menu__option" @click="createConvertStep('date')">
+        <div
+          :class="{
+            'data-types-menu__option--active': columnTypes[columnName] === 'string',
+            'data-types-menu__option--deactivated': !(columnTypes[columnName] === 'string'),
+          }"
```
rhuille, line 21:
```suggestion
          }"
          title="Only string column can be converted to date"
```

rhuille: This will display a small native html tooltip with the content of "title"

rhuille:
I have made two optional suggestions ! :)

src/components/DataTypesMenu.vue
vdestraitt: Cool, thanks !

src/components/DataTypesMenu.vue
vdestraitt: Nice, thanks !

PR #546, opened 2020-05-15
Playground enhancements

rhuille:
- Setup data is done by calling the updateDataset action
- Remove example warning
- Display the vqb pipeline in console

before:
image

image

After:
image

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -123,11 +123,7 @@ class MongoService {
         dataset = annotateDataset(dataset, types[0]);
         dataset = autocastDataset(dataset);
       }
-      const response = { data: dataset };
-      if (dataset.paginationContext.totalCount > 50) {
-        response.warning = [{ type: 'warning', message: 'Il y a plus de 50 lignes dans le dataset'}];
-      }
-      return response;
+      return { data: dataset };
     } else {
       return {
         error: [{type: 'error', message: responseContent.errmsg}],
@@ -169,34 +165,6 @@ class MongoService {
 const mongoservice = new MongoService();
 const mongoBackendPlugin = servicePluginFactory(mongoservice);
 
-async function setupInitialData(store, domain = null) {
-  const collections = await mongoservice.listCollections();
-  store.commit(VQBnamespace('setDomains'), {
-    domains: collections,
-  });
-  if (domain !== null) {
-    store.commit(VQBnamespace('setCurrentDomain'), {
-      currentDomain: domain,
-    });
-  } else {
-    const response = await mongoservice.executePipeline(
-      store,
-      store.getters[VQBnamespace('pipeline')],
-      store.state[VQB_MODULE_NAME].pagesize,
-    );
-    if (response.error) {
-      store.commit(VQBnamespace('logBackendMessages'), { backendMessages: response.error } );
-    } else {
-      if (response.warning) {
-        store.commit(VQBnamespace('logBackendMessages'), { backendMessages: response.warning });
-      }
-      store.commit(VQBnamespace('setDataset'), {
-        dataset: response.data,
-      });
-    }
-  }
-}
-
 async function buildVueApp() {
   Vue.use(Vuex);
   const store = new Vuex.Store({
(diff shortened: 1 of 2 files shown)


playground/dist/index.html
```diff
@@ -114,12 +130,23 @@ <h1 class="vqb__drop-zone__message">
...
+        <div class="error-message" v-if="backendErrors.length" v-for="error in backendErrors">
+          <i class="fas fa-exclamation-triangle"></i> {{ error.message }}
+        </div>
+        <div class="debug-panel__code">
+          <div class="debug-panel__code-mongo">
+            <div>Mongo</div>
+            <pre>{{ mongoQuery }}</pre>
+          </div>
+          <div class="debug-panel__code-vqb-pipeline">
+            <div>VQB pipeline</div>
```
davinov, line 146: Nice!
I'd prefer:
- to avoid showing "VQB" to users: it's not the name of the project anymore and the acronym won't make any sense for them. "Pipeline" is good enough I think :)
- instead of Mongo: "Mongo query" or "translated pipeline" (if we plan in the future to swap translators)
- invert the order of panel: usually in apps like this, input in on the left and generated/transpiled/converted stuff is at right (see coffeescript, pug or other websites for examples)

playground/app.js
```diff
@@ -288,21 +260,30 @@ async function buildVueApp() {
...
+        } else {
+          return []
         }
-        const query = mongo40translator.translate(activePipeline);
+      },
+      mongoQuery: function() {
+        const query = mongo40translator.translate(this.activePipeline);
         return JSON.stringify(query, null, 2);
       },
+      vqbPipeline: function(){
```
davinov, line 284: formattedPipeline ? or pipelineAsJSON ?

davinov: (same for mongoQuery upwards)

playground/dist/index.html
```diff
@@ -114,12 +130,23 @@ <h1 class="vqb__drop-zone__message">
...
-      </div>
-      <div class="error-message" v-if="backendErrors.length" v-for="error in backendErrors">
-        <i class="fas fa-exclamation-triangle"></i> {{ error.message }}
+      <div v-if="isCodeOpened" class="debug-panel">
+        <div class="warning-message" v-if="backendWarnings.length" v-for="warning in backendWarnings">
+          <i class="fas fa-exclamation-triangle"></i> {{ warning.message }}
+        </div>
+        <div class="error-message" v-if="backendErrors.length" v-for="error in backendErrors">
+          <i class="fas fa-exclamation-triangle"></i> {{ error.message }}
+        </div>
```
davinov, line 139: I'm not sure we want to display that only when code panel is opened. This should appear to users anytime it's needed!

Instead, I suggest we remove the default warning (over 50 lines) which was a nice testing case but shouldn't appear in the playground.

playground/app.js
```diff
@@ -288,21 +260,30 @@ async function buildVueApp() {
...
+      const collections = await mongoservice.listCollections();
+      store.commit(VQBnamespace('setDomains'), { domains: collections });
+      store.dispatch(VQBnamespace('updateDataset'));
     },
     computed: {
-      code: function() {
+      activePipeline: function(){
         let activePipeline = this.$store.getters[VQBnamespace('activePipeline')];
         if (!activePipeline) {
           return '';
```
davinov, line 271: I think the expected type of this.activePipeline is not a string but a pipeline object.
Better to return `undefined ` or `null` here, and let the computed below stringify this result.

playground/dist/index.html
```diff
@@ -74,29 +68,51 @@
...
+
+      .debug-panel__code-mongo, .debug-panel__code-vqb-pipeline {
+        width: 50%;
+        padding: 10px;
+       
+      }
+
+      .debug-panel__code-vqb-pipeline {
+        border-left: 1px solid;
+      }
```
davinov, line 115: :sparkles:

playground/dist/index.html
rhuille: image

PR #550, opened 2020-05-18
Add the 'left outer' to the schema and type enum for JoinStep.

### LINKED ISSUE #549: When I choose a left outer join type, I don't see an error message.
The `left outer` join type is not an accepted value by the join step.

It should be.

erenard:
Fixes #549 

Add `left outer` to the lists of join types.

### DIFF
--- a/src/components/stepforms/JoinStepForm.vue
+++ b/src/components/stepforms/JoinStepForm.vue
@@ -39,7 +39,7 @@
 import { Prop } from 'vue-property-decorator';
 
 import { StepFormComponent } from '@/components/formlib';
-import { default as JoinStepFormSchema } from '@/components/stepforms/schemas/join.ts';
+import JoinStepFormSchema from '@/components/stepforms/schemas/join.ts';
 import AutocompleteWidget from '@/components/stepforms/widgets/Autocomplete.vue';
 import JoinColumns from '@/components/stepforms/widgets/JoinColumns.vue';
 import ListWidget from '@/components/stepforms/widgets/List.vue';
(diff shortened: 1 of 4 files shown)


src/components/stepforms/schemas/join.ts
```diff
@@ -18,7 +18,7 @@ export default {
     },
     type: {
       type: 'string',
-      enum: ['left', 'inner'],
+      enum: ['left', 'inner', 'left outer'],
```
adimascio, line 21: Thanks for the fix!

I'm just wondering: shouldn't the `join.ts` schema file be imported in `JoinStepForm.vue` in order to make sure they're always sync?

erenard: Good suggestion, thanks.

src/components/stepforms/JoinStepForm.vue
```diff
@@ -39,6 +39,7 @@
 import { Prop } from 'vue-property-decorator';
 
 import { StepFormComponent } from '@/components/formlib';
+import { default as JoinStepFormSchema } from '@/components/stepforms/schemas/join.ts';
```
adimascio, line 42:
```suggestion
import JoinStepFormSchema from '@/components/stepforms/schemas/join.ts';
```

adimascio:
[As a conceptor, i would like to be able to do a "left outer" join in the VQB](https://trello.com/c/uerXPvlf/4131-as-a-conceptor-i-would-like-to-be-able-to-do-a-left-outer-join-in-the-vqb)

PR #556, opened 2020-06-15
Always show the result's row count.

erenard:
Bug_VQB


erenard:
[As an App Builder, I always want to see my rows number in the VQB](https://trello.com/c/eyS4hkMK/4152-as-an-app-builder-i-always-want-to-see-my-rows-number-in-the-vqb)

PR #557, opened 2020-06-18
Encapsulate VQB store in component

rhuille:
From now on, vqb store should not be used outside vqb
Vqb is an input component to edit a dictionary of pipelines, called `pipelines` with a `currentPipelineName` context because it can display and edit only one pipeline at the time.
NB: we cannot pass the vqb only one pipeline because, this pipeline is link to others.

We can still use the store from outside, BUT we aim at not using it by convention.


playground/dist/index.html
```diff
@@ -127,7 +127,21 @@ <h1 class="vqb__drop-zone__message">
...
       </div>
-      <vqb></vqb>
+      <div class="pipeline-selector">
+        Pipelines:
+        <select :value="currentPipelineName" @input="currentPipelineName = $event.target.value">
+          <option v-for="pipelineName in pipelinesNames" :key="pipelineName" :value="pipelineName">
+            {{ pipelineName }}
+          </option>
+        </select>
+      </div>
```
erenard, line 137: Pourquoi ne pas avoir gardé le composant pipeline selector ici ?

src/store/backend-plugin.ts
```diff
@@ -112,7 +112,8 @@ export function servicePluginFactory(service: BackendService) {
         mutation.type === 'vqb/selectStep' ||
         mutation.type === 'vqb/setCurrentDomain' ||
         mutation.type === 'vqb/deleteStep' ||
-        mutation.type === 'vqb/setCurrentPage'
+        mutation.type === 'vqb/setCurrentPage' ||
+        mutation.type === 'vqb/setCurrentPipelineName'
       ) {
         store.dispatch('vqb/updateDataset');
       }
```
erenard, line 119: On ne le ferai pas le TODO juste au dessus ?

src/components/Vqb.vue
```diff
@@ -1,32 +1,108 @@
...
     ResizablePanels,
   },
 })
-export default class Vqb extends Vue {}
+export default class Vqb extends Vue {
+  @Prop({
+    type: Object,
+    required: true,
+  })
+  pipelines!: { [name: string]: Pipeline };
```
davinov, line 33: Having a property here and in the store is kinda weird: which one holds the truth? The current prop value? The store?
The VQB component has authority on its store, but not on its props (this belongs to is parent).
Keeping the two in sync would mean watching the store state, emitting events from the VQB to ask the parent component to update the prop (upward), and when the prop changes, change the store value accordingly (downward). To avoid any loop, we would need to be careful about sharing the object references and break the loop when they are the same... This seems a bit complicated, doesn't it? ;)

So let's represent the store elements like this component's data instead of a prop.
In that case, the prop here is not "pipelines", but "initialPipelines". When the component instantiate, it loads into its store this pipelines. They are not to be modified after by the parent.
When working in the VQB, the pipelines in the store gets regularly updated.
We can still emit them to a parent component, but the prop initialPipelines will never be read again to avoid any loop.

(Note: it would not be impossible to achieve this kind of sync IMO. A simple example can be found in https://github.com/ToucanToco/tucana/blob/master/src/scripts/studio/utils/tc-code-editor/monaco/TcMonacoEditor.vue#L121 but it's already not trivial, and it's only for a very simple case with only one prop "value". This is what makes me think to other options).

PR #561, opened 2020-07-03
Pipeline can start by a reference to an other Pipeline

rhuille:
Also, the join and append step can make a reference to a source

TODO: 
- [ ]  update the doc of `domain-step` 
- [ ]  create the doc of  `source-step`


rhuille:
For now, I do not change the assumption that "dereferencing" a pipeline is an independent process of translation. But I sense, that dereferencing should be part of the translation process cf. https://github.com/ToucanToco/weaverbird/issues/564

PR #566, opened 2020-07-06
fix(multiselect): force the multiselect to open bellow

erenard:
Force the `vue-multiselect` component to open bellow.
This is a desirable behavior because no space is reserved above the widget.

Before (bad user experience):
BugMultiselect

After (more space is used below the widget):
BugMultiselectFix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.19.2] - 2020-07-06
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


erenard:
[As an app designer I can select the operation I want to proceed in the permissions interface](https://trello.com/c/KbCeTya2/4223-as-an-app-designer-i-can-select-the-operation-i-want-to-proceed-in-the-permissions-interface)

PR #569, opened 2020-07-20
fix: allow to filter with empty or undefined value

alice-sevin:
Removed minLength on filter input to allow empty or undefined value.
Empty input won't raise error anymore.

### DIFF
--- a/tests/unit/filter-step-form.spec.ts
+++ b/tests/unit/filter-step-form.spec.ts
@@ -20,12 +20,12 @@ describe('Filter Step Form', () => {
       data: {
         editedStep: {
           name: 'filter',
-          condition: { column: 'toto', operator: 'eq', value: '' },
+          condition: { column: '', operator: 'eq', value: '' },
         },
       },
       errors: [
         { keyword: 'if', dataPath: '.condition' },
-        { keyword: 'minLength', dataPath: '.condition.value' },
+        { keyword: 'minLength', dataPath: '.condition.column' },
       ],
     },
   ]);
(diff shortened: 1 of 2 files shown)


tests/unit/filter-step-form.spec.ts
```diff
@@ -20,12 +20,12 @@ describe('Filter Step Form', () => {
       data: {
         editedStep: {
           name: 'filter',
-          condition: { column: 'toto', operator: 'eq', value: '' },
+          condition: { column: 'toto', operator: 'eq', value: {} },
```
davinov, line 23: OK! IMO, we could have a more relevant test on this validation.
With the widgets available in the forms, I don't see any case where users can send an object as value (`value: {}`)
If you want to keep a validation test, I suggest testing the column name instead (it shouldn't be empty) https://github.com/ToucanToco/weaverbird/pull/569/files#diff-41a18ffe8fcf0208fad725aaee286fd9R91-R93

alice-sevin:
[As an AppBuilder, I can use templating even when my variable is undefined](https://trello.com/c/Lei0eyD1/4240-as-an-appbuilder-i-can-use-templating-even-when-my-variable-is-undefined)

tests/unit/filter-step-form.spec.ts
```diff
@@ -153,6 +153,35 @@ describe('Filter Step Form', () => {
...
+      formSaved: [
+        [
+          {
+            name: 'filter',
+            condition: { column: 'columnA', operator: 'eq', value: '' },
+          },
+        ],
+      ],
+    });
+  });
```
davinov, line 183: :+1: Thanks for this test :100:

tests/unit/filter-step-form.spec.ts
alice-sevin: Perfect, let's go for it!

PR #570, opened 2020-07-20
fix: keep `null` value filter for numeric columns

alice-sevin:
Keep null as null even with integer type (due to cast null values became 0)


alice-sevin:
[As an app designer I can use the "is null" or "is not null" in VQB](https://trello.com/c/Bz79ffT9/4222-as-an-app-designer-i-can-use-the-is-null-or-is-not-null-in-vqb)

PR #574, opened 2020-07-24
Pipeline can start by a reference to an other Pipeline

alice-sevin:
Also, the join and append step can make a reference to a domain

TODO: 
- [ ]  update the doc of `domain-step` 
- [ ]  create the doc of  `source-step`

### DIFF
--- a/src/store/state.ts
+++ b/src/store/state.ts
@@ -7,6 +7,7 @@ import { BackendError, BackendWarning } from '@/lib/backend-response';
 import { DataSet } from '@/lib/dataset';
 import { Pipeline, PipelineStepName } from '@/lib/steps';
 import { InterpolateFunction, PipelineInterpolator, ScopeContext } from '@/lib/templating';
+import { dereferencePipelines } from '@/store/utils/dereference-pipeline';
 
 export interface VQBState {
   /**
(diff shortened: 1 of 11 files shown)


alice-sevin:
[(5) ETQD un pipeline (a.k.a dataset) peut start / join / append un autre pipeline OU un domain (a.k.a les sources de pipelines)](https://trello.com/c/cNJOw3il/580-5-etqd-un-pipeline-aka-dataset-peut-start-join-append-un-autre-pipeline-ou-un-domain-aka-les-sources-de-pipelines)

src/store/mutations.ts
```diff
@@ -235,7 +235,7 @@ class Mutations {
     Vue.set(state.pipelines, state.currentPipelineName, pipeline);
     if (pipeline.length) {
       const firstStep = pipeline[0];
-      if (firstStep.name === 'domain') {
+      if (firstStep.name === 'domain' && typeof firstStep.domain === 'string') {
```
davinov, line 238: What does happen to `state.currentDomain` when a pipeline starts with another pipeline? Does that have an influence on something else?
(Searching in the project, I have trouble finding any use for this `state.currentDomain`. It seems directly computed from the current pipeline to me, so I wonder if we can just remove it.)

src/store/utils/dereference-pipeline.ts
```diff
@@ -0,0 +1,62 @@
...
+  for (const step of pipeline) {
+    switch (step.name) {
+      case 'domain':
+        // if domain is a pipeline replace step by selected dereferenced pipeline steps
+        dereferencedPipeline.push(..._getPipelineForDomain(step.domain as string, pipelines));
+        break;
+      case 'append':
+        dereferencedPipeline.push({
+          ...step,
+          pipelines: (step.pipelines as string[]).map(reference =>
```
davinov, line 44: to avoid doing `step.pipelines as string[]` and have better autocomplete in this switch/case statement, you could try:
```
case 'append':
  const append_step = step as AppendStep;
  // then, use append_step.pipelines
  dereferencedPipeline.push({...append_step, pipelines: append_step.pipelines.map( ref => ...)
  break;
case (...):
```

davinov: Actually, even this should be unnecessary, as Step should be a discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions) by its `name` property.
I wonder why this is not the case today

tests/unit/dereference-pipeline.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    const dererefencedPipeline: Pipeline = [
+      { name: 'domain', domain: 'toto' },
+      { name: 'filter', condition: { column: 'Price', operator: 'le', value: 10 } },
+      { name: 'filter', condition: { column: 'Price', operator: 'ge', value: 1200 } },
+      { name: 'append', pipelines: [pipelines['dataset1']] },
+    ];
+    expect(dereferencePipelines(pipeline, pipelines)).toStrictEqual(dererefencedPipeline);
+  });
+});
```
davinov, line 132: Very nice tests examples :100:

src/store/mutations.ts
alice-sevin: I'll check!

src/store/utils/dereference-pipeline.ts
alice-sevin: @davinov , as now domain can be a Pipeline or a String, in any case, i need to force type otherwise, it won't work. I checked youre idea but it didn't worked neither :cry: !

src/store/mutations.ts
alice-sevin: I found that it is useful to set default currentDomain if pipeline diseappear so i think that it's better to keep it.
If i remove the condition, there is broken tests on the Ci ( not in local ). Due to Reference : string or Pipeline as commentary below.

src/store/utils/dereference-pipeline.ts
davinov: @alice-sevin ahhh I think I see what you mean now!
I think the culprit is the AppendStep type: there should be two.
The `AppendStep` and the `DereferencedAppendStep`
The `AppendStep` has `pipelines: string[] ` while the `DereferencedAppendStep` has `pipelines: Pipeline[]`

Also, differentiating the two in terms of types will make sure we know at some point which one we have (is it dereferenced, is it not?). Do you think that would make the code clearer and cleaner?

alice-sevin: Yeah i will try it, may be a very good way to help

src/store/state.ts
```diff
@@ -187,7 +187,7 @@ export function preparePipeline(pipeline: Pipeline, state: VQBState) {
   }
   const { interpolateFunc, variables, pipelines } = state;
   if (pipelines && Object.keys(pipelines).length) {
-    pipeline = dereferencePipelines(pipeline, pipelines);
+    pipeline = dereferencePipelines(pipeline, pipelines) as Pipeline;
```
alice-sevin, line 190: @davinov , in any case we will have to force a type, here pipeline is reaffected so it causes issue (before Pipeline after DereferencedPipeline :cry: ). Remove all Old Pipeline occurencies will take a lot of time and i'm not sure were i'm able to keep original string Pipeline type

davinov: @alice-sevin Oh I understand, this may drag this ticket into a bigger refactoring than we expected.
So maybe the best option is to merge this as is, with the necessary `as` and discriminating conditions between pipeline reference and pipeline dereferenced, and to add an issue about making Pipeline into two different types.
We can do that later on.

PR #575, opened 2020-07-27
fix(mongo): join on column names with invalid characters

davinov:
Internal implementation of the mongo join translator requires a variable
name from each join column. Rules on variables names implies that
they cannot contain spaces an other non simple characters. Therefore, we
need to replace them by valid ones in the variable name.

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -53,6 +53,21 @@ function columnMap(colnames: string[]) {
   return _.fromPairs(colnames.map(col => [col, $$(col)]));
 }
 
+/**
+ * Generate a mongo [user variable](https://docs.mongodb.com/manual/reference/aggregation-variables/#user-variables)
+ * valid identifier from a column name.
+ *
+ * @param colname
+ */
+function columnToUserVariable(colname: string): string {
+  // User variable names can contain the ascii characters [_a-zA-Z0-9] and any non-ascii character.
+  const colnameWithoutInvalidChars = colname.replace(/[^_a-zA-Z0-9]/g, '_');
+
+  // User variable names must begin with a lowercase ascii letter [a-z] or a non-ascii character.
+  // Starting with the `vqb_` prefix guaranties that.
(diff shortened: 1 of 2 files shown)


davinov:
Closes #576

src/lib/translators/mongo.ts
```diff
@@ -53,6 +53,21 @@ function columnMap(colnames: string[]) {
   return _.fromPairs(colnames.map(col => [col, $$(col)]));
 }
 
+/**
+ * Generate a mongo [user variable](https://docs.mongodb.com/manual/reference/aggregation-variables/#user-variables)
+ * valid identifier from a column name.
+ *
+ * @param colname
+ */
+function columnToUserVariable(colname: string) {
```
PrettyWood, line 62:
```suggestion
function columnToUserVariable(colname: string): string {
```
doesn't cost anything

tests/unit/mongo.spec.ts
```diff
@@ -2485,6 +2485,74 @@ describe('Pipeline to mongo translator', () => {
...
+        right_pipeline: rightPipeline,
+        type: 'inner',
+        on: [
+          ['some column', 'some column'],
+          ['other column', 'other column'],
+        ],
+      },
+    ];
+    const querySteps = mongo36translator.translate(pipeline);
+    expect(querySteps).toEqual([
```
PrettyWood, line 2514:
```suggestion
    expect(querySteps).toStrictEqual([
```
doesn't cost anything as well

src/lib/translators/mongo.ts
```diff
@@ -53,6 +53,21 @@ function columnMap(colnames: string[]) {
   return _.fromPairs(colnames.map(col => [col, $$(col)]));
 }
 
+/**
+ * Generate a mongo [user variable](https://docs.mongodb.com/manual/reference/aggregation-variables/#user-variables)
```
PrettyWood, line 57: 👍 top !

PR #577, opened 2020-07-28
fix(pagination): always display pagination for multiple pages dataset

alice-sevin:
Pagination never displayed on last page of multiple pages dataset due to wrong condition.
We need to display pagination for every dataset that have more than one page.

### DIFF
--- a/src/components/Pagination.vue
+++ b/src/components/Pagination.vue
@@ -40,8 +40,8 @@ export default class Pagination extends Vue {
 
   @VQBModule.Mutation setCurrentPage!: MutationCallbacks['setCurrentPage'];
 
-  get showPager() {
-    return this.pageRows.max !== this.totalCount;
+  get showPager(): boolean {
+    return this.pageCount > 1;
   }
 
   get pageCount() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Changes
+
+### Fixed
+- Pagination: always display pagination for multiple pages dataset
+
 ## [0.19.3] - 2020-07-20
 
 ### Fixed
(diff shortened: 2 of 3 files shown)


alice-sevin:
[As a C, I can navigate back to previous pages when I'm at the last page of dataset viewer](https://trello.com/c/YLC91JW7/4253-as-a-c-i-can-navigate-back-to-previous-pages-when-im-at-the-last-page-of-dataset-viewer)

src/components/Pagination.vue
```diff
@@ -40,8 +40,8 @@ export default class Pagination extends Vue {
 
   @VQBModule.Mutation setCurrentPage!: MutationCallbacks['setCurrentPage'];
 
-  get showPager() {
-    return this.pageRows.max !== this.totalCount;
+  get showPager(): boolean {
```
adimascio, line 43: Excellent, thanks! Could you please add a test case to expose this bug (and make sure it's fixed :)) ?

alice-sevin: @adimasci , i thought these test was already enough (already implemented in `pagination-component.spec.ts`):
```
  it('should hide the pagination navigation if there is only one page', () => {
    const store = setupMockStore({
      dataset: {
        headers: [{ name: 'city' }, { name: 'population' }, { name: 'isCapitalCity' }],
        data: [
          ['Paris', 10000000, true],
          ['Marseille', 3000000, false],
        ],
        paginationContext: {
          totalCount: 2,
          pageno: 1,
          pagesize: 2,
        },
      },
      pagesize: 2,
    });
    const wrapper = mount(Pagination, { localVue, store });
    expect(wrapper.find('.pagination__list').exists()).toBeFalsy();
    expect(wrapper.find('.pagination-counter__current-min').exists()).toBeFalsy();
    expect(wrapper.find('.pagination-counter__current-max').exists()).toBeFalsy();
    expect(wrapper.find('.pagination-counter__total-count').text()).toEqual('2 rows');
  });
```

Let me know, if it doesn't respond to your issue :smile:

alice-sevin: If needed i also can add a specific test (going to last page, to check pagination on last page only) but isn't it too specific ?

adimascio: Well… as far as I understand it, there was a bug and yet, tests were passing, therefore I think there is a hole to fill :) Or am I missing something?

PR #578, opened 2020-07-28
fix(filters): keep value if typeof match on operator change

alice-sevin:
**Before:**  when changing operator with filter condition widget, value was reset to fit new type.
**After:** when changing operator with filter condition widget, value is reset only if its type don't match new type. Else keep previous value as selected one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Pagination: always display pagination for multiple pages dataset
+- Filters: keep value if typeof match on operator change
 
 ## Added
 - Util to compare value types
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Fixed
 - Pagination: always display pagination for multiple pages dataset
 
+## Added
+- Util to compare value types
+
 ## [0.19.3] - 2020-07-20
 
 ### Fixed
--- a/src/lib/helpers.ts
+++ b/src/lib/helpers.ts
@@ -78,7 +78,10 @@ export function* enumerate<T>(values: Iterable<T>, start = 0): Generator<[number
  * @param value the selected value
  * @param defaultValue the default to compare
  */
-export function compareType(value: ValueType | ValueType[], defaultValue: ValueType): ValueType {
+export function keepCurrentValueIfCompatibleType(
+  value: ValueType | ValueType[],
+  defaultValue: ValueType,
+): ValueType {
   if (Array.isArray(value)) return defaultValue;
   return typeof value === typeof defaultValue ? value : defaultValue;
 }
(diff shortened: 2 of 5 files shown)


alice-sevin:
[(3) As a C, I would like to keep my value if I'm changing my condition on a filter](https://trello.com/c/6ulC89cH/586-3-as-a-c-i-would-like-to-keep-my-value-if-im-changing-my-condition-on-a-filter)

src/lib/helpers.ts
```diff
@@ -69,3 +71,27 @@ export function* enumerate<T>(values: Iterable<T>, start = 0): Generator<[number
...
   }
 }
+
+/**
+ * Return default value if selected value has different typeof
+ *
+ * @param value the selected value
+ * @param defaultValue the default to compare
+ */
+export function compareType(value: ValueType | ValueType[], defaultValue: ValueType): ValueType {
```
davinov, line 81: I had trouble understanding what this function does due to its name.
To me, `compare` is supposed to return something saying that x is greater than y or the opposite.
Here, I have the impression it acts more like the `_.defaultTo` function of lodash/underscore : if the type is incompatible, then it return the default value, else it keep the actual value.

tests/unit/helpers.spec.ts
```diff
@@ -105,4 +112,25 @@ describe('castFromString', () => {
...
+  });
+
+  describe('compareArrayType', () => {
+    it('should return default if selected value is not an array', () => {
+      expect(compareArrayType('a', [])).toEqual([]);
+    });
+    it('should return selected value if its type is an array', () => {
+      expect(compareArrayType(['a'], [])).toEqual(['a']);
+    });
+  });
```
davinov, line 135: Tests made it clear what these functions do. Thanks!

src/lib/helpers.ts
davinov: Reading the next commit, may I suggest: `keepCurrentValueIfCompatibleType`? (it's a bit long, you'll probably find a better one ^^)

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Fixed
 - Filters: keep value if typeof match on operator change
+- Pagination: always display pagination for multiple pages dataset
```
davinov, line 7: In weaverbird, we generally avoid doing merges and prefer to rebase to keep a linear git history.
If you don't have to do this it's not a big deal, we can also squash and merge this PR (which will hide the merge) as long as the PR stays small.

alice-sevin: Good to know, i will take care for future conflicts :+1:

src/lib/helpers.ts
alice-sevin: keepCurrentValueIfCompatibleType seems great to me, i make the changes

PR #579, opened 2020-07-29
feat(variableInput): display value as tooltip on hover

alice-sevin:
Check behaviour here: https://dashboard.featurepeek.com/peek/jw0h9n53#/?path=/story/inputs-with-variables--wrapping-a-text-input

Display a tooltip with variable value, on variable tag hovering.

tooltip

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - "If...Then...Else": button to toggle step
 - "If...Then...Else": button to delete specific step
 - "If...Then...Else": button to add elseif nested condition
+- Variable input: add tooltip with value when hovering tag
 
 ## [0.19.3] - 2020-07-20
 
--- a/src/components/stepforms/widgets/VariableInput/VariableInput.vue
+++ b/src/components/stepforms/widgets/VariableInput/VariableInput.vue
@@ -5,6 +5,7 @@
         <div
           class="widget-input-variable__tag"
           v-tooltip="{
+            targetClasses: 'has-weaverbird__tooltip',
             classes: 'weaverbird__tooltip',
             content: variableValue,
             placement: 'bottom-center',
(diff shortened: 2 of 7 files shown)


alice-sevin:
[(5) ETQC, je vois la valeur courante d'une variable sélectionnée dans le VariableInput](https://trello.com/c/yCBReusy/454-5-etqc-je-vois-la-valeur-courante-dune-variable-s%C3%A9lectionn%C3%A9e-dans-le-variableinput)

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ## Added
 - Util to compare value types
+- Add v-tooltip dependency
```
davinov, line 11: That's very detailed!  It's fine if the changelog has just one entry by PR, it's good to keep it concise so it's easy to read.

src/components/stepforms/widgets/VariableInput/VariableInput.vue
```diff
@@ -170,6 +176,17 @@ export default class VariableInput extends Vue {
...
+  /**
+   * If the variable identifier is listed in the available variables display its value.
+   */
+  get variableValue() {
+    const matchingAvailableVariable = this.availableVariables.find(
+      aV => aV.identifier === this.variableIdentifier,
+    );
+    const value = matchingAvailableVariable?.value || '';
+    return Array.isArray(value) ? value.join(', ') : value;
+  }
```
davinov, line 188: Good! It will also be useful in the dropdown that shows the variable and their value: https://github.com/ToucanToco/weaverbird/blob/master/src/components/stepforms/widgets/VariableInput/VariableInput.vue#L47

tests/unit/variable-input.spec.ts
```diff
@@ -176,13 +176,18 @@ describe('Variable Input', () => {
       expect(wrapper.find('.widget-input-variable__variable-name').text()).toBe('hummus');
     });
 
+    it('should compute the variable value as empty', () => {
+      expect((wrapper as any).vm.variableValue).toBe('');
```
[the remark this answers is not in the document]
alice-sevin: @davinov , i checked to find a solution: i can check if a directive is called by using a mock but not the actual directive ( can't retrieve the arguments in any way :cry: ), to replace logic and check that the plugin is ok i added a test which check that a class have been added to the tag 'has-weaverbird__tooltip'. It's the best that i can do.

PR #581, opened 2020-08-03
fix: remove grouping of addFields steps in mongo query

alice-sevin:
We merged addFields.
It causes issue if we try to use another added column as variable ( she won't exist as she is added at the same time).
We need to degroup add fields.


alice-sevin:
[AS a Conceptor, I would like to have the right templating values when using VQB in my chart](https://trello.com/c/yqAG9fRD/4259-as-a-conceptor-i-would-like-to-have-the-right-templating-values-when-using-vqb-in-my-chart)

PrettyWood:
Can you add the fix in the changelog ?

PR #582, opened 2020-08-03
feat: allow to use availableVariables and variablesDelimiters from store

alice-sevin:
From now availableVariables and variableDelimiters (used in variableInput) can only be passed as props to widget.
It may be useful to use them in every widget and in QueryBuilder globally.
In prevision: added two mutations to populate state properties: `availableVariables` and `variableDelimiters`

### DIFF
--- a/src/lib/variables.ts
+++ b/src/lib/variables.ts
@@ -1,0 +1,18 @@
+interface AvailableVariable {
+  identifier: string; // how the variable will be written in the code
+  value: any; // current value of the variable
+  category?: string;
+  label: string;
+}
+
+export type VariablesBucket = AvailableVariable[];
+
+export interface VariablesCategory {
+  label: string | undefined;
+  variables: AvailableVariable[];
+}
+
+export interface VariableDelimiters {
+  start: string;
+  end: string;
+}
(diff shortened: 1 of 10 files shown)


alice-sevin:
[ETQC je peux ajouter des variables et delimiters globaux à mon VQB (3)](https://trello.com/c/iWAWswRj/650-etqc-je-peux-ajouter-des-variables-et-delimiters-globaux-%C3%A0-mon-vqb-3)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ## Added
 - Util to compare value types
 - "If...Then...Else": util file to display human format formula
+- Store: add two new mutations `setAvailableVariables` and `setVariablesDelimiters` to populate state properties `availableVariables` and `variablesDelimiters`
```
davinov, line 12: I think it's not necessary to add this in the functional changelog.
It will be more relevant to write here the external behavior of weaverbird will change, like when we will be using these properties in some form widgets, and when it will become necessary for weaverbird integrators to use this store mutations to make variables available there.

src/lib/variables.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export interface VariablesCategory {
+  label: string | undefined;
+  variables: AvailableVariable[];
+}
+
+export interface VariableDelimiters {
+  start: string;
+  end: string;
+}
```
davinov, line 18: Good idea to start to move variables related logic in a dedicated place

PR #583, opened 2020-08-03
feat: add variableInput to Autocomplete widget

alice-sevin:
Allow to use variables in Autocomplete Widget.
Move variable toggle button when parent has arrow
Disable multiselect when choosing a variable

### DIFF
--- a/src/components/stepforms/widgets/Autocomplete.vue
+++ b/src/components/stepforms/widgets/Autocomplete.vue
@@ -1,28 +1,35 @@
 <template>
   <div class="widget-autocomplete__container" :class="toggleClassErrorWarning">
     <label class="widget-autocomplete__label" v-if="name">{{ name }}</label>
-    <multiselect
+    <VariableInput
       :value="value"
-      :options="options"
-      :placeholder="placeholder"
-      :allow-empty="false"
-      :track-by="trackBy"
-      :label="label"
-      openDirection="bottom"
-      @input="$emit('input', $event)"
+      :available-variables="availableVariables"
+      :variable-delimiters="variableDelimiters"
+      @input="updateValue"
     >
-      <!-- If you want to use those templates you should provide a 'label' and 
+      <multiselect
+        :value="value"
+        :options="options"
+        :placeholder="placeholder"
+        :allow-empty="false"
+        :track-by="trackBy"
+        :label="label"
+        openDirection="bottom"
+        @input="updateValue"
+      >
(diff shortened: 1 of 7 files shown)


alice-sevin:
[ETQC je peux utiliser des variables dans le widget autocomplete (3)](https://trello.com/c/PxdgOZsa/651-etqc-je-peux-utiliser-des-variables-dans-le-widget-autocomplete-3)

src/components/stepforms/widgets/VariableInput/VariableInput.vue
```diff
@@ -84,6 +84,9 @@ export default class VariableInput extends Vue {
   @Prop()
   value!: any;
 
+  @Prop({ default: false })
+  hasArrow?: boolean;
```
davinov, line 88: Could you add a comment indicating what the `hasArrow` prop will have as effect in the VariableInput render ? That will help understanding it in the future.

tests/unit/variable-input.spec.ts
```diff
@@ -232,6 +232,15 @@ describe('Variable Input', () => {
       });
     });
   });
+
+  describe('with a parent component which has an arrow', () => {
+    beforeEach(() => {
+      wrapper.setProps({ hasArrow: true });
+    });
+    it('should move the variable chooser toggle button away from arrow', () => {
```
davinov, line 254: :+1:

src/components/stepforms/widgets/Autocomplete.vue
```diff
@@ -7,9 +7,11 @@
       :variable-delimiters="variableDelimiters"
       :has-arrow="true"
       @input="updateValue"
+      @choosing="chooseVariable"
     >
       <multiselect
         :value="value"
+        :disabled="isChoosingVariable"
```
davinov, line 14: That's a clever idea to avoid having problematic interactions while choosing.
However, it has several drawbacks:
- the internal state of the VariableInput needs to be synchronized with this upper-state variable (which is a shame, this state should only exists once somewhere, and not twice)
- this behavior will need to be adapted for each component we apply the VariableInput on, which is repetitive and diminish the interest of the VariableInput component itself.

Would you like to discuss a bit how we could solve this directly in the VariableInput component once and for all?

davinov:
Thanks @alice-sevin! Our variables nice UI are making their way into more widgets :smiling_imp: 
May I ask you to add a storybook entry for this Autocomplete widget with variables?

PR #590, opened 2020-08-06
feat: use variables in list widget

alice-sevin:


alice-sevin:
[ETQC je peux utiliser des variables dans le widget list (1)](https://trello.com/c/KYtgkCaW/655-etqc-je-peux-utiliser-des-variables-dans-le-widget-list-1)

src/components/stepforms/widgets/List.vue
```diff
@@ -11,6 +11,8 @@
             :is="widget"
             v-bind="componentProps"
             :value="child.value"
+            :available-variables="availableVariables"
+            :variable-delimiters="variableDelimiters"
```
rhuille, line 15: Why not store `availableVariables` et `variableDelimiters` into the vuex store ? In my understanding, all widgets will need these variables and they are global.

alice-sevin: @rhuille that's already the case but we want to pass it through all components ( store variables only keeped in steps) to allow to use components with props rather than with store

PR #595, opened 2020-08-14
fix(variableTag): display variableTag correctly in multiVariableInput

alice-sevin:
Fix cross position
Fix double trigger event with click
Fix tooltip position when multiinputtext is open

### DIFF
--- a/src/components/stepforms/widgets/MultiInputText.vue
+++ b/src/components/stepforms/widgets/MultiInputText.vue
@@ -303,6 +303,8 @@ export default class MultiInputTextWidget extends Vue {
 .widget-multiinputtext__multiselect {
   /deep/ .widget-variable__tag {
     display: inline-flex;
+    vertical-align: top;
+    margin-bottom: 12px;
     margin-right: 10px;
     padding: 0;
   }
--- a/src/components/stepforms/widgets/MultiInputText.vue
+++ b/src/components/stepforms/widgets/MultiInputText.vue
@@ -9,6 +9,7 @@
     >
       <multiselect
         class="widget-multiinputtext__multiselect"
+        :class="{ 'widget-multiinputtext__multiselect--big': isMultiselectBig }"
         :value="value"
         @input="updateValue"
         :options="options"
@@ -87,6 +88,11 @@ export default class MultiInputTextWidget extends Vue {
 
   get variableInputMode() {
     return this.multiVariable ? MultiVariableInput : VariableInput;
+  }
+
+  get isMultiselectBig() {
+    // add overflow to tag container when there is a lot of values
+    return Array.isArray(this.value) && this.value.length > 6;
   }
 
   updateOptions(newVal: string) {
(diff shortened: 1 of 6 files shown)


src/components/stepforms/widgets/MultiInputText.vue
```diff
@@ -89,6 +90,11 @@ export default class MultiInputTextWidget extends Vue {
     return this.multiVariable ? MultiVariableInput : VariableInput;
   }
 
+  get isMultiselectBig() {
+    // add overflow to tag container when tere is a lot of values
```
davinov, line 94:
```suggestion
    // add overflow to tag container when there is a lot of values
```

tests/unit/multi-input-text-widget.spec.ts
```diff
@@ -70,6 +70,26 @@ describe('Widget MultiInputText', () => {
...
+    const wrapper = mount(MultiInputTextWidget, {
+      propsData: {
+        value: ['a', 'b'],
+        variableDelimiters: { start: '{{', end: '}}' },
+      },
+    });
+    expect(wrapper.find('.widget-multiinputtext__multiselect--big').exists()).toBe(false);
+  });
+
+  it('...expect with a lot of value', () => {
```
davinov, line 83:
```suggestion
  it('...except with a lot of values', () => {
```

src/components/stepforms/widgets/VariableInputs/VariableTag.vue
```diff
@@ -12,7 +12,6 @@
     <span class="widget-variable__tag-name">{{ variableLabel }}</span>
     <i
       class="widget-variable__tag-close fa fa-times"
-      @click="removeVariableTag"
       tabindex="1"
       @keypress.enter.prevent="removeVariableTag"
       @mousedown.prevent="removeVariableTag"
```
davinov, line 17: Do you know why we need to use mousedown here rather than a click?

src/components/stepforms/widgets/MultiInputText.vue
```diff
@@ -265,6 +270,12 @@ export default class MultiInputTextWidget extends Vue {
   }
 }
 
+.widget-multiinputtext__multiselect--big.multiselect--active {
+  .multiselect__tags-wrap {
+    overflow-y: auto;
+  }
+}
```
davinov, line 277: It's not clear for my why we should not use an overflow when we have less than 6 values.
Could you please add a comment to describe why this is necessary?

src/components/stepforms/widgets/VariableInputs/VariableTag.vue
alice-sevin: To keep same behaviour that original multiselect one, it's more consistent!

davinov: Ah ok! I didn't know multiselect was using these event. Thanks for the answer :100:

src/components/stepforms/widgets/MultiInputText.vue
alice-sevin: Yes of course, it's because of tooltips they are not well displayed otherwise. I will let a comment

alice-sevin:
[(2) ETQC le tooltip de mon variable tag dans un multiinputtext est bien positionné.](https://trello.com/c/vFBYvs4b/715-2-etqc-le-tooltip-de-mon-variable-tag-dans-un-multiinputtext-est-bien-positionn%C3%A9)

PR #600, opened 2020-08-14
fix(variables): allow to use unique variable in multiInputText of fil…

alice-sevin:
Allow to use uniq variable (variableInput) rather than multiple variables (multiVariableInput) in MultiInputText of the filter editor (can be useful if variables values are array ).


alice-sevin:
[(2) ETQC je peux utiliser l'éditeur de permissions de tucana](https://trello.com/c/dBWu6VcT/717-2-etqc-je-peux-utiliser-l%C3%A9diteur-de-permissions-de-tucana)

PR #605, opened 2020-08-17
fix(multiselect): display the custom label

erenard:
This PR makes the `Multiselect` component use the `option[label]` as label, when option is an object and label is defined.

### DIFF
--- a/src/components/stepforms/widgets/Multiselect.vue
+++ b/src/components/stepforms/widgets/Multiselect.vue
@@ -19,15 +19,12 @@
         :taggable="true"
         :closeOnSelect="false"
         openDirection="bottom"
+        :customLabel="customLabel"
       >
-        <!-- If you want to use those templates you should provide a 'label' and 
-    'example' key in the options-->
-        <template v-if="withExample" slot="singleLabel" slot-scope="props">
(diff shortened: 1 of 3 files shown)


erenard:
[(3) ETQC dans append, quand j'ajoute des datasets, les tags correspondent juste au nom du dataset](https://trello.com/c/uDTZjh9t/719-3-etqc-dans-append-quand-jajoute-des-datasets-les-tags-correspondent-juste-au-nom-du-dataset)

src/components/stepforms/widgets/Multiselect.vue
```diff
@@ -124,6 +121,14 @@ export default class MultiselectWidget extends Mixins(FormWidget) {
     const identifier = extractVariableIdentifier(value, this.variableDelimiters);
     return identifier != null;
   }
+
+  customLabel(option) {
+    if (typeof option === 'object' && this.label) {
```
davinov, line 126: Maybe it would be better to test for `this.label != null`, just in case the property is one day `0` or something that evaluate falsy.

erenard: Done

PR #607, opened 2020-08-18
fix(dataset-selection): sort and remove duplicate names

erenard:
This PR sort the names of the datasets available for selection in the following steps:
- AppendStep
- DomainStep
- JoinStep
Duplicates names are also removed.


erenard:
[(1) ETQC je ne peux pas sélectionner le domain courant dans les datasets à combiner (dans append ou join)](https://trello.com/c/zoyDxfgJ/718-1-etqc-je-ne-peux-pas-s%C3%A9lectionner-le-domain-courant-dans-les-datasets-%C3%A0-combiner-dans-append-ou-join)

erenard:
[(1) ETQC dans les datasets à combiner (append/join), je n'ai pas de doublons dans la liste des datasets](https://trello.com/c/XBA4LtdL/720-1-etqc-dans-les-datasets-%C3%A0-combiner-append-join-je-nai-pas-de-doublons-dans-la-liste-des-datasets)

src/store/getters.ts
```diff
@@ -98,18 +98,18 @@ const getters: GetterTree<VQBState, any> = {
...
    * Return all available domains (including pipelines but excluding currentPipelineName)
    */
-  availablePipelines: (state: VQBState) =>
-    Object.keys(state.pipelines).filter((name: string) => name !== state.currentPipelineName),
-
+  availablePipelineNames: (state: VQBState) =>
+    Object.keys(state.pipelines)
+      .concat(state.domains)
+      .filter((name: string) => name !== state.currentPipelineName)
+      .sort((a, b) => a.localeCompare(b)),
```
davinov, line 105: OK to do this concatenation here, directly in the store, but I think the name `availablePipelineNames` is not right.
This contains domains and pipeline names, not just pipeline names. It's a bit confusing in terms of types.

Since we call these both "datasets" in the UI (see https://github.com/ToucanToco/weaverbird/blob/28bdcc5e28210cb3dc5aeb02b7b72ec9352aecf1/src/components/stepforms/JoinStepForm.vue#L7) may we rename that getter `availableDatasetsNames`?

davinov (changes requested):
Thanks! In particular for being careful about localCompare vs a basic sort :+1:  

Just one notice abut naming and it'll be good to merge I think :

PR #610, opened 2020-08-19
cumsum step :: simplify mongo translation and keep track of all columns

vdestraitt:
Before this PR, we would only keep columns referenced in the step form, whereas we would rather want to add the computation in a new column while keeping track of all the other ones. Now it's the case!
Also, we took the opportunity to simplify the mongo query.

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -376,36 +376,28 @@ function transformArgmaxArgmin(step: Readonly<S.ArgmaxStep> | Readonly<S.ArgminS
 function transformCumSum(step: Readonly<S.CumSumStep>): MongoStep {
   const groupby = step.groupby ?? [];
   return [
+    { $sort: { [step.referenceColumn]: 1 } },
     {
       $group: {
-        _id: columnMap([step.referenceColumn, ...groupby]),
-        [step.valueColumn]: { $sum: $$(step.valueColumn) },
-      },
-    },
-    { $sort: { _id: 1 } },
-    {
-      $group: {
-        _id:
-          step.groupby === undefined
-            ? null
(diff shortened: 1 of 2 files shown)


src/lib/translators/mongo.ts
```diff
@@ -376,36 +376,28 @@ function transformArgmaxArgmin(step: Readonly<S.ArgmaxStep> | Readonly<S.ArgminS
...
-            ? null
-            : Object.fromEntries(groupby.map(col => [col, `$_id.${col}`])),
-        [step.referenceColumn]: { $push: `$_id.${step.referenceColumn}` },
+        _id: step.groupby ? columnMap(groupby) : null,
         [step.valueColumn]: { $push: $$(step.valueColumn) },
+        _vqbArray: { $push: '$$ROOT' },
       },
     },
-    { $unwind: { path: $$(step.referenceColumn), includeArrayIndex: '_VQB_INDEX' } },
+    { $unwind: { path: '$_vqbArray', includeArrayIndex: '_VQB_INDEX' } },
```
davinov, line 387: Do you think it could be even easier with the [$mergeObjects](https://docs.mongodb.com/manual/reference/operator/aggregation/mergeObjects/index.html) operator?

([This answer](https://stackoverflow.com/a/60512462) made me think of it)

vdestraitt: Hum I'm not sure to see what you imagine at this stage (here I need to generate the index of documents in the `$_vqbArray` array). But I use your suggested trick just [below](https://github.com/ToucanToco/weaverbird/pull/610/files/53243e88671c04a067f449212cc920fbeec1f4cd#diff-b4949f396e00c36a178d54c27a25dcf3R399) to recover all my columns :)

erenard (approved):
LGTM

PR #612, opened 2020-08-24
fix(data-viewer): contextual menus now hide on dataset change

erenard:
This PR hides the column menus when the columns headers changes.
They are redefined each time a dataset is (re)loaded.

### DIFF
--- a/src/components/DataViewer.vue
+++ b/src/components/DataViewer.vue
@@ -59,7 +59,7 @@
 </template>
 <script lang="ts">
 import Vue from 'vue';
-import { Component } from 'vue-property-decorator';
+import { Component, Watch } from 'vue-property-decorator';
 
 import Pagination from '@/components/Pagination.vue';
 import { DataSet, DataSetColumn, DataSetColumnType } from '@/lib/dataset';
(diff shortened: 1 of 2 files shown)


erenard:
[[Mini Bug] AS a VQB User, when click on a column arrow and then another dataset, I would like the popup disappear.](https://trello.com/c/JVWbcJFA/4219-mini-bug-as-a-vqb-user-when-click-on-a-column-arrow-and-then-another-dataset-i-would-like-the-popup-disappear)

src/components/DataViewer.vue
```diff
@@ -8,12 +8,16 @@
...
+                v-for="column in formattedColumns"
                 :class="column.class"
                 :key="column.name"
                 @click="toggleColumnSelection({ column: column.name })"
               >
-                <span v-if="column.type" :class="iconClass" @click.stop="openDataTypeMenu(index)">
+                <span
+                  v-if="column.type"
+                  :class="iconClass"
+                  @click.stop="openDataTypeMenu(column.name)"
```
alice-sevin, line 19: :+1:

alice-sevin (approved):
Trop top, much better de pas passer par l'index! Thanks thanks

PR #613, opened 2020-08-25
feat (variables): add advanced variable option

alice-sevin:
- Add "Advanced variable option" to variable chooser
- Allow to use advanced variable in widget and steps even if no availableVariables are provided
( As we keep need to hide variableChooser in fields that can't be variable, we need to had a boolean from parent through variableInput ).


erenard:
[(2) ETQC j'ai une option "Add advanced variable" dans le dropdown des variables](https://trello.com/c/11efLIZc/762-2-etqc-jai-une-option-add-advanced-variable-dans-le-dropdown-des-variables)

erenard (approved):
LGTM !

PR #616, opened 2020-08-26
fix(muti-variable-input): fix button picker style and multiselect pick

alice-sevin:
- Fix position of button picker for variable in multiselect and multiinputtext
- Fix oldValue issue can be undefined at first (so value is never apply )


alice-sevin:
[(2) ETQC, j'ai un counter qui m'annonce le nombre de variables choisi si j'en ai plus de une.](https://trello.com/c/3AyTUkRL/774-2-etqc-jai-un-counter-qui-mannonce-le-nombre-de-variables-choisi-si-jen-ai-plus-de-une)

erenard (approved):

PR #617, opened 2020-08-26
feat(variableInput): add counter in button picker for multi selected …

alice-sevin:
Add a counter in multiVariableInputs when there is more than 2 variables to specifiy user how many variables he picked.

Before:
before
After with more than 2 variables (is always visible):
more
test

### DIFF
--- a/src/components/stepforms/widgets/VariableInputs/VariableInputBase.vue
+++ b/src/components/stepforms/widgets/VariableInputs/VariableInputBase.vue
@@ -11,11 +11,14 @@
         v-if="canBeVariable"
         class="widget-variable__toggle"
         :class="{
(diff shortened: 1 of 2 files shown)


alice-sevin:
[(2) ETQC, j'ai un counter qui m'annonce le nombre de variables choisi si j'en ai plus de une.](https://trello.com/c/3AyTUkRL/774-2-etqc-jai-un-counter-qui-mannonce-le-nombre-de-variables-choisi-si-jen-ai-plus-de-une)

erenard (approved):
LGTM

alice-sevin:
After UAT doesn't seems to be a good implementation to use

PR #620, opened 2020-08-28
Js translator

### LINKED ISSUE #519: JavaScript backend
For small datasets, transformations could be done client-side.
This could be handy to:
- fetch only some data once from a server and dispatch it across usages in the client
- enable offline transformations
- have a standalone version of weaverbird, serverless (not serverless as in "run functions on a server that we don't manage", but as running completely client-side without any server, including for the querying part)

davinov:
As a side-project, I started to develop a translator that transform pipelines into js functions.
This enables the playground to run client-side only. Once complete (all steps supported), it can also be used as reference for how transformations are expected to work (with tests) or to preview transformations when the database/server is not available.

Closes #519

### DIFF
--- a/.github/workflows/playground.yml
+++ b/.github/workflows/playground.yml
@@ -1,18 +1,15 @@
 # From https://docs.featurepeek.com/github-actions/
-name: Build storybook
+name: Build playground
 on: push
 jobs:
   build:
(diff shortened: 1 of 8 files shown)


davinov:
Closing this PR, as we've taken a different route (to implement backends in python)

PR #621, opened 2020-08-28
aggregate step :: option to keep original granularity and add aggregations in new columns

vdestraitt:
Before this PR, the aggregate step aggregated the input dataset, and all the columns that were not specified in the step form were lost.

This PR adds an option in the step to explicitly state wether to keep the original dataset granularity. If set to `true`, then the computed aggregations are added in new columns (usually leading to repetitions across rows).

The PR takes care of guaranteeing retrocompatibility with historical configurations.


src/lib/translators/mongo.ts
```diff
@@ -309,7 +309,10 @@ function transformAggregate(step: Readonly<S.AggregationStep>): MongoStep[] {
   const idblock: PropMap<string> = columnMap(step.on);
   const group: { [id: string]: {} } = {};
   const project: PropMap<any> = {};
+  let ret: MongoStep[] = [];
```
erenard, line 312: This let is not needed, just return at the end of the if/else block.
At the very least rename it `returned`, `result` or `steps`.

tests/unit/aggregate-step-form.spec.ts
```diff
@@ -24,17 +24,61 @@ describe('Aggregate Step Form', () => {
...
+    });
+
+    it('keepOriginalGranularity should be set to false if undefined in inititalValue', async () => {
+      const wrapper = runner.shallowMount(undefined, {
+        propsData: {
+          initialStepValue: {
+            name: 'aggregate',
+            on: [''],
+            aggregations: [],
+            keepOriginalGranularity: false,
```
erenard, line 49:
```suggestion
            keepOriginalGranularity: undefined,
```
From there, we don't need any `keepOriginalGranularity: false,` in this file

erenard (changes requested):
Thanks for this dev, just small things left.

src/lib/translators/mongo.ts
vdestraitt: 👍

tests/unit/aggregate-step-form.spec.ts
vdestraitt: 👍

vdestraitt: I only kept it in 1 test afterwards, (`runner.testValidate()`) that checks that we emit the same step as the input

erenard (approved):
LGTM, thanks !

PR #622, opened 2020-08-28
feat(advanced-variable): update variable delimiters mutation to add advanced delimiters

alice-sevin:
Need to use specific delimiters for advanced variable. Keep same logic than before for now ( willl maybe be better to improve logic by using variableDelimiters as an array or object at end)

### DIFF
--- a/src/store/mutations.ts
+++ b/src/store/mutations.ts
@@ -81,7 +81,13 @@ type VariableDelimitersMutation = {
   payload: Pick<VQBState, 'variableDelimiters'>;
 };
 
+type AdvancedVariableDelimitersMutation = {
+  type: 'setAdvancedVariableDelimiters';
+  payload: Pick<VQBState, 'advancedVariableDelimiters'>;
+};
+
 export type StateMutation =
+  | AdvancedVariableDelimitersMutation
   | AvailableVariablesMutation
   | BackendMessageMutation
   | DatasetMutation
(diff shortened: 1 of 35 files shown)


alice-sevin:
[(3) 4 / ETQD, je peux populer mon store avec des delimiters pour les advanced variables](https://trello.com/c/J08mtfjj/777-3-4-etqd-je-peux-populer-mon-store-avec-des-delimiters-pour-les-advanced-variables)

src/store/mutations.ts
```diff
@@ -329,9 +330,18 @@ class Mutations {
...
+    {
+      variableDelimiters,
+      advancedVariableDelimiters,
+    }: {
+      variableDelimiters: VariableDelimiters;
+      advancedVariableDelimiters: VariableDelimiters;
+    },
   ) {
     state.variableDelimiters = variableDelimiters;
+    if (advancedVariableDelimiters) {
```
erenard, line 342: Why testing this ? If `state.advancedVariableDelimiters` is undefined, what's the deal ?

erenard: Or two different mutations.

erenard (commented):
Juste une petite question:

src/store/mutations.ts
alice-sevin: Because we can have variableDelimiters but no advancedVariableDelimiters maybe, if an app doesn't want to use advanced variable he won't provide the second param. We can make two different mutations if you prefer, but i think that in a few days/weeks we will use all the variable delimiters as one and only object (as i said in the description) that's why it thought adding a new mutation would be a waste of time!

erenard (approved):
LGTM

PR #626, opened 2020-09-01
feat(codeEditor): use a specific config in a codeEditorWidget

alice-sevin:
Add `setAvailableCodeEditors`, a new method, to provide custom configs associated to a lang key.
CodeEditorWidget can now take a lang param.
It will use the custom config if available otherwise it will use the default one

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -10,6 +10,7 @@ const {
   servicePluginFactory,
   registerModule,
   setCodeEditor,
+  setAvailableCodeEditors,
 } = vqb;
 
 const TRANSLATOR = 'mongo40';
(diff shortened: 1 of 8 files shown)


alice-sevin:
[(5) ETQD, je peux utiliser un JSEditor dans weaverbird](https://trello.com/c/3Zt3WIPX/776-5-etqd-je-peux-utiliser-un-jseditor-dans-weaverbird)

src/components/stepforms/widgets/CodeEditorWidget.vue
```diff
@@ -38,12 +38,15 @@ export default class CodeEditorWidget extends Mixins(FormWidget) {
...
+  @Prop({ type: String, default: '' })
+  lang!: string;
+
   editedValue = this.value;
 
   isFocused = false;
 
-  CodeEditor = CodeEditor;
   // Code editor is set through a responsive data so it can be change after import
+  CodeEditor: CustomCodeEditor = CodeEditor;
```
erenard, line 49:
```suggestion
  codeEditor: CustomCodeEditor = CodeEditor;
```
C'est important de différencier un champ `data` d'un import.

playground/app.js
```diff
@@ -38,6 +39,60 @@ setCodeEditor({
...
+          },
+          blur: event => {
+            this.$emit('blur');
+          },
+          focus: event => {
+            this.$emit('focus');
+          },
+        },
+      });
+    },
```
davinov, line 93: Maybe we can avoid duplicating these two by creating a small function:
```js
function codeEditorForLang(lang) {
  return {
    props: ['value', 'placeholder'],
    render(createElement) {
      return createElement('textarea', {
        domProps: {
          value: this.value,
          placeholder: `OMG I have to write code in ${lang} in here`,
        },
        attrs: {
          type: 'text',
        },
        on: {
          input: event => {
            this.$emit('input', event.target.value);
          },
          blur: event => {
            this.$emit('blur');
          },
          focus: event => {
            this.$emit('focus');
          },
        },
      });
    };
}
```
and use it for both elements

src/components/stepforms/widgets/CodeEditorWidget.vue
erenard: Même si derrière c'est un switch sur un type d'Éditeur

erenard (changes requested):
Merci pour ce dev, par contre je pense qu'on devrait se conformer au camelCase pour les membres des composants.

davinov (commented):
Thanks for working on this!
I don't really see the value of keeping `CodeEditor` and `setCodeEditor`. Wouldn't it be easier to just replace it by `AvailableCodeEditors` everywhere?
This means a slight change of external API, but as it's only used in Toucan's project for now and we're still in 0.x, it's OK to change this API.

erenard (approved):

davinov (approved):
Thanks :)

I prefer the approach by lang as it paves the way for advanced translators ui customization, for example the Mongo translator could say its custom step needs a json editor and the Js translator needs a Js one.

:+1:

PR #641, opened 2020-09-08
fix(variables): fix types

alice-sevin:
Update types default and optionnal for variables


davinov (approved):
Thanks for correcting this!

It's weird that @Prop({default: undefined}) is not equivalent to just @Prop()
... maybe a TS weirdness :p

PR #642, opened 2020-09-08
fix(load-all-values): keep action menu open

alice-sevin:
Due to previous fix, we closed the action menu in any cases including when loading all values we should verify if the headers as change or if the current selected column is being modified ( don't close the action menu in this case ).

Before:
before

After:
after

### DIFF
--- a/src/components/DataViewer.vue
+++ b/src/components/DataViewer.vue
@@ -58,6 +58,7 @@
   <div class="data-viewer data-viewer--no-pipeline" v-else />
 </template>
 <script lang="ts">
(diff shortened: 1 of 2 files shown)


alice-sevin:
[AS a conceptor, when loading value on a dataprep dataset I would like my tooltip not to be closed](https://trello.com/c/Z7AphzrC/4299-as-a-conceptor-when-loading-value-on-a-dataprep-dataset-i-would-like-my-tooltip-not-to-be-closed)

erenard (approved):
Nice, thank you.

PR #646, opened 2020-09-15
fix: remove autocast in templating

davinov:
This was a not-so-good idea:
- same mechanism wasn't implement in hosts apps => created incoherences
between queries in the weaverbird preview, and what was actually
converted after in the host app
- column types can change during the pipeline => depending on the step
you come from, the interpolated pipeline is not always the same

Solution for this problem is to implement carefully the interpolator
function so that types are kept if a variable is declared alone in its
string. Such implem has landed in Tucana (main weaverbird host app for
now), so the auto-casting is not relevant anymore.

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -10,6 +10,7 @@ const {
   servicePluginFactory,
   registerModule,
   setAvailableCodeEditors,
+  exampleInterpolateFunc,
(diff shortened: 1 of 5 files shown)


erenard (approved):
LGTM

PR #650, opened 2020-09-21
fillna: change retrocompatibility implementation

vdestraitt:
Before this PR, we updated the `column` parameter to support array of strings as well as a simple string (to fill null values in several columns at once).
But at this end, we target to keep only a `columns` (in the plural) parameter which can only be a list of strings creates a mandatory `columns` parameter for the new, preferred configuration, while still supporting retrocompatibility with the deprecated `column` parameter.


alice-sevin (approved):
LGTM! Merci pour ce dev :)!

Issue #564, opened 2020-07-03
The BaseTranslator class should handle the dereference


rhuille:
A vqb pipeline has two states:
- A state "referenced" where the configuration of the `join` and `append` steps contains references to other pipelines, for example:

```
{
  name: 'append',
  pipelines: ['an_other_pipeline']
}
```

- a state "derefenced" where the references are replaced by the actual value of the pipelines:

```
{
  name: 'append',
  pipelines: [{name: 'domain', domain: 'plpop'}, ...]
}
```

The state `referenced` is used for edition.
The state `derefenced` is useless if not translated: we never dereference a pipeline without translating it right after. For this reason, **dereferencing a pipeline should be part of the translation process. This means the translator take as input a bunch of pipelines, not a single pipeline anymore.**

PR #654, opened 2020-10-05
fix(templating): append and join steps should template dereferenced pipelines

rhuille:
https://trello.com/c/WbYGUUxA/4308-as-a-conceptor-i-could-use-some-templating-on-my-right-dataset-when-im-using-the-join-vqb-step

### DIFF
--- a/src/lib/steps.ts
+++ b/src/lib/steps.ts
@@ -6,6 +6,10 @@ type PrimitiveType = number | boolean | string | Date;
 type Templatable<T> = T | string;
 export type Reference = Pipeline | string;
 
+export function isNotDeReferenced(pipelineOrReference: Reference): pipelineOrReference is string {
+  return typeof pipelineOrReference === 'string';
+}
+
 export type AddTextColumnStep = {
   name: 'text';
   text: string;
--- a/src/lib/steps.ts
+++ b/src/lib/steps.ts
@@ -6,7 +6,12 @@ type PrimitiveType = number | boolean | string | Date;
 type Templatable<T> = T | string;
 export type Reference = Pipeline | string;
 
-export function isNotDeReferenced(pipelineOrReference: Reference): pipelineOrReference is string {
+/**
+ * Some step can contains either:
+ * - a reference to the pipeline i.e. the name of the pipeline, then a string
+ * - the pipeline itself
+ */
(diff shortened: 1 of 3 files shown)


src/lib/steps.ts
```diff
@@ -6,6 +6,10 @@ type PrimitiveType = number | boolean | string | Date;
 type Templatable<T> = T | string;
 export type Reference = Pipeline | string;
 
+export function isNotDeReferenced(pipelineOrReference: Reference): pipelineOrReference is string {
```
davinov, line 9: Maybe `isReference` would a simpler name that avoids the double negation

src/lib/templating.ts
```diff
@@ -263,7 +271,11 @@ export class PipelineInterpolator implements StepMatcher<S.PipelineStep> {
   }
 
   join(step: Readonly<S.JoinStep>) {
-    return { ...step };
+    if (S.isNotDeReferenced(step.right_pipeline)) {
+      return { ...step };
+    } else {
+      return { ...step, right_pipeline: this.interpolate(step.right_pipeline) };
+    }
```
davinov, line 278: Is there a case when we arrive here and the pipeline is still referenced?
Maybe for validation when we submit the step form?

I suggest we indicate when this happens, so further readers will understand why the two cases are necessary.
(and if not... remove one :D)

davinov (commented):
Thanks, here a two questions:

src/lib/steps.ts
rhuille: https://github.com/ToucanToco/weaverbird/pull/654/commits/b5e00576c0ba6322cb27e9e757ea083c5f26ec0a

src/lib/templating.ts
rhuille: Bonne idée! https://github.com/ToucanToco/weaverbird/pull/654/commits/7891b0e4be90b70dafe86daf14f6817b06c05065

davinov:
@rhuille does this solve https://github.com/ToucanToco/weaverbird/issues/564 ? :)

rhuille:
Tout à fait ! :)

rhuille:
Tout à fait ! :)

davinov (approved):
Thanks!

PR #655, opened 2020-10-06
fix: set the multivariableinput z-index

erenard:
This PR sets the multiple variable input z-index.

Before:
image

After:
image


erenard:
[AS a Conceptor, the VQB step "Filter rows" when I have a value, I would like to see it](https://trello.com/c/mInvDofj/4330-as-a-conceptor-the-vqb-step-filter-rows-when-i-have-a-value-i-would-like-to-see-it)

alice-sevin (approved):
LGTM! Thanks :D!

PR #659, opened 2020-10-19
fix(multiinput-text): display dropdown without be overlapped by next input

alice-sevin:
Due to https://github.com/ToucanToco/weaverbird/pull/655 to fix overlapped input in filter condition, we created a bug.
Now in other step, as multi variable has a z-index 1. The dropdown are overlapped by the next input.

To fix it, we need to add a z-index only on the **active** multiinputtext ( added a z-index to variable toggler to display button even when multiinputtext is active).

Before:
before

After:
after

### DIFF
--- a/src/components/stepforms/widgets/MultiInputText.vue
+++ b/src/components/stepforms/widgets/MultiInputText.vue
@@ -153,7 +153,7 @@ export default class MultiInputTextWidget extends Vue {
   isolation: isolate;
 
   &.multiselect--active {
-    z-index: 0;
+    z-index: 1;
   }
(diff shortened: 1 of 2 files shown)


alice-sevin:
[AS a C, when I use a dropdown in the VQB there are somes values displayed above my dropdown](https://trello.com/c/AwqEnCsn/4366-as-a-c-when-i-use-a-dropdown-in-the-vqb-there-are-somes-values-displayed-above-my-dropdown)

davinov (approved):
Thanks for fixing this!

PR #662, opened 2020-10-20
fix(resizable-directive): should not sort the columns options

alice-sevin:
**What did you wizard :mage_woman: : ?**

It sorted the columns options to verify if columns are still the same or if we need to rerender the resizable table directive.
The thing is that we don't need to roll it :rofl: . So i sorted the columns in the ui (oupsssi).

Did a quick fix to avoid columns to get crazy.

Before:
before

After:
after


bloodstorms (approved):
Haha wooops

PR #667, opened 2020-10-22
pivot step: allow empty index

vdestraitt:
Before this PR, users were forced to specify at least 1 column in the index parameter, which is actually not expected in frequent cases, but lead users to create first a temporary column with the same value in all rows to be able to apply the step afterwards.

After this PR, we accept empty index.

### DIFF
--- a/src/lib/translators/mongo.ts
+++ b/src/lib/translators/mongo.ts
@@ -889,7 +889,7 @@ function transformPivot(step: Readonly<S.PivotStep>): MongoStep[] {
     addFieldsStep[`_vqbAppTmpObj.${col}`] = `$_id.${col}`;
   }
 
-  return [
+  const pivotMongoAggStages: MongoStep[] = [
     /**
      * First we perform the aggregation with the _id including the column to pivot
      */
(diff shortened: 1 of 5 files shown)


src/lib/translators/mongo.ts
```diff
@@ -932,17 +932,25 @@ function transformPivot(step: Readonly<S.PivotStep>): MongoStep[] {
...
+      { $addFields: addFieldsStep },
+      /**
+       * Then we replace the root of the documents tree to get our columns ready for
+       * our needed table-like, unnested format
+       */
+      { $replaceRoot: { newRoot: '$_vqbAppTmpObj' } },
+    ];
+  } else {
+    return [...partialRet, { $replaceRoot: { newRoot: '$_vqbAppTmpObj' } }];
+  }
```
davinov, line 953: This is a bit complex, and leads to repetition.
I think it reads way easier if you just mutate the array that you're creating:

Also `partialRet` does not convoy a lot of meaning IMHO

```js
let pivotMongoAggStages = [...]

if (Object.keys(addFieldsStep).length > 0) {
      /**
       * Then we include the index columns back in every document created in the previous step
       * (still accessible in the _id object)
       */
      pivotMongoAggStages.push({ $addFields: addFieldsStep })
}
/**
       * Then we replace the root of the documents tree to get our columns ready for
       * our needed table-like, unnested format
       */
pivotMongoAggStages.push({ $replaceRoot: { newRoot: '$_vqbAppTmpObj' } });

return pivotMongoAggStages;
```

What do you think?

tests/unit/mongo.spec.ts
```diff
@@ -1685,7 +1685,52 @@ describe('Pipeline to mongo translator', () => {
...
+            $arrayToObject: {
+              $zip: { inputs: ['$_vqbAppArray.column_3', '$_vqbAppArray.column_4'] },
+            },
+          },
+        },
+      },
+      { $replaceRoot: { newRoot: '$_vqbAppTmpObj' } },
+      { $project: { _id: 0 } },
+    ]);
+  });
```
davinov, line 1731: :100:

davinov (changes requested):
Thanks! Only a small code style detail:

src/lib/translators/mongo.ts
vdestraitt: Yes much better ! thx

davinov (approved):

PR #669, opened 2020-10-23
Fix types in JoinStepForm.vue

vdestraitt:
Everything is in the title. The linter was not happy....

### DIFF
--- a/src/components/stepforms/JoinStepForm.vue
+++ b/src/components/stepforms/JoinStepForm.vue
@@ -52,7 +52,14 @@ import { VQBModule } from '@/store';
 import BaseStepForm from './StepForm.vue';
 import Multiselect from './widgets/Multiselect.vue';
 
-const joinTypes = JoinStepFormSchema.properties.type.enum;
+const joinTypes = JoinStepFormSchema.properties.type.enum as JoinStep['type'][];
+
+interface DropdownOption {
+  label: string;
+  trackBy: string;
+  $isDisabled?: boolean;
+  tooltip?: string;
+}
 
 @StepFormComponent({
   vqbstep: 'join',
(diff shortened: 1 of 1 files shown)


src/components/stepforms/JoinStepForm.vue
```diff
@@ -52,7 +52,12 @@ import { VQBModule } from '@/store';
 import BaseStepForm from './StepForm.vue';
 import Multiselect from './widgets/Multiselect.vue';
 
-const joinTypes = JoinStepFormSchema.properties.type.enum;
+const joinTypes = JoinStepFormSchema.properties.type.enum as JoinStep['type'][];
+
+interface DropdownOption {
+  label: string;
+  trackBy: string;
+}
```
davinov, line 60: :+1:

src/components/stepforms/JoinStepForm.vue
```diff
@@ -105,10 +110,10 @@ export default class JoinStepForm extends BaseStepForm<JoinStep> {
     return this.availableDatasetNames.map(name => {
       const option = { label: name, trackBy: name };
       if (this.referencingPipelines.includes(name)) {
-        option['$isDisabled'] = true;
-        option[
-          'tooltip'
-        ] = `Circular reference: you cannot combine ${name} because it references the current dataset.`;
+        Object.assign(option, {
+          $isDisabled: true,
+          tooltip: `Circular reference: you cannot combine ${name} because it references the current dataset.`,
```
davinov, line 115: Does that means we should add `$isDisabled` and `tooltip` as optional members of the `DropdownOption` interface?

src/components/stepforms/JoinStepForm.vue
```diff
@@ -77,14 +82,14 @@ export default class JoinStepForm extends BaseStepForm<JoinStep> {
   joinColumns = JoinColumns;
   joinTypes: JoinStep['type'][] = joinTypes;
 
-  get rightPipeline(): object {
+  get rightPipeline(): DropdownOption {
     return {
-      label: this.editedStep.right_pipeline,
-      trackBy: this.editedStep.right_pipeline,
+      label: this.editedStep.right_pipeline as string,
+      trackBy: this.editedStep.right_pipeline as string,
```
davinov, line 88: Yes, this is quite unconfortable as  we do not have different types for steps with pipeline references and steps with compete pipelines. I think we should create an issue in that.

davinov (commented):
Thanks for clarifying types in this file :+1:

src/components/stepforms/JoinStepForm.vue
vdestraitt: True!

davinov (approved):

PR #672, opened 2020-10-23
unpivot step: support variables UI and interpolation

vdestraitt:


tests/unit/templating.spec.ts
```diff
@@ -1143,18 +1143,27 @@ describe('Pipeline interpolator', () => {
...
       },
     ];
-    expect(translate(pipeline)).toEqual(pipeline);
+    expect(translate(pipeline)).toEqual([
+      {
+        name: 'unpivot',
+        keep: ['bar', 'spam', 'column2'],
+        unpivot: ['bar', 'spam', 'column4'],
+        unpivot_column_name: '<%= foo %>',
+        value_column_name: '<%= foo %>',
```
davinov, line 1163: That means these fields can not be variabilized?

davinov (approved):
:+1:

tests/unit/templating.spec.ts
davinov: (make sense, but not sure we should write these as examples in tests)

PR #675, opened 2020-10-25
Pandas aggregate step

luc-leonard:


server/weaverbird/steps/aggregate.py
```diff
@@ -0,0 +1,47 @@
...
+import pandas as pd
+from mypy.applytype import Optional
+from pandas import DataFrame
+from pydantic import Field
+from pydantic.main import BaseModel
+
+from weaverbird.steps.base import BaseStep
+
+ColumnName = Union[str, int, float]
+AggregateFn = Literal['avg', 'sum', 'min', 'max']
```
vdestraitt, line 12: Careful here: we will need a mapping (e.g. `avg` in the weaverbird language => `mean` in pandas language)

server/weaverbird/steps/aggregate.py
```diff
@@ -0,0 +1,60 @@
...
+            aggs = self.make_aggregation(aggregation)
+            all_results[self.on + aggregation.new_columns] = grouped_by_df.agg(aggs)[
+                self.on + aggregation.columns
+            ]
+        # it is faster this way, than to trasnform the original df
+        if self.keep_original_granularity:
+            return df.merge(all_results, on=self.on, how='left')
+        return all_results
+
+    def make_aggregation(self, aggregation):
```
fspot, line 56: maybe put this in a function instead of a method since it does not rely on `self` attributes

fspot (approved):

PR #677, opened 2020-10-27
feat(server): rename step

fspot:
Add rename step in pandas backend.

:information_source:  I only implemented the new syntax (`toRename`), not the legacy one (`oldname` / `newname`).

### DIFF
--- a/server/weaverbird/steps/rename.py
+++ b/server/weaverbird/steps/rename.py
@@ -1,0 +1,17 @@
+from typing import List, Tuple
+
+from pandas import DataFrame
+from pydantic import Field
+
+from weaverbird.steps.base import BaseStep
+
+
+class RenameStep(BaseStep):
+    name = Field('rename', const=True)
+    to_rename: List[Tuple[str, str]] = Field(..., alias='toRename')
+
+    class Config:
+        allow_population_by_field_name = True
+
+    def execute(self, df: DataFrame, domain_retriever) -> DataFrame:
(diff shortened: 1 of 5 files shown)


server/weaverbird/steps/rename.py
```diff
@@ -0,0 +1,17 @@
...
+
+from pandas import DataFrame
+from pydantic import Field
+
+from weaverbird.steps.base import BaseStep
+
+
+class RenameStep(BaseStep):
+    name = Field('rename', const=True)
+    to_rename: List[Tuple[str, str]] = Field(..., alias='toRename')
```
fspot, line 11: :warning: ici j'ai mis `str` plutôt que `ColumnName = Union[str, int, float]` car 1) dans la doc de weaverbird on parle explicitement de strings et 2) le `Union[str, int, float]` fait un cast automatique en `str` de toute façon.

Si on veut vraiment gérer les noms de colonnes `int` et `float` il faut là aussi éviter que pydantic fasse du cast.

davinov (commented):
Thanks ! Just a question about the typings and pydantic cast:

server/weaverbird/steps/rename.py
fspot: Yes, tweaking the order is not a solution because if we put `int` before `str`, then `"2020"` will be casted to `2020`.

davinov (approved):
I took a deeper look into our front-end code, and it seems that we only support string column names for now. So it's OK to have ColumnName = str.

PR #679, opened 2020-10-28
feat(playground): use optional pandas backend

davinov:
Using `?backend=pandas` switches to the `server/playground.py` server.

### DIFF
--- a/server/playground.py
+++ b/server/playground.py
@@ -29,7 +29,7 @@ def after_request(response):
 @app.route('/', methods=['GET', 'POST'])
 def handle_request():
     if request.method == 'GET':
-        return json.dumps(get_available_domains())
+        return get_available_domains()
     elif request.method == 'POST':
         return execute_pipeline(request.get_json(), **request.args)
 
--- a/server/playground.py
+++ b/server/playground.py
@@ -36,4 +36,4 @@ def get_available_domains():
 def execute_pipeline(pipeline):
     executor = PipelineExecutor(lambda domain: DOMAINS[domain])
     df = executor.execute_pipeline(pipeline)
-    return Response(df.to_json(orient='records'), mimetype='application/json')
+    return Response(df.to_json(orient='table'), mimetype='application/json')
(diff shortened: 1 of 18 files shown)


src/lib/dataset/pandas.ts
```diff
@@ -1,115 +1,71 @@
...
+export function pandasDataTableToDataset(pandasTable: PandasDataTable): DataSet {
+  const dataset: DataSet = { headers: [], data: [] };
+  dataset.headers = pandasTable.schema.fields.map(({ name, type }) => ({
+    name,
+    type: tableSchemaTypeToDataSetColumnType(type),
+  }));
+  for (const row of pandasTable.data) {
+    dataset.data.push(
+      dataset.headers.map(column =>
+        Object.prototype.hasOwnProperty.call(row, column.name) ? row[column.name] : null,
```
fspot, line 66: C'est possible que pandas ne renvoie pas toutes les colonnes pour chaque ligne de donnée ?

davinov: It's true that mongo definitely can, but pandas shouldn't.
I can remove this condition as it's not necessary

fspot: Ok I just wanted to be sure I didn't miss something :+1: We can keep it

server/playground.py
```diff
@@ -16,13 +17,21 @@
...
+    return response
+
 
 @app.route('/', methods=['GET', 'POST'])
 def handle_request():
     if request.method == 'GET':
         return json.dumps(get_available_domains())
     elif request.method == 'POST':
-        return execute_pipeline(request.get_json())
+        return execute_pipeline(request.get_json(), **request.args)
```
fspot, line 34: I the dict > json conversion is automatic, we should be able to get rid of the `json.dumps` above, too ?

fspot (approved):
Very clear :+1:

server/playground.py
davinov: ah yeah true! Flask handles it for us

davinov: cb1f366

davinov: Arf, actually no: `
TypeError: The view function did not return a valid response. The return type must be a string, dict, tuple, Response instance, or WSGI callable, but it was a list.
`

because this methods returns a list, not a dict

PR #681, opened 2020-10-29
feat(playground): load all csv files in playground/datastore

davinov:
Put any csv file in playground/datastore, and it will be loaded in panda's server.

### DIFF
--- a/server/playground.py
+++ b/server/playground.py
@@ -8,7 +8,8 @@ POST: execute the pipeline in the body of the request and returns the transforme
 Run it with `FLASK_APP=playground FLASK_ENV=development flask run`
 """
 import json
-from typing import List
+from glob import glob
+from os.path import basename, splitext
 
 import pandas as pd
 from flask import Flask, Response, request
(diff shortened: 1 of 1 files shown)


server/playground.py
```diff
@@ -34,8 +35,9 @@ def handle_request():
         return execute_pipeline(request.get_json(), **request.args)
 
 
-# Load domains in memory
-DOMAINS = {'sales': pd.read_csv('../playground/datastore/sales.csv')}
+# Load all csv in playground's datastore
+csv_files = glob('../playground/datastore/*.csv')
+DOMAINS = {splitext(basename('sales'))[0]: pd.read_csv(csv_file) for csv_file in csv_files}
```
fspot, line 40:
```suggestion
DOMAINS = {splitext(basename(csv_file))[0]: pd.read_csv(csv_file) for csv_file in csv_files}
```

davinov: ahah oupsie, nicely spotted

fspot (approved):

PR #685, opened 2020-11-02
build(server): sanitize branch name for image tagging

davinov:
Some PRs are failing because branches named with `/` cannot be used as docker image tags.


alice-sevin (approved):
LGTM! Merci pour ce fix :pray:

PR #691, opened 2020-11-02
feat(server): add convert step

fspot:
:warning: pandas cast can slightly differ from mongo's cast. For example, any string evaluates to `true` in mongo, whereas in pandas an empty string evaluates to `false`. I chose to keep pandas's cast for simplicity / performance (and because those are edge cases).

### DIFF
--- a/server/weaverbird/steps/convert.py
+++ b/server/weaverbird/steps/convert.py
@@ -20,9 +20,14 @@ def cast_to_int(s: Series) -> Series:
     return s_integer.append(s_nan).sort_index()
 
 
+def cast_to_str(s: Series) -> Series:
+    return s.astype(str)
+
+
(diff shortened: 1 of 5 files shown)


server/weaverbird/steps/convert.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def cast_to_datetime(s: Series) -> Series:
+    return to_datetime(s, errors='coerce')  # cast errors will result in NaT values
+
+
+def cast_to_bool(s: Series) -> Series:
+    return s.astype(bool)
+
+
```
luc-leonard, line 34: Super clean de bosser directement sur les Series :+1:

luc-leonard (approved):

PR #703, opened 2020-11-04
Pandas ifthenelse step

luc-leonard:


server/weaverbird/steps/ifthenelse.py
```diff
@@ -0,0 +1,32 @@
...
+    ) -> DataFrame:
+        # first, I make a series with the True / False result of the condition
+
+        return df.assign(
+            **{
+                self.new_column: numpy.where(
+                    self.condition.filter(df), df.eval(self.then), df.eval(self.else_value)
+                )
+            }
+        )
```
luc-leonard, line 32: This computes the then and else formula for the whole dataframe. It would be faster to just eval the good side of the if/else for each row, but is it doable with pandas ?

vdestraitt: I think that numpy will apply each evaluated expression only to its related filtered vector so it's just fine in my opinion

luc-leonard: I tried to apply each eval only to a sub dataframe. Results were not conclusive. cf. https://gist.github.com/luc-leonard/74963aaa2886443bfdd60595e6d8d805

server/weaverbird/steps/ifthenelse.py
```diff
@@ -0,0 +1,32 @@
...
+        df: DataFrame,
+        domain_retriever: DomainRetriever = None,
+        execute_pipeline: PipelineExecutor = None,
+    ) -> DataFrame:
+        # first, I make a series with the True / False result of the condition
+
+        return df.assign(
+            **{
+                self.new_column: numpy.where(
+                    self.condition.filter(df), df.eval(self.then), df.eval(self.else_value)
```
fspot, line 29: you may want to use `clean_formula` from https://github.com/ToucanToco/weaverbird/pull/702 to handle mongo's syntax for columns containing spaces?

server/weaverbird/steps/ifthenelse.py
```diff
@@ -0,0 +1,32 @@
...
+    else_value: Any = Field(alias='else')
+
+    def execute(
+        self,
+        df: DataFrame,
+        domain_retriever: DomainRetriever = None,
+        execute_pipeline: PipelineExecutor = None,
+    ) -> DataFrame:
+        # first, I make a series with the True / False result of the condition
+
```
fspot, line 25:
```suggestion
```

fspot (changes requested):
some remarks above

server/weaverbird/steps/ifthenelse.py
luc-leonard: :+1:

luc-leonard:
[VQB pandas: ifthenelse step](https://trello.com/c/DClnjyXM/1187-vqb-pandas-ifthenelse-step)

fspot (approved):

PR #718, opened 2020-11-05
fix(server): handle "reference_column" of cumsum step

fspot:


server/weaverbird/steps/cumsum.py
```diff
@@ -19,6 +19,7 @@ class Config:
         allow_population_by_field_name = True
 
     def execute(self, df: DataFrame, domain_retriever=None, execute_pipeline=None) -> DataFrame:
+        df = df.sort_values(self.groupby or [] + [self.reference_column])  # type: ignore
```
vdestraitt, line 22: I'm not sure of the overhead of sorting on several columns rather than on 1 column, but we can get the same result by sorting on `reference_column` only. A sort will run on the groupby columns by default. 

Cf this piece of doc found [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html) about the `sort` parameter of `groupby`:

> sort bool, default True: Sort group keys. Get better performance by turning this off. **Note this does not influence the order of observations within each group. Groupby preserves the order of rows within each group.**

(In terms of pure performance we could decide to set this `sort` param to `false` to avoid sorting groups at the end, but I think we prefer to favour usability - it's easier to read a sorted dataset - over speed at this stage)

vdestraitt (commented):
Thank you ! Just a small comment/suggestion. A bit too picky maybe...

server/weaverbird/steps/cumsum.py
fspot: 40dba0b

vdestraitt (approved):
LGTM thank you!

PR #725, opened 2020-11-07
build(docs): switch to github pages build

davinov:
Docs build is broken for a while.
@zegonz Do you remember why it as on CircleCI and not built by GitHub pages in the first place ?
I think at the time, some plugins you were using may not be supported by GitHub pages, but they all seem now.

I've tested it in a sample repo here: https://david.nowinsky.net/weaverbird-docs-test/ https://github.com/davinov/weaverbird-docs-test and it seems fine.


zegonz (approved):
As you said if I remember well there was some issues with some dependencies but if now it's working well we can use github pages to do this, it's  way more standard 🔥 

Just a question : is the local build still work as well?

davinov:
> As you said if I remember well there was some issues with some dependencies but if now it's working well we can use github pages to do this, it's way more standard fire
> 
> Just a question : is the local build still work as well?

Yes ! I adapted the Dockerfile accordingly

PR #729, opened 2020-11-09
fix(server): playground :: display error messages when using pandas

fspot:
Before : 
image

After : 
image


server/playground.py
```diff
@@ -30,11 +30,15 @@ def after_request(response):
 @app.route('/', methods=['GET', 'POST'])
 def handle_request():
     if request.method == 'GET':
-        return Response(json.dumps(get_available_domains()), mimetype='application/json')
+        return jsonify(get_available_domains())
```
davinov, line 33: Oh way better, thanks ^^

davinov (approved):

PR #731, opened 2020-11-09
fix(server): fix percentage step

fspot:
We had a weird error when using a pivot step : 

```
  File "/home/fred/tctc/weaverbird/server/weaverbird/steps/percentage.py", line 20, in default_ts_created
    return v or f'{values["column"]}_PCT'
KeyError: 'column'
```

this error is solved by using a fallback for `new_column_name` without a validator, the way I did in some other steps


vdestraitt (approved):
Thanks for the fix!

PR #738, opened 2020-11-10
Pandas ifthenelse step: else if

luc-leonard:
handled the else if... case. I am a little bit disapointed with my solution that I find not very elegant. The model is almost repeated. We have a weird isinstance(self.else_value, dict).


luc-leonard:
[(2) VQB ifthenelse: il faut qu'on gère les condition imbriquées (avec des `else if`)](https://trello.com/c/GekgyRvE/1280-2-vqb-ifthenelse-il-faut-quon-g%C3%A8re-les-condition-imbriqu%C3%A9es-avec-des-else-if)

fspot (approved):

PR #749, opened 2020-11-17
Substring: parameter badly named 'new_column_name' is now aliased to newColumnName

luc-leonard:


vdestraitt (approved):
Thank you !

davinov (approved):

davinov:
Rebasing for the poetry fix in GitHub actions:

fspot:
@luc-leonard I would add
```
    class Config:
        allow_population_by_field_name = True
```
so the step can be instanciated both with the alias name or the regular name

PR #753, opened 2020-11-19
Upgrade from node-sass to dart-sass and other build upgrades

davinov:
As libsass, and therefore node-sass, get deprecated, this ensures the build will continue to work on recent node versions.

During this upgrade, I have to change how StepForms are registred or use in QueryBuilder, as the decorator StepFormComponent (which was doing a bit of dark bit magic), does not get the template/render anymore from the vue file loader.


austil (approved):
Looks good to me.

Juste part clartée : 
- la première partie du refactor est due au passage à dart-sass
- la deuxième, est due à la mise à jour de rollup-plugin-vue qui passe lui aussi à dart-sass en plus d'une mise à jour sous-jacente de @vue/component-compiler qui change certains comportements sans trop s'en vanter dans les changelogs

PR #754, opened 2020-11-20
feat(variables): allow to flatmap array variable when contained in an other array

alice-sevin:
For now when we use an array variable:
```[ <%= my_awesome_array %>, <%= my_awesome_string %> ]```
in an array (for step filer, $in, $nin for example ), we got:
```[ [ 'first_value',  'second_value'], 'other_value' ]```
while we want to flat this variable array:
```[ 'first_value',  'second_value', 'other_value' ]```

We need to flatmap array variables only. See exampleInterpolateFunc.
The logic will be reproduce in any projects needed

### DIFF
--- a/src/lib/templating.ts
+++ b/src/lib/templating.ts
@@ -509,7 +509,7 @@ const aloneVarsRegExp = new RegExp('^' + (_.templateSettings.interpolate as RegE
  *
  * It's based on lodash templates (which looks like embedded Ruby, but executes JavaScript),
  * with two important twists designed to avoid type-induced headaches to final users:
- * 1. If a variable is provided "alone", i.e. with no surrounding characters, then it returns the variable value untouched and not coerced to a string. 
+ * 1. If a variable is provided "alone", i.e. with no surrounding characters, then it returns the variable value untouched and not coerced to a string.
  * Examples:
  * - `exampleInterpolateFunc('<%= count %>', { count: 42 })` returns `42`
  * - `exampleInterpolateFunc('<%= 2 > 1 %>', {})` returns `true`
--- a/src/lib/templating.ts
+++ b/src/lib/templating.ts
@@ -11,7 +11,7 @@ export type ScopeContext = {
   [varname: string]: any;
 };
 
-export type InterpolateFunction = (template: string, context: ScopeContext) => any;
+export type InterpolateFunction = (template: string | any[], context: ScopeContext) => any;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by alice-sevin** [comment]
[ETQD, je peux utiliser une variable de type array comme valeur d'un $in/$nin step](https://trello.com/c/37odKqPh/1364-etqd-je-peux-utiliser-une-variable-de-type-array-comme-valeur-dun-in-nin-step)

File: src/lib/templating.ts [hunk 1]
```diff
@@ -517,14 +517,28 @@ const aloneVarsRegExp = new RegExp('^' + (_.templateSettings.interpolate as RegE
...
+  value: string | any[],
   context: ScopeContext,
 ) {
   if (typeof value === 'string' && value.match(aloneVarsRegExp)) {
     return Function(
       'context',
       `with(context) { return ${(value.match(aloneVarsRegExp) as string[])[1]} }`,
     )(context);
+  } else if (Array.isArray(value)) {
+    // TODO: make doc
```
davinov, line 529: Here is my proposal for this doc:
```
// Expand variables that are also arrays into their parent array
```

And for the main docstring of the function:
```
/**
 * A simple interpolate function, provided as an example.
 *
 * It's based on lodash templates (which looks like embedded Ruby, but executes JavaScript),
 * with two important twists designed to avoid type-induced headaches to final users:
 * 1. If a variable is provided "alone", i.e. with no surrounding characters, then it returns the variable value untouched and not coerced to a string. 
 * Examples:
 * - `exampleInterpolateFunc('<%= count %>', { count: 42 })` returns `42`
 * - `exampleInterpolateFunc('<%= 2 > 1 %>', {})` returns `true`
 * - `exampleInterpolateFunc('There is <%= count %> bananas', { count: 42 })` returns `'There is 42 bananas'`
 *
 * 2. Inside arrays, variables that are also arrays are expanded
 * Examples:
 * - `exampleInterpolateFunc(['<%= roles %>'], { group: ['crewmate', 'impostor'] })` returns `['crewmate', 'impostor']` (and not `[['crewmate', 'impostor']]`)
 */
```

davinov (commented):
:+1: 
Here is a proposal for the comments:

davinov (approved):
:clap: 

@alice-sevin Before merging, would you mind adding a entry to the changelog ? This seems quite a change, so it'll be nice to document it

davinov (approved):

PR #758, opened 2020-11-23
docs: missing slash is breaking a link on project website nav

austil:
The "UI components" is currently unreachable from the nav of the project website.
Probably the most interesting page of this website for someone who want to start using this project/contributing to it :grimacing: 

image


davinov (approved):
:scream: nice catch

PR #759, opened 2020-11-24
all steps, and everything in it can be populated by field name

luc-leonard:
Careful of ifthenelse and its multiple inheritance.
It should not be an issue, especially since the alias is only used because 'else' is a reserved keyword. we do not really care about the field name here.


austil:
[(5) L'UI des prepared datasets pointe vers les datasets préparés pandas (et non plus mongo) (feature flag jusqu'à release)](https://trello.com/c/oDTbKSjt/1333-5-lui-des-prepared-datasets-pointe-vers-les-datasets-pr%C3%A9par%C3%A9s-pandas-et-non-plus-mongo-feature-flag-jusqu%C3%A0-release)

fspot (approved):

PR #764, opened 2020-11-25
fix(front): updateDataset can reset the preview to an empty state when pipeline is empty

davinov:
This will avoid the horrible "pipeline is empty" when instantiating the component.

### DIFF
--- a/src/lib/backend.ts
+++ b/src/lib/backend.ts
@@ -37,11 +37,21 @@ export interface BackendService {
  * While no backend service is set, this will throw errors if we try to call its methods
  */
 export const UnsetBackendService: BackendService = {
-  listCollections() {
+  listCollections(): BackendResponse<string[]> {
+    /* istanbul ignore next */
+    if (process.env.NODE_ENV === 'test') {
+      // Act like a mock during tests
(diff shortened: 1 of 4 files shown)


austil:
> On some cases, dispatching the "updateDataset" action from outside
> weaverbird's store breaks the vuex module context in the decorator.
> This was easily fixed by using regular calls instead of the decorator.

Ces décorateurs ils sont de nous ou ça vient d'une dépendance ?

davinov:
> > On some cases, dispatching the "updateDataset" action from outside
> > weaverbird's store breaks the vuex module context in the decorator.
> > This was easily fixed by using regular calls instead of the decorator.
> 
> hushed Ces décorateurs ils sont de nous ou ça vient d'une dépendance ?

@austil They were added by us (see actions.ts). I had a hard time figuring out this problem: in some cases, the vuex `context` passed through the decorator was not the same that the one from the original function call. In that case, the decorator was receiving the root store, and not the VQB module one. This lead to an error stating that the mutation does not exist in this context.

austil (approved):
Thx for all those fixes.

PR #768, opened 2020-11-27
build(front): expose types

davinov:
This was an effort started by @alice-sevin.
It's not as simple as it may seem, as the build tools are not very nicely integrated together.
It required upgrading to rollup-plugin-typescript2, which is way slower than the original but seems to be the only way of getting types declarations exports, and then using Microsoft'API extractor to bundle them together.
In the process, I upgraded rollup and its plugins. This revealed many (really many) ts diagnostics were previously ignored. It will be an on-going effort to solve them all, so I just put them as warnings instead of errors (hopefully this won't stay long).


api-extractor.json
```diff
@@ -0,0 +1,364 @@
+/**
```
davinov, line 1: This was generated by api-extractor init, no need to review it all

api-extractor.json
```diff
@@ -0,0 +1,364 @@
...
+  },
+
+  /**
+   * Configures how the .d.ts rollup file will be generated.
+   */
+  "dtsRollup": {
+    /**
+     * (REQUIRED) Whether to generate the .d.ts rollup file.
+     */
+    "enabled": true,
```
davinov, line 184: This is the only enabled option (the dstRollup)

api-extractor.json
```diff
@@ -0,0 +1,364 @@
...
+   * analyzes the symbols exported by this module.
+   *
+   * The file extension must be ".d.ts" and not ".ts".
+   *
+   * The path is resolved relative to the folder of the config file that contains the setting; to change this,
+   * prepend a folder token such as "<projectFolder>".
+   *
+   * SUPPORTED TOKENS: <projectFolder>, <packageName>, <unscopedPackageName>
+   */
+  "mainEntryPointFilePath": ".tmp/types/src/main.d.ts", // The rollup bundling will output types the .tmp/types folder, according to tsconfig.json
```
davinov, line 48: And the is the interesting part: it points to where the typescript compiler leaves the .d.ts files

PrettyWood:
A **HUGE** thanks for looking at this @davinov! I took couple of hours on an evening to try to do it but stopped as I got many errors afterwards and it was way harder than it seemed. Very well done 👏 👏 👏

davinov:
@PrettyWood it's not finished unfortunately, as it break some tests for unknown reasons (the code has'nt changed a bit !)
I'll continue looking into it, as long as I ave some patience for these js build tools :p

davinov:
And actually it's not even enough, as aliases (@) are not resolved in the declaration files
:rage:

davinov:
There is no need to keep that PR opened for ages
We'll reopen it later when we will have found a better solution

PR #771, opened 2020-12-01
fix(fillna-step): use undefined as column default and add error when columns length is < 1

alice-sevin:
In previous PR: https://github.com/ToucanToco/weaverbird/pull/648,
We setted the column default value in filterStep equal to 
```
['']
```
So it displayed the value in multiselect as a selected column (because we considered that column was always a string so we included it into an array ).
We need to use undefined as default

Also added a minItems validation to schema and tests to validate that columns has at least one item selected

Before:
before

After:
after


austil (approved):
Déjà vu :upside_down_face:

PR #784, opened 2020-12-03
feat(server): raise more helpful error on pipeline failure

davinov:
To help debugging failing pipelines, I suggest to raise more helpful errors.
On failure, we now indicate:
- the index of the failing step,
- the name of the failing step,
- and the original exception message.

Later, we could improve the exceptions messages inside the steps, on a step-by-step basis.


davinov:
Btw, it's notmal that we don't ping featurepeek for the storybook, because it's a pure-backend PR (so the storybook build is skipped)

austil (approved):
Déjà culte !

PR #785, opened 2020-12-07
feat(pandas log memory usage): each step of the pipeline logs memory usage

luc-leonard:


server/weaverbird/utils/size.py
```diff
@@ -0,0 +1,11 @@
+import math
+
+
+def convert_size(size_bytes):
```
davinov, line 4:
```suggestion
def convert_size(size_bytes):
   """
   Format a byte count in a human-legible format, using S.I. suffixes, and keeping only a few significant digits. 
   """
```

server/weaverbird/pipeline_executor.py
```diff
@@ -26,12 +30,17 @@ def execute_pipeline(self, pipeline_steps: List[dict]) -> DataFrame:
...
         steps = Pipeline(steps=pipeline_steps).steps
         df = None
+        stopwatch = StopWatch()
         for index, step in enumerate(steps):
             try:
-                df = step.execute(
-                    df,
-                    domain_retriever=self.retrieve_domain,
-                    execute_pipeline=self.execute_pipeline,
+                with stopwatch:
```
davinov, line 36: Nice usage as context manager :+1:

davinov (approved):
Nice, thanks :+1: 
Here is a few optional suggestions:

server/weaverbird/pipeline_executor.py
```diff
@@ -26,12 +30,17 @@ def execute_pipeline(self, pipeline_steps: List[dict]) -> DataFrame:
...
+                    df = step.execute(
+                        df,
+                        domain_retriever=self.retrieve_domain,
+                        execute_pipeline=self.execute_pipeline,
+                    )
+                logger.info(
+                    f'step {step} used {convert_size(df.memory_usage().sum())}, took {stopwatch.interval}s to execute'
```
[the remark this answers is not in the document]
davinov: Also, do you think it's useful to test this ? (like, logger.info should be called as many times as there is steps)
It's difficult to know if logging is a feature or not

davinov, line 43: This makes me think of something else :
to be sure that the memory usage stays close to the memory of the current dataframes, maybe it'll be nice to ensure previous dataframes are freed from the memory.

So maybe we should call the garbage collector explicitly after each step?
https://stackoverflow.com/questions/1316767/how-can-i-explicitly-free-memory-in-python

PR #791, opened 2020-12-14
Mongo 4.2 translator

davinov:
Adds a new translator for Mongo 4.2.
Inherits from the Mongo 4.0 with a twist: regexes are now supported in conditions, thanks to the new [$regexMatch](https://docs.mongodb.com/manual/reference/operator/aggregation/regexMatch/) aggregation operator.


davinov:
I've investigated the coverage issue: only a branch of the `class ... extends ...` seems to be missing.
This is probably due to TS polyfills added by target: es5. However, changing this to es6 for tests (see 540c5be) triggers a LOT of errors, that I wasn't able to correct quickly.
Therefore I suggest we abandon this small bit of coverage for now, and I'll create an issue to switch the TS target to ES6 later.

tests/unit/mongo.spec.ts
```diff
@@ -3657,6 +3657,32 @@ describe('Pipeline to mongo translator', () => {
...
+          name: 'ifthenelse',
+          newColumn: 'NEW_COL',
+          if: { column: 'TEST_COL', operator: 'notmatches', value: '^a' },
+          then: '"True"',
+          else: '"False"',
+        },
+      ]),
+    ).toThrow('Unsupported operator');
+  });
+
```
austil, line 3685: Est ce que ce test ne devrait pas aussi être fait pour mongo 4.0 ?

austil (approved):
All good.

J'ai seulement une question mais elle est peut être pas très pertinente comme les translators mongo ont l'air d'hériter les uns des autres, de la version la plus ancienne à la plus récente.

tests/unit/mongo.spec.ts
davinov: Indeed, I've been a bit lazy and wanted to avoid copy/pasting these tests, as Mongo40 inherits directly from Mongo36.
Di you think we should copy this test to mongo 4.0 test suite?

austil: I doesn't cost much but also doesn't provide a lot of value. 
The best setup would be parameterized tests for common translator behaviors but if this does not exist we'll do without.

davinov: It's an excellent idea to parametrize tests with conditional version support. And it's also really easy ^^
I've done it in 65294bd ; what do you think?

tests/unit/mongo.spec.ts
```diff
@@ -3657,31 +3649,85 @@ describe('Pipeline to mongo translator', () => {
     ]);
   });
 
-  it('should fail if regexes are used in conditions', () => {
-    expect(() =>
-      mongo36translator.translate([
+  if (version <= '40') {
+    it('should fail if regexes are used in conditions', () => {
```
austil, line 3653: :+1:

PR #793, opened 2020-12-14
feat(front): add a tooltip for each column name

davinov:
The tooltip is selectable for easy copy/paste

Screenshot from 2020-12-14 18-37-50
Screenshot from 2020-12-14 18-37-42


PrettyWood (approved):
Ahhhh trop cooool 😎

PR #798, opened 2020-12-18
fix pandas total step: missing results

luc-leonard:
fix(pandas/total): some combination of totals were missing.


davinov (approved):
Neat :tada: thx!

PR #799, opened 2020-12-18
fix(front): preview request only once when changing domain

davinov:
The `selectCurrentDomain` and `selectStep` both request an update of the preview.
To avoid asking twice the work to the server, keep only the first one.


austil (approved):
Well, that was easy !

PR #800, opened 2020-12-18
fix(mongo): fix usage of $regexMatch

davinov:
This is what happen when you skip reading the docs before using an operator :stuck_out_tongue: 
I've tested this against my local mongo (4.4) to be sure it works (should be the same with 4.2).


austil (approved):
Also checked MongoDB doc, just to be sure :stuck_out_tongue:

PR #807, opened 2020-12-30
fix(if then else widget): remove z-index context on dropdowns

austil:
Fix a css bug that was preventing multiselect dropdowns inside the ifThenElse Widget to use their expected z-index because they were wrapped into another z-index context.

I insist that this issue was only affecting dropdowns inside the ifThenElse Widget, not any of the other one like the filter step widget for example.

Before :
vqb_ifelse_dropdown_bug-before

After :  
vqb_ifelse_dropdown_bug-after

### DIFF
--- a/src/components/stepforms/widgets/IfThenElseWidget.vue
+++ b/src/components/stepforms/widgets/IfThenElseWidget.vue
@@ -274,7 +274,6 @@ export default class IfThenElseWidget extends Vue {
   ::v-deep .condition-row {
     background-color: #e9eff5;
   }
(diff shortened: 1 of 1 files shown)


austil:
[(3) Conditions Options dropdowns does not go above others when opened](https://trello.com/c/89gsETGW/1391-3-conditions-options-dropdowns-does-not-go-above-others-when-opened)

alice-sevin (approved):
LGTM! Thanks pour ce dev! Il faudra juste qu'on relance la pipeline si elle foire toujours.

PR #811, opened 2021-01-07
fix(getters): bound active step index to pipeline size

austil:
This PR prevent the computed active step index to end up out of bound (meaning above the pipeline length).

Side note : I'm not a fan of having a specific `selectedStepIndex` value associated with a specific meaning, here `-1 <=> pipeline.length - 1`. It seems better to me to have `selectedStepIndex` explicitly set to `pipeline.length - 1` when necessary i.e. when loading/selecting pipelines. Fortunately it's not this big of a deal :wink:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+
+- Bound active step to pipeline size to avoid ending up with no step selected in some specific cases
+
 ## [0.34.1] - 2020-12-18
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


austil:
[(2) ETQC, je ne peux pas mettre le Weaverbird dans un état où aucune step n'est selectionnée](https://trello.com/c/yT7io5QE/1552-2-etqc-je-ne-peux-pas-mettre-le-weaverbird-dans-un-%C3%A9tat-o%C3%B9-aucune-step-nest-selectionn%C3%A9e)

vdestraitt (approved):
Thank you for this fix!

PR #817, opened 2021-01-13
fix(server): fix util rename_duplicated_columns

fspot:
This util had issues with input like:

```foo, foo, foo_1```

which gave:

```foo, foo_1, foo_1```
(which also contains duplicates :/)

and now gives:

```foo, foo_2, foo_1```
:ok_hand: 

**EDIT** I removed the caching of poetry's virtualenv because it caused issues :shrug: It adds maybe 1 minute duration in the python CI, but it's still the fastest of all our github checks, so imo it's not a problem (and also I'm lazy to find how to fix it properly).


davinov (approved):
👍

PR #818, opened 2021-01-15
fix(mongo): "not in" (`nin`) in if/then/else conditions

davinov:
$nin doesn't exist in https://docs.mongodb.com/manual/meta/aggregation-quick-reference/#array-expression-operators


alice-sevin (approved):
LGTM thanks for this fix :pray:

PR #822, opened 2021-01-18
feat(server): add movingaverage step

fspot:
The last missing step :v:


davinov (approved):
Crystal clear :gem: thx!

PR #823, opened 2021-01-18
poc: pipeline with variables

fspot:


server/tests/test_pipeline.py
```diff
@@ -0,0 +1,31 @@
...
+        FilterStepWithVariables(
+            name='filter', condition={'column': 'foo', 'operator': 'lt', 'value': '{{ foo }}'}
+        ),
+    ]
+
+    pipeline = pipeline_with_variables.render({'foo': 42})
+    expected_result = Pipeline(
+        steps=[
+            TopStep(name='top', rank_on='price', sort='desc', limit=42),
+            FilterStep(name='filter', condition={'column': 'foo', 'operator': 'lt', 'value': 42}),
```
fspot, line 28: Problème ici : la `value` vaut `"42"` au lieu de `42` car le type de `value` est Any dans la classe ComparisonCondition, donc il n'y a pas de cast en `int` effectué par pydantic lors de l'instanciation du modèle.

(il n'y a pas ce problème sur la `limit` de la TopStep car le champ est bien de type int)

fspot: Problème résolu avec l'util `nosql_apply_parameters_to_query` de toucan-connectors (+ https://github.com/ToucanToco/toucan-connectors/pull/299 pour gérer les dates) :muscle:

server/pyproject.toml
```diff
@@ -20,6 +20,7 @@ isort = "^5.6.4"
 mypy = "^0.790"
 Flask = "^1.1.2"
 pytest-mock = "^3.3.1"
+toucan-connectors = ">0"
```
luc-leonard, line 23: so we have the nosql_apply_parameters_to_query

server/weaverbird/steps/totals.py
```diff
@@ -19,7 +26,7 @@ class Config(PopulatedWithFieldnames):
 class TotalsStep(BaseStep):
     name = Field('totals', const=True)
     total_dimensions: List[TotalDimension] = Field(alias='totalDimensions')
-    aggregations: List[Aggregation] = Field(min_items=1)
+    aggregations: Sequence[Aggregation]
```
luc-leonard, line 29: @fspot  I needed to replace List by Sequence because of a weird error about covariance. However, we cannot have a min_items constraint with a Sequence. I could add it as a root_validator, is it an acceptable workaround ?

server/tests/test_pipeline.py
```diff
@@ -0,0 +1,32 @@
...
+from toucan_connectors.common import nosql_apply_parameters_to_query
+
+from weaverbird.pipeline import Pipeline, PipelineWithVariables
+
+
+def get_test_cases():
+    test_cases = []
+    globs = glob.glob('./tests/fixtures/fixtures_templating/*.json')
+    for file in globs:
+        print(file)
```
fspot, line 14:
```suggestion
```

server/tests/test_pipeline.py
```diff
@@ -0,0 +1,32 @@
...
+            for test in file_content:
+                test_cases.append(test)
+    return test_cases
+
+
+@pytest.mark.parametrize('data, context, expected_result', get_test_cases())
+def test_aggregate_step_with_variables(data, context, expected_result):
+    print(data)
+    print(context)
+    print(expected_result)
```
fspot, line 26:
```suggestion
```

server/weaverbird/types.py
```diff
@@ -1,11 +1,12 @@
...
+from typing import Any, Callable
 
 from pandas import DataFrame
 
 ColumnName = str
-Pipeline = List[dict]
+# Pipeline = List[dict]
 DomainRetriever = Callable[[str], DataFrame]
-PipelineExecutor = Callable[[List[dict]], DataFrame]
+PipelineExecutor = Callable[[Any], DataFrame]
```
luc-leonard, line 8: Maybe i should let PipelineExecutor#execute_pipeline take a list of dict, instead of a pipeline to avoid circular imports ?

server/weaverbird/pipeline_executor.py
```diff
@@ -22,13 +21,13 @@ class PipelineExecutor:
     def __init__(self, domain_retriever: DomainRetriever):
         self.retrieve_domain = domain_retriever
 
-    def execute_pipeline(self, pipeline_steps: List[dict]) -> DataFrame:
+    def execute_pipeline(self, pipeline: Pipeline) -> DataFrame:
```
luc-leonard, line 24: @fspot I think it's cleaner to pass a Pipeline and not a List[dict]. However, it leads to some circular imports. Is it worth it ?

server/weaverbird/steps/combination.py
```diff
@@ -15,7 +16,9 @@ def resolve_pipeline_for_combination(
     """
     Combined pipelines can be either single domains (str), or complete pipeline (list of steps)
     """
+    from weaverbird.pipeline import Pipeline
```
luc-leonard, line 19: to avoid circular imports :/

server/weaverbird/pipeline_executor.py
davinov: I think you can also use the string `'Pipeline'`. See [Forward references](https://www.python.org/dev/peps/pep-0484/#forward-references) section of the PEP 484 on type hints

davinov: It seems quite logical than PipelineExecutor imports Pipeline.
Maybe the other way around isn't ^^ So you know what causes the import of PipelineExecutor by Pipeline? If it's easy, we can correct it.

luc-leonard:
[(5) [BACK] sauvegarde de pipeline weaverbird qui contiennent des champs templatisables](https://trello.com/c/P6MVYRnG/1636-5-back-sauvegarde-de-pipeline-weaverbird-qui-contiennent-des-champs-templatisables)

server/tests/test_pipeline.py
```diff
@@ -0,0 +1,28 @@
...
+    globs = glob.glob('./tests/fixtures/fixtures_templating/*.json')
+    for file in globs:
+        with open(file) as json_file:
+            file_content = json.load(json_file)
+            for test in file_content:
+                test_cases.append(test)
+    return test_cases
+
+
+@pytest.mark.parametrize('data, context, expected_result', get_test_cases())
```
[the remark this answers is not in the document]
davinov: Detail: I don't know if there is a simple way of adding a different id per test - maybe the fixture file name ?- so it's easy to know which one failed.

davinov (approved):
Awesome! Thanks for putting that together

server/weaverbird/pipeline_executor.py
luc-leonard: Pipeline imports steps
some steps (append, join mostly) use the pipeline executor
I think that we can workaround this by importing the PipelineExecutor for those step when they are called.

davinov: Awww true
In that case, maybe we should import locally the pipeline executor in the execute method of these steps? That would break the circular import

server/tests/test_pipeline.py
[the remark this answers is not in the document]
luc-leonard: :+1:

PR #828, opened 2021-02-04
feat(front): add text and center loading spinner

davinov:
Also prevent to have "No data available" below the spinner while loading

Screenshot from 2021-02-04 13-23-55


austil (approved):
Go on ! :green_circle:

PR #833, opened 2021-02-11
fix(group by null): we no longer drop null keys when grouping values by keys

luc-leonard:


fspot (approved):
Arf bien vu :+1:

davinov (approved):
LGTM!

PR #839, opened 2021-02-17
feat(front): order of combinable tables in selector

davinov:
We want to display:
- first pipelines
- then domains.
Each category must be sorted alphabetically


austil (approved):
GG IZI :checkered_flag:

austil:
Should we add this to the changelog ?

davinov:
> Should we add this to the changelog ?

Yep! I've amend the commit

PR #840, opened 2021-02-19
feat(pandas): add a new aggregation to count, including null

luc-leonard:


server/weaverbird/steps/aggregate.py
```diff
@@ -8,9 +8,11 @@
 from weaverbird.steps.base import BaseStep
 from weaverbird.types import ColumnName, PopulatedWithFieldnames, TemplatedVariable
 
-AggregateFn = Literal['avg', 'sum', 'min', 'max', 'count', 'count distinct', 'first', 'last']
+AggregateFn = Literal[
+    'avg', 'sum', 'min', 'max', 'count', 'count distinct', 'first', 'last', 'count_even_null'
+]
 
-functions_aliases = {'avg': 'mean', 'count distinct': 'nunique'}
```
luc-leonard, line 13: Je ne suis pas sur du type de len ici.
Ni meme de si c'est une bonne idée de s'en servir comme tel. ca serait peut etre mieux de le wrap dans une fonction ?

server/weaverbird/steps/aggregate.py
```diff
@@ -8,9 +8,11 @@
 from weaverbird.steps.base import BaseStep
 from weaverbird.types import ColumnName, PopulatedWithFieldnames, TemplatedVariable
 
-AggregateFn = Literal['avg', 'sum', 'min', 'max', 'count', 'count distinct', 'first', 'last']
+AggregateFn = Literal[
+    'avg', 'sum', 'min', 'max', 'count', 'count distinct', 'first', 'last', 'count_even_null'
```
davinov, line 12: Would it be more coherent to name it: "count distinct including empty" ?
So it shows it's close to "count distinct", and provide a friendlier label

server/weaverbird/steps/aggregate.py
```diff
@@ -8,9 +8,11 @@
...
+]
 
-functions_aliases = {'avg': 'mean', 'count distinct': 'nunique'}
+functions_aliases = {'avg': 'mean', 'count distinct': 'nunique', 'count_even_null': len}
```
davinov, line 15: Is there a difference between using `len` and using `nuunique(drop_na=False)`?
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.nunique.html

davinov (commented):
:ok_hand: Thanks!
Here a 2 questions:

server/weaverbird/steps/aggregate.py
luc-leonard: I think it returns the same result, however, using len allow us to have a simpler code.

If we go this way, I think it would be cleaner to make it its own step than to modify the  aggregate step for this very specific behavior

server/weaverbird/steps/aggregate.py
luc-leonard: :+1:

server/weaverbird/steps/aggregate.py
davinov: Good, thanks for the answer

davinov (approved):
:+1: Thanks for all changes

PR #843, opened 2021-02-22
feat(unique-values): display checkboxes and label correctly

alice-sevin:
For now, we got sometimes an issue with checkboxes (not display correctly) and very long label, it's very weird in unique values.
We also add a title on hover to be able to see the label entirely if cropped.
These options are disabled on other checkboxes but could be reimplemented in some cases if needed

Before:
before

After:
after

### DIFF
--- a/src/components/ListUniqueValues.vue
+++ b/src/components/ListUniqueValues.vue
@@ -14,6 +14,7 @@
       >
         <CheckboxWidget
           :label="`${stringify(option.value)} (${option.count})`"
+          :croppedLabel="true"
           :value="isChecked(option)"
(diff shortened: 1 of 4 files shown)


alice-sevin:
[(2) Les checkboxes des valeurs uniques s'affichent bien dans le menu d'une colonne de type date](https://trello.com/c/nAjpA94T/1706-2-les-checkboxes-des-valeurs-uniques-saffichent-bien-dans-le-menu-dune-colonne-de-type-date)

davinov (approved):
Excellent, thanks!

PR #849, opened 2021-03-08
fix(rollup-step): use correct schema when updating step

alice-sevin:
Two errors were provided when editing a rollup step:
1. If user didn't provided a groupby value, field disappeared
2. If user leave the optionnal field empty, he had an error for string type missing in fields.

How fixes work?
1.We apply a default empty array to groupby value to avoid an error in line:
```
@input="setSelectedColumns({ column: editedStep.groupby[editedStep.groupby.length - 1] })"
```
The code look for an array that is 'null' while it should be empty if null 

2. We authorized null as value for optionnnal field in schema to avoid the string type error to be displayed

### DIFF
--- a/src/components/stepforms/schemas/rollup.ts
+++ b/src/components/stepforms/schemas/rollup.ts
@@ -35,7 +35,7 @@ export default {
       },
     },
     labelCol: {
-      type: 'string',
+      type: ['null', 'string'],
       minLength: 1,
       title: 'Label column name',
       description: 'The name of the label column that will be created',
@@ -44,7 +44,7 @@ export default {
       },
     },
     levelCol: {
-      type: 'string',
+      type: ['null', 'string'],
       minLength: 1,
       title: 'Level column name',
       description: 'The name of the level column that will be created',
@@ -53,7 +53,7 @@ export default {
       },
     },
     parentLabelCol: {
-      type: 'string',
+      type: ['null', 'string'],
       minLength: 1,
       title: 'Parent column name',
       description: 'The name of the parent column that will be created',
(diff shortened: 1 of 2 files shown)


alice-sevin:
[1- As a conceptor, when using the prepared dataset, i don't want my hierarchical rollup step to crash after saving](https://trello.com/c/bKFMpRbn/4612-1-as-a-conceptor-when-using-the-prepared-dataset-i-dont-want-my-hierarchical-rollup-step-to-crash-after-saving)

src/components/stepforms/schemas/rollup.ts
```diff
@@ -35,7 +35,7 @@ export default {
       },
     },
     labelCol: {
-      type: 'string',
+      type: ['null', 'string'],
```
davinov, line 38: @alice-sevin I'm wondering if sending `labelCol: null` will be something that Pydantic's will validate.
... I fear not :/ Maybe the problem is that the widget should set the value to undefined rather than to null if nothing is in the test input?

alice-sevin: I take a look :pray:

alice-sevin: Et merde, ça fixe pas le soucis quand on réupdate un champ :cry: , c'est galère, du coup à ce moment là on se retrouve à avoir "" et on a un minLength dans le schema. D'ailleurs je comprends pas ce minLength du coup on pourrait pas simplement le supprimer? Ca fait pas sens pour moi

davinov: Est-ce que le soucis ne viendrait pas du InputTextWidget?
Si, quand le champ de texte est vide, il émet `undefined` plutôt que `""` ou `null` alors on aurait bien un objet cohérent avec le schéma, non?

davinov (approved):
:+1: great, thanks!

PR #854, opened 2021-03-16
feat!(front): do not prepare pipeline outside backendService

davinov:
BREAKING CHANGE: methode executePipeline of backend service changes
signature and scope
Dereferencing pipelines/domains should be done inside this function
Same for variables interpolation

This enables custom behavior, handled by the host app or even the server
answering to the data request

### DIFF
--- a/playground/app.js
+++ b/playground/app.js
@@ -125,7 +125,7 @@ class MongoService {
     return response.json();
   }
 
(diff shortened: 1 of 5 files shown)


alice-sevin (approved):
:raised_hands: Une petite erreur de lint à corriger apparement et on sera bon pour attaquer côté front :pray:

PR #858, opened 2021-03-25
Better structured exceptions

luc-leonard:
feat(structured exceptions): the exceptions thrown now has a field 'details'. It is a dict that will be returned as-is up to the front end. it MUST contains only json serializable objects.

cf. https://github.com/ToucanToco/laputa/pull/4265

### DIFF
--- a/server/tests/test_pipeline_executor.py
+++ b/server/tests/test_pipeline_executor.py
@@ -146,3 +146,5 @@ def test_errors(pipeline_executor):
     assert 'delete' in exception_message
     assert 'columnThatDoesNotExist' in exception_message
     assert 'whatever' in exception_message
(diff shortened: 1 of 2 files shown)


davinov (changes requested):
Could you add a little test to showcase this behavior?
Thanks

luc-leonard:
[(3) ETQ C, lorsque j'execute un pipeline qui ne fonctionne pas (preview des prepared dataset), mon front obtient cette réponse](https://trello.com/c/gQD4eaVw/1819-3-etq-c-lorsque-jexecute-un-pipeline-qui-ne-fonctionne-pas-preview-des-prepared-dataset-mon-front-obtient-cette-r%C3%A9ponse)

fspot (approved):

PR #865, opened 2021-04-07
feat(back): no more None values in step.dict()

luc-leonard:
this pr excludes by default None values from .dict().
These None values did trigger some validation error on the front end

### DIFF
--- a/server/tests/test_pipeline.py
+++ b/server/tests/test_pipeline.py
@@ -7,6 +7,8 @@ from pydantic.main import BaseModel
 from toucan_connectors.common import nosql_apply_parameters_to_query
 
 from weaverbird.pipeline import Pipeline, PipelineWithVariables
+from weaverbird.steps import DomainStep, RollupStep
+from weaverbird.steps.aggregate import Aggregation
 
 
 class Case(BaseModel):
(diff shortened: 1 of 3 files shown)


luc-leonard:
[ETQC, je peux éditer une step rollup dans un prepared dataset préalablement enregistrée](https://trello.com/c/gzwrFBbA/1843-etqc-je-peux-%C3%A9diter-une-step-rollup-dans-un-prepared-dataset-pr%C3%A9alablement-enregistr%C3%A9e)

davinov (changes requested):
Thanks! That will solve a bunch of our problems :p

To make this code clearer and make sure noone deletes it afterwards, would you please:
- add a docstring in the dict explaining why we're excluding None here
- add a small test of some pipeline or some steps .dict() which will serve as examples and showcase the result (or find existing tests to enhance with this case)
- detail: try to format the name of your PR using  [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) (makes easier to search the PR list afterwards). For example, we usually use `front`, `server`, or `pandas` as scopes of commits/PRs

Thanks!

fspot (approved):

PR #866, opened 2021-04-07
feat(pipeline): display more readable errors

alice-sevin:
Display errors directly in step.
( errors are going to be filtered from Tucana side, we still return step errors for now ).
Tucana will also return an error in index case (another PR to come, this one need to be merged first)

Before:
before

After:
after

Need to review: https://github.com/ToucanToco/weaverbird/pull/864 first

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -1,5 +1,6 @@
 import { ActionContext, ActionTree } from 'vuex';
 
+import { BackendError } from '@/lib/backend';
 import { addLocalUniquesToDataset, updateLocalUniquesFromDatabase } from '@/lib/dataset/helpers.ts';
 import { pageOffset } from '@/lib/dataset/pagination';
 import { Pipeline } from '@/lib/steps';
@@ -24,6 +25,13 @@ function loading(type: 'dataset' | 'uniqueValues') {
 
     return descriptor;
   };
+}
+
+// format error to fit BackendError interface props
+function formatError(error: any): BackendError {
+  return typeof error === 'string'
+    ? { type: 'error', message: error.toString() }
+    : { type: 'error', ...error };
 }
 
 /**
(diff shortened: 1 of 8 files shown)


alice-sevin:
[ETQC, je différencie visuellement la step sur laquelle survient une erreur sur la vue pipeline](https://trello.com/c/X0i7lJOL/1503-etqc-je-diff%C3%A9rencie-visuellement-la-step-sur-laquelle-survient-une-erreur-sur-la-vue-pipeline)

tests/unit/store.spec.ts
```diff
@@ -336,6 +336,25 @@ describe('getter tests', () => {
...
+      const state = buildState({
+        backendMessages: [{ type: 'error', message: 'lalalolilol', index: 3 }],
+      });
+      expect(getters.stepErrors(state, {}, {}, {})(0)).toBeUndefined();
+    });
+    it('should return undefined if pipeline has no errors', function() {
+      const state = buildState({});
+      expect(getters.stepErrors(state, {}, {}, {})(0)).toBeUndefined();
+    });
+  });
```
davinov, line 357: :100:

tests/unit/step.spec.ts
```diff
@@ -239,4 +239,46 @@ describe('Step.vue', () => {
...
+          ],
+        }),
+      );
+      const wrapper = mount(PipelineComponent, { store, localVue });
+      const stepsArray = wrapper.findAll(Step);
+      const replaceStep = stepsArray.at(1);
+      expect(replaceStep.classes()).toContain('query-pipeline-step__container--errors');
+    });
+
+    it('should override with disabled style when disabled', async () => {
```
davinov, line 285: Good point! Normally we should never have errors on a disabled step, but in a temporary state when an error is there and we ask for a new preview before the error ; I would never have thought of that case, nice catch :p

src/store/actions.ts
```diff
@@ -26,6 +27,13 @@ function loading(type: 'dataset' | 'uniqueValues') {
   };
 }
 
+// format error to fit BackendError interface props
```
davinov, line 30: Nice to have extracted that :+1:

davinov (approved):
So nice! Thanks a lot!

tests/unit/store.spec.ts
alice-sevin: Ah mince je viens de voir que j'appelle pas la bonne méthode ligne 344, je corrige ;)

davinov: Ah bien vu, j'avais pas vu! :eyeglasses:

PR #870, opened 2021-04-08
feat(step-form): display backend error if any

alice-sevin:
after

:information_source: Good to know, as we pass error directly from pipeline if edit a disabled step with errors it won't handle an error in the step form edition interface (for now we don't do runtime errors for backend so we can't do better )

### DIFF
--- a/src/components/QueryBuilder.vue
+++ b/src/components/QueryBuilder.vue
@@ -4,9 +4,11 @@
       <component
         key="stepForm"
         :is="formComponent"
+        ref="step"
         :initialStepValue="stepFormInitialValue"
         :stepFormDefaults="stepFormDefaults"
         :isStepCreation="isStepCreation"
+        :backendError="editedStepBackendError"
         @back="closeStepForm"
         @formSaved="saveStep"
       />
@@ -51,6 +53,7 @@ import StepFormsComponents from './stepforms';
 })
 export default class QueryBuilder extends Vue {
   version = version; // display the current version of the package
+  editedStepBackendError: string | undefined = undefined;
 
   @VQBModule.State currentStepFormName!: PipelineStepName;
   @VQBModule.State stepFormInitialValue!: object;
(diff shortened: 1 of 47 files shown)


alice-sevin:
[ETQC je peux afficher le détail de l'erreur depuis le step de pipeline ciblé](https://trello.com/c/nBd5cbd6/1504-etqc-je-peux-afficher-le-d%C3%A9tail-de-lerreur-depuis-le-step-de-pipeline-cibl%C3%A9)

src/components/QueryBuilder.vue
```diff
@@ -51,6 +53,7 @@ import StepFormsComponents from './stepforms';
 })
 export default class QueryBuilder extends Vue {
   version = version; // display the current version of the package
+  editedStepBackendError: string | undefined = undefined;
```
alice-sevin, line 56: Might not be the better implementation but we need to keep data even if the store is refreshed (data is updated when entering the step so the error won't exist anymore)

davinov: It's a good idea to have done this! With a quick comment describing the issue it'll perfectly do the job :+1:

src/components/stepforms/StepFormHeader.vue
```diff
@@ -1,33 +1,41 @@
...
+          rel="noopener"
+          :data-version="props.version"
+        >
+          <i class="fas fa-question-circle" aria-hidden="true" />
+        </a>
+      </div>
+      <div class="step-edit-form__empty" />
+    </div>
+    <div v-if="props.backendError" class="step-edit-form__error">
+      <strong>{{ props.backendError }}</strong>
```
davinov, line 22: Did you have a specific reason to put it in he header of the form rather than outside in a specific component ?
(Both works, I'm just curious :p)

davinov (approved):
:+1:

src/components/QueryBuilder.vue
alice-sevin: I'll do it :ok_hand:

src/components/stepforms/StepFormHeader.vue
alice-sevin: Less code, introduce a component directly in stepform need to add it to every steps. It's just more text to add so a bigger package but we could totally do it and it will be easier to change if necessary

PR #872, opened 2021-04-12
fix(variables): retrieve variable identifier from malformatted variables

alice-sevin:
Seems that we shouldn't use the -1 length in this case, work for all cases including new tests for malformatted variables (without spaces or some spaces only)


alice-sevin:
[ETQC, je ne vois pas advanced variable s'il manque juste un espace dans ma variable <%= %>](https://trello.com/c/x2Q5zXqu/1926-etqc-je-ne-vois-pas-advanced-variable-sil-manque-juste-un-espace-dans-ma-variable)

davinov (approved):
:+1:

PR #875, opened 2021-04-13
fix(filter-condition): update value when changing widget

alice-sevin:
Case 1: changing operator.
Add a case for date input to keep date for matches operator (otherwise date should be a string to match the keepCurrentValueIfCompatibleType logic (which apply a default value if type is not the same than default value one ).
Ex: we choose an eq operator, we change to ne operator (default value of ne oprator is '', and we have a date selected, we compare a date to a string so we reapply '' but we want to keep the date).
So we add a case for date with method keepCurrentValueIfCompatibleDate when column type is date

Case 2: changing column type
We need to reverify operator in this case too because we may don't want to keep old value.

Before:
before-o

After:
after-o

### DIFF
--- a/tests/unit/filter-simple-condition-widget.spec.ts
+++ b/tests/unit/filter-simple-condition-widget.spec.ts
@@ -341,5 +341,21 @@ describe('Widget FilterSimpleCondition', () => {
         { column: 'columnA', value: null, operator: 'eq' },
       ]);
     });
+    it('should transform invalid dates to valid date when changing the column type', async () => {
+      createWrapper(shallowMount, {
+        value: { column: 'columnA', value: new Date('2021-01-01'), operator: 'eq' },
+      });
+      wrapper.setProps({ columnTypes: { columnA: 'string' } });
+      await wrapper.vm.$nextTick();
+      // relaunch operator validation automatically when changing the column type
+      expect(wrapper.emitted().input[0]).toEqual([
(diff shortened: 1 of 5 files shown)


alice-sevin:
[Dans une FilterStep, lorsque je change d'opérateur, et que je sauve, la date devient invalide](https://trello.com/c/yrmk59m5/1924-dans-une-filterstep-lorsque-je-change-dop%C3%A9rateur-et-que-je-sauve-la-date-devient-invalide)

tests/unit/filter-simple-condition-widget.spec.ts
```diff
@@ -351,11 +351,6 @@ describe('Widget FilterSimpleCondition', () => {
       expect(wrapper.emitted().input[0]).toEqual([
         { column: 'columnA', value: '', operator: 'eq' },
       ]);
-      wrapper.setProps({ columnTypes: { columnA: 'date' } });
```
alice-sevin: Le test ne peux pas passer car il ne se passe rien déerrière l'emit (on a pas le comportement réel)

davinov (approved):
:+1: Thanks for treating this weird/delicate topic
I've added a technical debt card about it so we can make this easier in a near future: https://trello.com/c/pZqtMkjh/1957-simplification-de-la-logique-des-types-dans-la-filterstep

PR #878, opened 2021-04-19
pandas/waterfall: the column 'TYPE_waterfall' was wrong, leading to bad drilling

luc-leonard:


luc-leonard:
[Waterfall groupby bug](https://trello.com/c/hSF7C5Bw/1978-waterfall-groupby-bug)

server/tests/steps/test_waterfall.py
```diff
@@ -153,3 +153,58 @@ def test_bug_duplicate_rows():
...
+    base_df = pd.DataFrame(
+        {
+            'grand parent': ['A', '1/A', 'i/1/A'] * 2,
+            'parent': ['1/A', 'i/1/A', 'U/i/1/A'] * 2,
+            'label': ['i/1/A', 'U/i/1/A', 'x/U/i/1/A'] * 2,
+            'variable': ['A'] * 3 + ['B'] * 3,
+            'value': [1, 2, 3, 11, 12, 13],
+        }
+    )
+    print(base_df)
```
davinov, line 168: Forgotten print?

server/tests/steps/test_waterfall.py
```diff
@@ -153,3 +153,58 @@ def test_bug_duplicate_rows():
     )
 
     assert_dataframes_equals(result_df, expected_df)
+
+
+def test_waterfall_bug_drill():
```
davinov, line 158: Mayeb it would be nice to add a commentary explaining this bug:
```suggestion
def test_waterfall_bug_drill():
```
```suggestion
def test_waterfall_bug_drill():
"""
Tuple (label, parent) should be unique only among one "group by" sub-df.
"""
```

server/tests/steps/test_waterfall.py
```diff
@@ -153,3 +153,58 @@ def test_bug_duplicate_rows():
...
+    base_df = pd.DataFrame(
+        {
+            'grand parent': ['A', '1/A', 'i/1/A'] * 2,
+            'parent': ['1/A', 'i/1/A', 'U/i/1/A'] * 2,
+            'label': ['i/1/A', 'U/i/1/A', 'x/U/i/1/A'] * 2,
```
davinov, line 163: This example is quite complex to undersand due to labels :
Maybe it's me but I think something a bit more concrete would be easier to read:
```suggestion
            'grand parent': ['France', 'IDF', 'Paris'] * 2,
            'parent': ['IDF', 'Paris', 'XXe arr'] * 2,
            'label': ['Paris', 'XXe arr', 'Place des Fêtes'] * 2,
```
or any other well-known hierarchy

davinov: Actually my sugegstion is very frenchy-centric.
Food > Vegetarian > Fruits > Berries > Blueberries may be easier to understand for everyone :p

server/tests/steps/test_waterfall.py
```diff
@@ -153,3 +153,58 @@ def test_bug_duplicate_rows():
...
+    print(base_df)
+    waterfall_step = WaterfallStep(
+        name='waterfall',
+        valueColumn='value',
+        milestonesColumn='variable',
+        start='A',
+        end='B',
+        labelsColumn='label',
+        parentsColumn='parent',
+        groupby=['grand parent'],
```
davinov, line 177: I'm quite surprise here, I would have guessed that it was the "variable" column which was the group by.
I think then that it would be a clearer example to not use something in the hierarchy as group by column, but an other unrelated dimenion.
Like `type: ["alpha", "beta"]`

davinov (changes requested):
Thanks for this PR!
The fix is very clear 👌 

Here are a few suggestions to make the test clearer:

server/tests/steps/test_waterfall.py
luc-leonard: OOps :sweat:

davinov (approved):

PR #879, opened 2021-04-19
fix(date-step-form): always display custom input format when custom format is selected

alice-sevin:
Issue: We used a computed to define selected format (this computed was updated when changing value of custom format input (not only selected format in dropdown). 
We also automatically hide the custom format input if the format was present in the dropdown.

Resolve:
Remove computed property and use method to set selected format (custom format input update won't impact dropdown anymore)
Display custom format input when selected format is 'custom' 

:information_source: Made it on both steps (from and to) because logic is the same for the two steps

Before:
before

After:
after

### DIFF
--- a/src/components/stepforms/FromDateStepForm.vue
+++ b/src/components/stepforms/FromDateStepForm.vue
@@ -125,17 +125,6 @@ export default class FromDateStepForm extends BaseStepForm<FromDateStep> {
   ];
   selectedFormat?: FormatOption;
 
-  created() {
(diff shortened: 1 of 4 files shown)


alice-sevin:
[Bug saisie d'un format de date custom reprenant ceux part défauts](https://trello.com/c/r68dQqTH/1980-bug-saisie-dun-format-de-date-custom-reprenant-ceux-part-d%C3%A9fauts)

austil (approved):
All good :ok_hand:

PR #890, opened 2021-04-29
fix(popover): put z-index: 2 to popover

alice-sevin:


aklom:
[As a dashboard builder, I can see popovers in the VQB](https://trello.com/c/sUX4OuW4/682-as-a-dashboard-builder-i-can-see-popovers-in-the-vqb)

PR #895, opened 2021-05-03
feat(front): customize popover container

Dalanir:


davinov (commented):
Thanks for taking care of this!
It seems like a good idea of use provide/inject for this. However, as it's something quite advanced to configure, is it enough to use the injection with a default value on the popover, and let the host component define the provide? That would simplify this PR, because it's a bit unfortunate to have to add this prop to injection logic in every "component entrypoint".

Dalanir:
Yes I wasn't sure, I wanted to let the Popover as dumb as possible but I think you're right for our use case there

Dalanir:
@davinov let me know what you think about this :)

davinov (approved):
Thanks @Dalanir , I've simplify a bit this PR and provided an example on how to customize the container from the host app (see this story: https://github.com/ToucanToco/weaverbird/pull/895/files#diff-a43471fe27166b483502ba1d9553ecbe860a3119e9a434d6a7132f2a47a8be5aR23-R46) What do you think ?

Dalanir:
Hi @davinov, have you test it? Because it has broken the behavior intended I think

PR #903, opened 2021-05-10
fix(conditionnal-step): enable to use dates in condition

alice-sevin:
Enable to use dates in conditionnal step (same behaviour than in filter step)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- IfThenElseStep: enable to use dates in condition
+
 ## [0.47.0] - 2021-04-29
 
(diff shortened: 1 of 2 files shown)


alice-sevin:
[As a C, I would like to be able to use a date format in a conditional column](https://trello.com/c/mysddkm3/2075-as-a-c-i-would-like-to-be-able-to-use-a-date-format-in-a-conditional-column)

alice-sevin:
[1- As a C, I would like to be able to use a date format in a conditional column](https://trello.com/c/czjoo3cl/4735-1-as-a-c-i-would-like-to-be-able-to-use-a-date-format-in-a-conditional-column)

davinov (approved):
:+1:

PR #909, opened 2021-05-17
fix(filter-condition): enable to have undefined value

alice-sevin:
In case of 'is null' or 'is not null' condition, we remove the value field and set value to 'null'. It seems that Pandas delete 'null' value field from properties.
So when you come back to edit step, you can't validate because value field is required.
We should not required 'value' field

### DIFF
--- a/src/components/stepforms/schemas/filter.ts
+++ b/src/components/stepforms/schemas/filter.ts
@@ -128,7 +128,7 @@ export default {
           },
         },
       },
-      required: ['column', 'value', 'operator'],
+      required: ['column', 'operator'],
       additionalProperties: false,
     },
   },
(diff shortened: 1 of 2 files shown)


alice-sevin:
[1- As app designer, I can edit a filter step of a prepared dataset afterwards](https://trello.com/c/THsfsk2p/4742-1-as-app-designer-i-can-edit-a-filter-step-of-a-prepared-dataset-afterwards)

alice-sevin:
[As app designer, I can edit a filter step of a prepared dataset afterwards](https://trello.com/c/qlLIFeCn/2100-as-app-designer-i-can-edit-a-filter-step-of-a-prepared-dataset-afterwards)

davinov (approved):

PR #914, opened 2021-06-02
feat(server/pandas): add benchmarks

luc-leonard:
this pr adds benchmark for most of the steps in the pandas implementation


fspot:
@luc-leonard nice ! est-ce que à partir de là tu as une idée de comment tester qu'on n'a pas de régression de perfs ?

davinov (approved):
The benchmark table is very neat! 
Screenshot from 2021-06-04 10-24-57

It gives a clear idea on which step may be slow. Once correlated with steps usage metrics, we'll have insights on which ones to optimize :+1:

PR #919, opened 2021-06-11
fix(variable chooser): always display date in UTC

austil:
Before, tooltip & value where different (June 8 midnight vs June 7 22hours) : 

variable_chooser-date-before

After, tooltip & value are consistent : 

variable_chooser-date-after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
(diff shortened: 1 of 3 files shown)


austil:
[Variables de date côté front : premier du mois == 2e jours du mois ?](https://trello.com/c/GbZmxVTi/2024-variables-de-date-c%C3%B4t%C3%A9-front-premier-du-mois-2e-jours-du-mois)

alice-sevin (approved):

PR #920, opened 2021-06-14
fix(code-editor): move fixed height to container

alice-sevin:
There is some new css code in our other repository that override the height: 300px of the code editor, we should move it to the container and keep a 100% height logic to be similar on both repository (behaviour is the same that before)

Before:
(You can see that height: 300% is override byt height: 100%)
before

After:
after

(No impact on custom step because it was 300px height too, same behaviour than before :ok_hand: )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- Code editor: move static height to container to avoid css override
+
 ## [0.49.1] - 2021-06-09
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


alice-sevin:
[AdvancedVariable popup size](https://trello.com/c/8N4RaBDV/2213-advancedvariable-popup-size)

austil (approved):
:muscle:

alice-sevin:
[1- As a C, I would like my Advariable pop up to be displayed properly](https://trello.com/c/ath0FICV/4790-1-as-a-c-i-would-like-my-advariable-pop-up-to-be-displayed-properly)

PR #921, opened 2021-06-14
fix(backend-errors): use same message key for errmsg formatted error

alice-sevin:
Some mongo errors seems to have different interfaces than other errors (string, object with index to associate a specific step, or mongo error with code and errmsg property).
To be sure to use the same key to retrieve message outside of weaverbird, we add the message key to the backendError object associated with errmsg property coming from Mongo lib

Current Mongo error is : 
`{
        code: 241,
        codeName: 'ConversionFailure',
        errmsg: 'global error message',
        ok: 0,
      }`
      
 Panda error is:
 `{ message: 'step specific error message', index: 1  }`
 
 Other error can be only a string:
 `ConversionFailure: global error message`
      
 We should have a `message` property to be able to retrieve the message based on a similar value for each errors

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -31,7 +31,7 @@ function loading(type: 'dataset' | 'uniqueValues') {
 export function formatError(error: any): BackendError {
   return typeof error === 'string'
     ? { type: 'error', message: error.toString() }
-    : { type: 'error', ...error };
+    : { type: 'error', ...error, message: error.message ?? error.errmsg };
 }
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- Backend errors: use `message` as returned key for all type of errors
+
 ## [0.49.1] - 2021-06-09
 
 ### Fixed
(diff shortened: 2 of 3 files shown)


alice-sevin:
[Mongo error display](https://trello.com/c/2ppZ2sQ9/2210-mongo-error-display)

alice-sevin:
[1- As a C, I would like to have my mongo error properly displayed](https://trello.com/c/3lW5CxSa/4782-1-as-a-c-i-would-like-to-have-my-mongo-error-properly-displayed)

src/store/actions.ts
```diff
@@ -31,7 +31,7 @@ function loading(type: 'dataset' | 'uniqueValues') {
 export function formatError(error: any): BackendError {
   return typeof error === 'string'
     ? { type: 'error', message: error.toString() }
-    : { type: 'error', ...error };
+    : { type: 'error', ...error, message: error.message ?? error.errmsg };
```
davinov, line 34: Hm... I think that should be harmonized back-end side, or we will have to implement a different rule for every possible backend. Let's discuss what we choose to implement.

alice-sevin: Totally agree!

PR #929, opened 2021-06-22
fix: prevent uint32 in date extract results

fspot:
Cast them to int64, so we don't have problems when exporting to json.


luc-leonard (approved):
lgtm

davinov (approved):

PR #930, opened 2021-06-23
fix(step): change drag & drop grip icon

austil:
Change step's drag handle icon to something more similar to the intended design.

Before : 

weaverbird-handle-before

After : 

image

### DIFF
--- a/src/components/Step.vue
+++ b/src/components/Step.vue
@@ -32,7 +32,7 @@
             v-if="!isFirst"
             @click.stop
           >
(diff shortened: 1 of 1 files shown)


austil:
[Retour design: icône handle pas la même que celle dans la maquette ?](https://trello.com/c/xdTm8Mrf/2129-retour-design-ic%C3%B4ne-handle-pas-la-m%C3%AAme-que-celle-dans-la-maquette)

davinov (approved):
This is a revolution /20

PR #931, opened 2021-06-24
feat(action toolbar): new search integration

austil:
## What

Rework of the search integration to better fit our design mockups.

## How

Because the original integration heavily rely on `vue-multiselect` to provide search, it wasn't as straightforward as expected.
Vue-multiselect doesn't have props to control it's dropdown opening so we end up relying of the dom `.focus` method to have it open when the (new) parent popover is opened. 

- I created a new component that's a mix between the existing `action-toolbar-button` & the original `search-bar`
- then swapped `search-bar` with it in `action-toolbar`
- last but not least, tweaked styles & deleted the original implem

## Screenshot

Before : 

toucan_weaverbird_search-before

After : 

toucan_weaverbird_search-after

It's not perfect tho : the multiselect height depend on the browser window height & clicking on the button again make the search loose focus 

toucan_weaverbird_search-after-flows


austil:
[ETQC, je vois la barre des catégories de steps VQB bien intégrée](https://trello.com/c/Sv3yWQJR/2183-etqc-je-vois-la-barre-des-cat%C3%A9gories-de-steps-vqb-bien-int%C3%A9gr%C3%A9e)

davinov (approved):
Seems really good to me! Kudos for the clear commits log.

PR #934, opened 2021-07-05
docs(python): description of backends and their API

davinov:
**EDIT** : this is now the base branch for the new weaverbird architecture (will require a new major version because the api has slightly changed)

---

Provisional docs to describe the purpose and API of python's module.

Changes with the existing are:
- separate the Pipeline model for validation from any backend
- isolate the pandas executor in a submodule `weaverbird.backends.pandas_executor`
- expose only one function `execute_pipeline` (and take `domain_retriever` as argument) instead of instantiating a class `PipelineExecutor` and then calling the `execute_pipeline` method

:warning: This is just the docs! Implementation will come in another PR


docs/_docs/tech/python-package.md
```diff
@@ -5,24 +5,35 @@ permalink: /docs/python-package/
...
 
-The python module provide a way to run transformations described in a pipeline.
+The python module provide a way to turn pipelines into transformation functions or queries.
+
 
 ## Purpose
 
-This package provides utility functions to translate weaverbird's pipelines into Python functions, powered by [pandas](https://pandas.pydata.org/).
-It is meant as a building block to create servers capable of understanding and executing such pipelines, and returning results to clients.
+This package provides is meant as a building block to create servers capable of understanding and executing such pipelines, and returning results to clients.
```
fspot, line 13:
```suggestion
This package is meant as a building block to create servers capable of understanding and executing such pipelines, and returning results to clients.
```

fspot (approved):
juste une typo, très clair sinon :ok_hand:

PR #935, opened 2021-07-05
fix(formula): avoid Infinity in resulting dataframe

fspot:
The `formula` step of the python package could lead to `Infinity` in the resulting dataframe (after a division by 0).
`Infinity` does not have a json representation so it was sent as "null" to the frontend, which was misleading (dropna or fillna didn't work on these values).

Now, we replace these values by regular `np.nan`.

### DIFF
--- a/server/weaverbird/formula.py
+++ b/server/weaverbird/formula.py
@@ -1,5 +1,6 @@
 import re
 
+import numpy as np
 from pandas import DataFrame
 
 COLUMN_PATTERN = re.compile(r'(\`[^\`]*\`)')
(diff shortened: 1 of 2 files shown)


server/weaverbird/formula.py
```diff
@@ -30,7 +31,16 @@ def clean_formula(formula: str) -> str:
...
         # for all cases not handled by NumExpr
-        return df.eval(formula, engine='python')
+        result = df.eval(formula, engine='python')
+
+    try:
+        # eval can introduce Infinity values (when dividing by 0),
+        # which do not have a JSON representation.
+        # Let's replace them by NaN:
+        return result.replace([np.inf, -np.inf], np.nan)
+    except Exception:
```
davinov, line 44: Sonar doesn't cry about this? ^^

server/weaverbird/formula.py
```diff
@@ -30,7 +31,16 @@ def clean_formula(formula: str) -> str:
...
+        result = df.eval(formula)
     except Exception:
         # for all cases not handled by NumExpr
-        return df.eval(formula, engine='python')
+        result = df.eval(formula, engine='python')
+
+    try:
+        # eval can introduce Infinity values (when dividing by 0),
+        # which do not have a JSON representation.
+        # Let's replace them by NaN:
```
davinov, line 42: :+1: Thanks for the very clear comment

davinov (approved):
:+1:

davinov:
We're gonna get conflicts with the reorg ahaha :p

davinov:
To be backported on 0.4.x

PR #936, opened 2021-07-05
refactor(server): isolate pipeline and pandas_executor modules

davinov:
Implement changes described in #934

- [x] isolate `pipeline` module with pipeline and steps models
- [x] create a new `backends.pandas_executor` module
- [x] move all steps implementations in that new module
- [x] adapt `pandas_executor` core functions signatures


server/weaverbird/backends/pandas_executor/steps/formula.py
```diff
@@ -0,0 +1,15 @@
+from pandas import DataFrame
+
+from weaverbird.backends.pandas_executor.types import DomainRetriever, PipelineExecutor
+from weaverbird.formula import clean_formula, eval_formula
```
davinov, line 4: clean_formula may stay out of a specific backend, but IMO it seems appropriate to move eval_formula inside the pandas_backend. Maybe in the weaverbird.backends.pandas_executor.steps.utils module?

fspot:
:information_source: je n'ai pas réorganisé l'arborescence des fichiers de tests car ça viendra avec les prochaines PRs

server/weaverbird/backends/pandas_executor/steps/ifthenelse.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any
+
+import numpy as np
+from pandas import DataFrame
+
+from weaverbird.backends.pandas_executor.types import DomainRetriever, PipelineExecutor
+from weaverbird.formula import clean_formula, eval_formula
```
davinov: Same note here, eval_formula may reside inside the pandas_executor module?

server/weaverbird/backends/pandas_executor/steps/join.py
```diff
@@ -0,0 +1,33 @@
+from pandas import DataFrame, merge
+
+from weaverbird.backends.pandas_executor.types import DomainRetriever, PipelineExecutor
+from weaverbird.pipeline.steps import JoinStep
+
+from .utils.cleaning import rename_duplicated_columns
+from .utils.combination import resolve_pipeline_for_combination
```
davinov, line 7: Should we use relative imports or always from weaverbird? (Idk if there is a good practice for this, ; I think my IDE puts always path from the root module I think when refactoring)

server/weaverbird/backends/pandas_executor/steps/waterfall.py
```diff
@@ -0,0 +1,166 @@
...
+        return _RESULT_COLUMN
+    else:
+        return step.labelsColumn
+
+
+def get_join_key(step: WaterfallStep):
+    if step.parentsColumn is None:
+        return [step.labelsColumn] + step.groupby
+    else:
+        return [step.labelsColumn, step.parentsColumn] + step.groupby
```
davinov, line 166: You've made it to the end of steps :tada: :tada: :tada: :tada: :tada:  big thanks and congrats!

server/weaverbird/backends/pandas_executor/pipeline_executor.py
```diff
@@ -34,7 +34,7 @@ def execute_pipeline(
     for index, step in enumerate(steps):
         try:
             with stopwatch:
-                df = steps_executors[step.name](
+                df = steps_executors[step.name](  # type: ignore
```
davinov, line 37: Oh sad that we have to ignore that :'(
I tried to create a type `StepExecutor` and would have like to type every of these functions with is, but I din't find a way in python to declare the type of function directly, using `Callable`. I was counting on mypy "understanding" the compatibity of this type with the definitions.
Up to search a bit more today, but in the meantime ignoring is fine

server/weaverbird/backends/pandas_executor/steps/ifthenelse.py
davinov: Resolved in the final commit, thx!

server/weaverbird/backends/pandas_executor/steps/join.py
fspot: Both are fine to me, I just avoid to use relative imports for parents modules (`from ...foo.bar`) because I find it difficult to understand.

fspot (approved):

server/weaverbird/backends/pandas_executor/pipeline_executor.py
fspot: fixed by ad20c88

PR #937, opened 2021-07-07
feat(aggregate steps): add unique column names validation

raphaelvignes:
Issue: https://toucantoco.atlassian.net/browse/ASK-60 
Solution: Enforce unique column names for aggregation column

PR #939, opened 2021-07-07
chore(version): bump server 0.4.3

raphaelvignes:
Bump the server's version to 0.4.3

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "weaverbird"
-version = "0.4.2"
+version = "0.4.3"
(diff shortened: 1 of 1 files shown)


davinov:
Congrats for your contributions !

Is this a duplicate of https://github.com/ToucanToco/weaverbird/pull/938 ?
If the aim is to release https://github.com/ToucanToco/weaverbird/pull/937, no need to increment package.json and the front changelog, the change of version can only be reflected in /server.

(Which makes me think it would be a good idea to create a CHANGELOG also for the python module, not juste the js one)

raphaelvignes:
Sorry, I guess https://github.com/ToucanToco/weaverbird/pull/938 needs to be revert. 
I updated this one.

PR #942, opened 2021-07-08
tests(server): generate tests from fixtures

davinov:
To make creating backends easier, let's pull test cases out of pandas executor's specific test python files and make them available for other "test runners".
This will ensure more consistency between backends results, or make explicit when they differ.
It will also enable to write one runner for a new backend, and have the whole cases available at once.


server/tests/backends/pandas_executor/test_pandas_executor_steps.py
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    # Path of the file specifying the expected output dataframe (xxx.out.json)
+    out_file_path = path.join(path.dirname(x), f'{case_name}.out.json')
+
+    test_cases.append(
+        pytest.param(
+            case_id, step_file_path, in_file_path, other_in_files_domains, out_file_path, id=case_id
+        )
+    )
```
davinov, line 41: This will be to extract so that other backends test runners can use it

piotch:
Est-ce qu'on ne pourrait avoir les tests case contenu dans un seul fichier plutôt que trois ? Cela permettrai d'expliciter dans une seule data structure JSON ce qui aujourd'hui est encodé dans une convention de noms de fichiers (cela me paraît plus fragile).

davinov:
@piotch Here is what it would be with only one file by test case: https://github.com/ToucanToco/weaverbird/pull/942/commits/7cf68613046c9ab2a99d16b71e4d850ea29e06c2
Seems easier indeed! (maybe the only drawback is the human legibility of the file)

server/tests/backends/pandas_executor/test_pandas_executor_steps.py
fspot: :ok_hand:

fspot (approved):

PR #950, opened 2021-07-16
feat(sql translator): create a SQL translator

raphaelvignes:
See PAT: https://toucantoco.atlassian.net/wiki/spaces/TTA/pages/3058991105


raphaelvignes:
[(3) [Snowflake VQB Translator] - As a developer I implement the logic to switch between pipeline execution and pipeline translation](https://trello.com/c/SnLkjiWS/1081-3-snowflake-vqb-translator-as-a-developer-i-implement-the-logic-to-switch-between-pipeline-execution-and-pipeline-translation)

server/weaverbird/backends/pipeline_translator/pipeline_translator.py
```diff
@@ -0,0 +1,43 @@
...
+from server.weaverbird.backends.pipeline_translator.types import QueryRetriever, PipelineTranslationReport, \
+    StepTranslationReport
+from server.weaverbird.backends.pipeline_translator.steps import steps_translators
+
+
+def translate_pipeline(
+    pipeline_to_translate: Pipeline,
+    query_retriever: QueryRetriever
+) -> Tuple[str, PipelineTranslationReport]:
+    """ For now, steps_translators are only SQL steps translators """
```
davinov: I'm nit sure I understand this docstring. Do you mean than not all steps are implemented?

server/weaverbird/backends/pipeline_translator/pipeline_translator.py
```diff
@@ -0,0 +1,43 @@
...
+                query_retriever=query_retriever,
+                translate_pipeline=translate_pipeline,
+            )
+            translate_report.append(
+                StepTranslationReport(
+                    step_index=index
+                )
+            )
+        except Exception as e:
+            raise PipelineTranslationFailure(step, index, e) from e
```
davinov: Nice to have created this exception :100:

server/weaverbird/backends/pipeline_translator/steps/utils/query_transformation.py
```diff
@@ -0,0 +1,23 @@
...
+
+    query = f'select * from ({query}) '
+    sql_operator = {
+        'eq': '=',
+        'ne': '!=',
+        'lt': '<',
+        'le': '<=',
+        'gt': '>',
+        'ge': '>=',
+    }
```
davinov: You can put this in a constant outside the function, so it will not be reallocated on each function call.

server/weaverbird/backends/pipeline_translator/types.py
```diff
@@ -0,0 +1,28 @@
...
+
+class StepTranslator(Protocol):
+    def __call__(
+        self,
+        step: Any,
+        query: str,
+        query_retriever: Optional[QueryRetriever],
+        translate_pipeline: Optional[PipelineTranslator]
+
+    ):
```
davinov: What is the return type of a StepTranslator?

server/weaverbird/pipeline/conditions.py
```diff
@@ -1,5 +1,5 @@
 from abc import ABC
-from typing import Any, List, Literal, Union
+from typing import Any, List, Literal, Union, Dict
```
davinov: Is this useful?

davinov (commented):
Thanks for this POC!

Two notes:
- I think it would be wise to prefix your types by SQL: SQLPipelineTransator, SQLQueryRetriever, etc. As they will be more than one pipeline translator in a near future, this will help not getting lost while searching for symbols with IDEs.
- I think it's strange that we have to provide a function than takes a domain and return a complete sql query as an input. I would imagine we can have a domain to table name function, but I would expect the translator to be able to craft the query from the table name autonomously.
- Be ready to switch from str to a more structured type for the return of StepTranslator: some steps will act on the select part, others on the where, other on the from, some will add a join, etc. Not necessary for now of course, the two steps implemented now doesn't require it.
- About the tests: it's cool to have unit tests, but it would be even better to have "end to end" tests, i.e trying to execute this against a real sql database. (A SQLite would be fine) It's easy to extract some tests cases for the filter step we have to a JSON file (see how to in the tests/fixtures folder) and not so easy but interesting to implement a test runner for these cases as we did for the pandas executor. Also not mandatory to do right now, but would save you a great amount of tests writing.

docs/_data/docs.yml
```diff
@@ -51,7 +51,7 @@
     - steps
     - new-step
     - known-limitations
-    - translators
+    - pipeline_translator
```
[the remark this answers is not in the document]
raphaelvignes: Yes, my bad.

server/weaverbird/backends/pipeline_translator/pipeline_translator.py
raphaelvignes: I will clarify. I was meaning, for now the only translator we will implement is a SQL one. 
But maybe in the future we'll have other ones :)

raphaelvignes:
Thanks for this review ! 
My answers about your global comment (in the same order):
* Agreed, using an SQL prefix will clarify the essence of these objects.
* Yes, providing a function to retrieve only a query string is a bit dumb, I blindly duplicated the logic of PipelineExecutor. The first step will be provided a table name instead of a user defined query.  
* One of the next step, will be to define a more robust & structured way of forwarding the transformed query between steps. But not for this week ! 
* I keep in mind this last comment, we definitely do need integration tests (for connectors as well).

server/weaverbird/backends/pipeline_translator/pipeline_translator.py
davinov: Ah OK!
But it makes no sense to have a translator that would use a sql implementation for one step and another like mongo for another step. It seems to me that, at least for now, we don't mix translators. So inside the `SQLTransaltor`, I think it's logical that there will always be only SQL steps.

davinov:
> One of the next step, will be to define a more robust & structured way of forwarding the transformed query between steps. But not for this week !

Great! Congrats again on this PR :) It's awesome to see new translators coming up nicely :clap:

server/weaverbird/backends/pipeline_translator/steps/utils/query_transformation.py
raphaelvignes: Great, thanks !

server/weaverbird/backends/pipeline_translator/types.py
raphaelvignes: A string for now. 
It will be updated, when I'll change the data structure forwarded between steps.

server/weaverbird/backends/sql_translator/sql_pipeline_translator.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            query = sql_steps_translators[step.name](
+                step,
+                query,
+                sql_query_retriever=sql_query_retriever,
+                sql_translate_pipeline=translate_pipeline,
+            )
+            translate_report.append(SQLStepTranslationReport(step_index=index))
+        except Exception as e:
+            raise SQLPipelineTranslationFailure(step, index, e) from e
```
f-bb-toucantoco, line 28: plutôt que de raise directement l'exception dès qu'une translation ne marche pas, ça serait bien de finir le translate, de conserver toutes les exceptions et de remontées toutes les causes
Ca donnera plus une visibilité d'ensemble que de casser à chaque tour et d'avancer pas à pas

server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
```diff
@@ -0,0 +1,20 @@
...
+    'gt': '>',
+    'ge': '>=',
+}
+
+
+def apply_condition(condition: Condition, query: str) -> str:
+    query = f'select * from ({query}) '
+
+    if isinstance(condition, ComparisonCondition):
+        query += f'where {condition.column} {SQL_COMPARISON_OPERATORS[condition.operator]} {condition.value}'
```
f-bb-toucantoco, line 17: un check existe bien sur "condition.value" et "condition.column" pour éviter d'avoir des choses bizarres dans la requête générée ?

f-bb-toucantoco (approved):

server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
raphaelvignes: Bonne idée. J'ajouterai ça dans la carte: https://trello.com/c/H7tzJFOw/1088-5-snowflake-vqb-translator-as-an-app-maker-i-can-filter-using-snowflake-vqb

server/weaverbird/backends/sql_translator/sql_pipeline_translator.py
raphaelvignes: Je ne sais pas si c'est compatible avec la gestion d'erreur par le front. 
De mémoire quand tu réalises tes transformations tu dois avoir des erreurs à chaques étape, comme ici: 
image

PR #953, opened 2021-07-20
fix(search): display search correctly

alice-sevin:
Due to overflow hidden, the search was not display correctly. But as search height by default is going over the popover, we need to fix it with important (added dynamically to DOM by multiselect plugin, force it by css using !important).

Before:
before

After:
after

After FF:
after-firefox

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Fixed
+- Display search correctly in chrome browser
+
 ## [0.53.0] - 2021-07-15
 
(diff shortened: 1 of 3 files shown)


alice-sevin:
[Décalage dans l'inté du champ de recherche du weaverbird](https://trello.com/c/u3Dba9jg/2290-d%C3%A9calage-dans-lint%C3%A9-du-champ-de-recherche-du-weaverbird)

davinov (approved):
Nicely found, thanks

austil (approved):
:ok_hand: :+1:

PR #928, opened 2021-06-22
feat(pipeline): handle cmd for copy/pasting + backspace key to delete

Dalanir:
To ease the life of Mac users.

PR #954, opened 2021-07-20
fix(pipeline): enable delete shortcut only when no other input is focused

austil:
Following #928 

Before, pressing backspace went editing a input trigger the current step deletion modal : 

toucan_weaverbird_delete_shortcut-before

After, it's only triggered when not input field are focused : 

toucan_weaverbird_delete_shortcut-after


davinov (approved):
:+1:

PR #957, opened 2021-07-21
doc(readme): better introduction

austil:
If I had to introduce Weaverbird to someone, here is what I'll say.


davinov (approved):
More than 40 transformations steps!
Friendly!
Serves coffee and cookies!
:100: /20

PR #966, opened 2021-08-02
feat(resizable): adapt col width to fit max char per col option

alice-sevin:
Add an option to v-resizable table to display a certain amount of chars
Add another option to inform about the current chars displayed in col (can't be calculated due to white space no-wrap) that will cropp label
Add another option to target label in DOM

Adapt width of col based on max chars to display (retrieve pixel size for current chars and add missingwidth to fit the missing chars).
Adapt css to fit the new behaviour.

Tests:
We can't really tests DOM events (width with jest so we just test that function is correctly called)

Before:
before
before

After (more characters display (20 at max)):
after
after

### DIFF
--- a/src/components/DataViewer.vue
+++ b/src/components/DataViewer.vue
@@ -13,7 +13,7 @@
           data-cy="weaverbird-data-viewer-table"
           v-resizable="{
             columns: columnNames,
-            labelClass: 'data-viewer__header-label',
+            labelTargetClass: 'data-viewer__header-label',
             classes: {
               table: 'data-viewer-table--resizable',
               handler: 'data-viewer__header-cell--resizable',
--- a/src/components/DataViewer.vue
+++ b/src/components/DataViewer.vue
@@ -354,6 +354,7 @@ export default class DataViewer extends Vue {
   cursor: pointer;
   font-weight: bold;
   padding: 6px 8px;
+  padding-right: 23px;
 }
 
 .data-viewer__header-cell--active {
@@ -375,9 +376,6 @@ export default class DataViewer extends Vue {
 }
 
 .data-viewer__header-cell--disabled {
-  .data-viewer__header-label {
-    white-space: nowrap;
-  }
   .data-viewer__header-icon {
     pointer-events: none;
   }
@@ -389,7 +387,7 @@ export default class DataViewer extends Vue {
   text-overflow: ellipsis;
   max-width: calc(100% - 40px);
   overflow: hidden;
-  padding-right: 23px;
+  white-space: nowrap;
 }
 
 .data-viewer__header-icon {
--- a/src/components/DataViewer.vue
+++ b/src/components/DataViewer.vue
@@ -13,6 +13,7 @@
           data-cy="weaverbird-data-viewer-table"
           v-resizable="{
             columns: columnNames,
+            labelClass: 'data-viewer__header-label',
             classes: {
               table: 'data-viewer-table--resizable',
               handler: 'data-viewer__header-cell--resizable',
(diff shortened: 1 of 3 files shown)


alice-sevin:
[DataTable: taille par défaut adaptable avec le nbr de caractères du nom de la colonne](https://trello.com/c/gasEy5EE/2326-datatable-taille-par-d%C3%A9faut-adaptable-avec-le-nbr-de-caract%C3%A8res-du-nom-de-la-colonne)

src/components/DataViewer.vue
```diff
@@ -13,6 +13,7 @@
           data-cy="weaverbird-data-viewer-table"
           v-resizable="{
             columns: columnNames,
+            labelClass: 'data-viewer__header-label',
             classes: {
               table: 'data-viewer-table--resizable',
               handler: 'data-viewer__header-cell--resizable',
```
austil, line 19: Why this class is not part of the `classes` prop below ?
Also I think that "label" is a little to vague, may be we can find a slightly more precise name like "headerLabels" ?

src/directives/resizable/ResizableTable.ts
```diff
@@ -20,6 +23,9 @@ export interface ResizableTableOptions {
     handler: string; // class applied to col handler
   };
   columns: string[] | number[]; // the columns associated to cols handlers
+  charsPerCol: number; // The actual number of chars displayed per col
+  maxCharsPerCol: number; // The max number of chars to display per col
```
austil, line 27: I'm not quite sure if I understand what `charPerCol` is, "actual number of chars" is a little vague (find a better name and/or improve the comment ?).
Is is the number of chars displayed on first render before we correct it if the text overflows ?
Also where does the 7.5 default value come from ?

src/directives/resizable/ResizableTable.ts
```diff
@@ -77,25 +83,52 @@ export default class ResizableTable {
...
+    if (!labelElement) return colElement.offsetWidth;
+    // Find the padding width in total col width
+    const padding = colElement.offsetWidth - labelElement.offsetWidth;
+    // Find the pixel width for a char based on label width and actual chars per col
+    const pixelPerChar = labelElement.offsetWidth / this.options.charsPerCol;
+    // Calculate needed width for col based on chars in content
+    const contentWidth = pixelPerChar * contentChars;
+    // Calculate max available width for col based on max chars per col
+    const maxWidth = pixelPerChar * this.options.maxCharsPerCol;
+    return padding + Math.ceil(contentWidth < maxWidth ? contentWidth : maxWidth);
```
austil, line 106: Logic is fine IMHO.
Here is just a little suggestion "pour la route" :wink: 

```suggestion
    return padding + Math.ceil(Math.min(contentWidth, maxWidth));
```

src/components/DataViewer.vue
alice-sevin: I will update the property name, it should be labelTargetClass because what's under classes are the classes added to element while the labelClass, is the DOM element to target ;)

src/directives/resizable/ResizableTable.ts
alice-sevin: Yes that is the chars render at first display to be able to calculate a ratio between label first width and number of chars display, that will give use the pixelPerChar that we can apply on the full chars length to get the final width

src/directives/resizable/ResizableTable.ts
alice-sevin: yes right ;)

austil (approved):
:green_circle: :checkered_flag: :+1:

PR #970, opened 2021-08-10
fix(makefile): adding poetry installation

Sanix-Darker:
# Summary
- Added the installation of poetry in `make install` inside the `Makefile` 

# Checks
- [x] Unit tests

### DIFF
--- a/server/Makefile
+++ b/server/Makefile
@@ -11,6 +11,8 @@ clean:
 
 .PHONY: install
 install:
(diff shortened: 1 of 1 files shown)


davinov (approved):
Nice catch ^^ totally forgot this and considered

So maybe we can change what GitHub actions are doing, calling `make install` instead of the action to install poetry?

PR #982, opened 2021-08-16
Add Trim Step

CharlesRngrd:
Hi Toucan,

As a conceptor, I frequently need to remove spaces in columns.
So here is my PR for the trim step.

Have a nice day ;)


davinov:
Hi @CharlesRngrd ! Thanks for contributing :hugs: 
I'll take a look at your PR very soon. Thanks for this work!

davinov:
Because of some private environment variables (our docker hub API key for example), this job https://github.com/ToucanToco/weaverbird/pull/982/checks?check_run_id=3380759457 couldn't go through
To avoid this, I've recreated your PR from the ToucanToco organization here: https://github.com/ToucanToco/weaverbird/pull/1007 and tests are green now :)
Also Circle tests seems not to execute when a PR comes from another source.... but we have a plan to migrate them to GitHub Actions to mitigate that in a near future :)

PR #983, opened 2021-08-16
chore(esjson): remove all occurencies to esjson

alice-sevin:
We want to have date input for every date column now, we can remove the feature flag security to always display the date input in this case.


alice-sevin:
[ETQAB, je peux filtrer avec l'input de type date quand ma colonne est de type date](https://trello.com/c/h4lqlucv/2385-etqab-je-peux-filtrer-avec-linput-de-type-date-quand-ma-colonne-est-de-type-date)

austil (approved):
:broom: yeah !

PR #990, opened 2021-08-17
feat(mongo): concatenate won't fail on non-string columns

CharlesRngrd:
Hi Toucan,

As a conceptor, I need to concat string and numeric columns.
Here is my PR to do so in one step !

Notes :
- In MongoDB, the query is updated for 4.0 (because $convert is available since 4.0)

Have a nice day,

Charles

### DIFF
--- a/src/lib/translators/mongo4.ts
+++ b/src/lib/translators/mongo4.ts
@@ -1,7 +1,7 @@
 /** This module contains mongo specific translation operations */
 
 import { $$ } from '@/lib/helpers';
-import { ConvertStep, ToDateStep, TrimStep } from '@/lib/steps';
+import { ConcatenateStep, ConvertStep, ToDateStep, TrimStep } from '@/lib/steps';
 import { Mongo36Translator } from '@/lib/translators/mongo';
 
 type PropMap<T> = { [prop: string]: T };
@@ -11,6 +11,15 @@ type PropMap<T> = { [prop: string]: T };
  */
 export interface MongoStep {
   [propName: string]: any;
+}
+
+/** transform a 'concatenate' step into corresponding mongo steps */
+function transformConcatenate(step: Readonly<ConcatenateStep>): MongoStep {
+  const concatArr: any[] = [{ $convert: { input: $$(step.columns[0]), to: 'string' } }];
+  for (const colname of step.columns.slice(1)) {
+    concatArr.push(step.separator, { $convert: { input: $$(colname), to: 'string' } });
+  }
+  return { $addFields: { [step.new_column_name]: { $concat: concatArr } } };
 }
 
 /** transform a 'convert' step into corresponding mongo steps */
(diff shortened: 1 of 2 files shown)


davinov (commented):
Hi @CharlesRngrd , and thanks a lot for this contribution!

I have one comment and one question :
1. On the code itself: It'll be best to avoid duplicating the logic in the `transformConcatenate` function. Here is a quick way to do it by exporting it and allowing to override the functio to transform each column name: https://github.com/ToucanToco/weaverbird/commit/1cab58c982bf1680963488adcae75b6ecbd99afe  
2. On the $convert use: do you know if this impact positively or negatively the performance of a query? I mean: before this patch, most concat queries do not have $convert, then they do. Will it slow them? If yes, do we have an idea of by which amount? This will help to know if this a good idea to merge this as is, or if we should only add the convert step for columns that needs it (and it'll not be that easy to figure out how, as the type of columns is not necessarily known when translating a query ^^).

CharlesRngrd:
> Hi @CharlesRngrd , and thanks a lot for this contribution!
> 
> I have one comment and one question :
> 
> 1. On the code itself: It'll be best to avoid duplicating the logic in the `transformConcatenate` function. Here is a quick way to do it by exporting it and allowing to override the functio to transform each column name: [1cab58c](https://github.com/ToucanToco/weaverbird/commit/1cab58c982bf1680963488adcae75b6ecbd99afe)
> 2. On the $convert use: do you know if this impact positively or negatively the performance of a query? I mean: before this patch, most concat queries do not have $convert, then they do. Will it slow them? If yes, do we have an idea of by which amount? This will help to know if this a good idea to merge this as is, or if we should only add the convert step for columns that needs it (and it'll not be that easy to figure out how, as the type of columns is not necessarily known when translating a query ^^).

In production queries, we only convert strings to strings. I guess this is not costing to much time and memory.
Do you have some tools to mesure that loss of performance ?

src/lib/translators/mongo4.ts
```diff
@@ -13,6 +13,15 @@ export interface MongoStep {
   [propName: string]: any;
 }
 
+/** transform a 'concatenate' step into corresponding mongo steps */
+function transformConcatenate(step: Readonly<ConcatenateStep>): MongoStep {
+  const concatArr: any[] = [{ $convert: { input: $$(step.columns[0]), to: 'string' } }];
+  for (const colname of step.columns.slice(1)) {
+    concatArr.push(step.separator, { $convert: { input: $$(colname), to: 'string' } });
+  }
```
davinov, line 21: It should be possible to avoid having to duplicate this from mongo 3.6 translator, I'll make you a small PR to demonstrate this

davinov (approved):
Hi @CharlesRngrd 
I've made a quick study to see if concert affects performance when used on string properties.
I took a mongo atlas free insatance and loaded the sample data. I used the sample_training.routes collection (10k documents), and ran two aggregations timed in a notebook:
Screenshot from 2022-01-12 11-09-36

This seems to demonstrate that the perfomance is either not affected, or too small to be measured reliably :tada: 

So we can merge this and be confident the performance of our concatenate steps will stay sharp :muscle:

src/lib/translators/mongo4.ts
davinov: See https://github.com/CharlesRngrd/weaverbird/pull/3

davinov:
Fixed the duplication in #1248 :)
Thanks again for this contrib :+1:

PR #999, opened 2021-08-19
feat(cumsum step): cumsum multiple value columns

CharlesRngrd:
Hi Toucan,

As a conceptor, I need to cumulate multiple columns.
Here is my PR to enable that ;)

Have a nice day,

Charles


src/lib/translators/mongo.ts
```diff
@@ -595,24 +595,58 @@ function transformArgmaxArgmin(step: Readonly<S.ArgmaxStep> | Readonly<S.ArgminS
...
+            $sum: {
+              $slice: [$$(step.valueColumn), { $add: ['$_VQB_INDEX', 1] }],
+            },
+          },
+          _vqbArray: 1,
+        },
+      },
+      { $replaceRoot: { newRoot: { $mergeObjects: ['$_vqbArray', '$$ROOT'] } } },
+      { $project: { _vqbArray: 0 } },
+    ];
```
davinov, line 624: That'll definitly work, but you'll have an easier time converting the `step.valueColumn` into the new syntax and passing into the same logic after.
Something like you did when reading the python pipeline: https://github.com/ToucanToco/weaverbird/pull/999/files#diff-2fddffd5011b87bd7d07cfcd8ab0a50c9eab64afcbeb1a9c90b3fb216671e5a5R18-R20

Unit tests will ensure both cases results are coherent :)

src/lib/labeller.ts
```diff
@@ -174,7 +174,16 @@ class StepLabeller implements StepMatcher<string> {
...
+    // For retrocompatibility with old configurations
+    if (step.valueColumn) {
+      return `Compute cumulated sum of "${step.valueColumn}"`;
+    }
+
+    if (step.toCumSum.length === 1) {
+      return `Compute cumulated sum of "${step.toCumSum[0][0]}"`;
+    } else {
+      return `Compute cumulated sum of ${formatMulticol(step.toCumSum.map(a => a[0]))}`;
+    }
```
davinov, line 186: It would be nice to add test case here: https://github.com/ToucanToco/weaverbird/blob/master/tests/unit/labeller.spec.ts#L668
for this new case

davinov (commented):
Thank you for this contrib :+1: 
Here are a few comments, nothing big! Congrats on this new contribution :)

davinov (changes requested):
Thanks for the modifications!
I see yet a few things before we can merge:
- a reference to `step.newcolumn` will make the SQL translation fail (see below)
- it's good to have updated the tests, yet we should keep at least one with the legacy syntax to make sure we're not gonna break it later on


You're nearly there :muscle: thanks!

CharlesRngrd:
@davinov : I updated the tests.
Me vs Pydantic: 1-0 ;)

davinov:
I've added a few commits and reopoened the PR here: 1246
Thanks for this nice feature :+1:

PR #1008, opened 2021-08-20
feat(ui): select columns from the right dataset in the join step

CharlesRngrd:
Hi Toucan,

Columns are usualy not the same between the left and right dataset.
So here is my PR to allow the user to select the right column from a list.

Have a nice day,

Charles

### DIFF
--- a/src/store/actions.ts
+++ b/src/store/actions.ts
@@ -106,6 +106,38 @@ class Actions {
     dispatch('updateDataset');
   }
 
+  async selectRightColumnNames(
+    { commit, state }: ActionContext<VQBState, any>,
+    { rightPipelineLabel }: { rightPipelineLabel: string },
+  ) {
+    if (rightPipelineLabel != '') {
+      let activePipeline: Pipeline = [];
+
+      if (rightPipelineLabel in state.pipelines) {
+        activePipeline = state.pipelines[rightPipelineLabel];
+      } else {
+        activePipeline = [
+          {
+            name: 'domain',
+            domain: rightPipelineLabel,
+          },
+        ];
+      }
+
+      const response = await state.backendService.executePipeline(
+        activePipeline,
+        state.pipelines,
+        1,
+        0,
+      );
+      if (response.data) {
+        commit('setRightColumnNames', {
+          rightColumnNames: response.data.headers.map(col => col.name),
+        });
+      }
+    }
+  }
+
   deleteSteps(
     { commit, dispatch }: ActionContext<VQBState, any>,
     { indexes }: { indexes: number[] },
--- a/src/store/state.ts
+++ b/src/store/state.ts
@@ -44,6 +44,11 @@ export interface VQBState {
   pipelines: { [name: string]: Pipeline };
 
   /**
+   * the right column names to be joined.
+   */
+  rightColumnNames: string[];
+
+  /**
    * object used to fill an edit step form
    */
   stepFormInitialValue?: object;
@@ -130,6 +135,7 @@ export function emptyState(): VQBState {
         totalCount: 0,
       },
     },
+    rightColumnNames: [],
     domains: [],
     currentStepFormName: undefined,
     stepFormInitialValue: undefined,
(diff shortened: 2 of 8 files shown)


src/store/actions.ts
```diff
@@ -105,6 +105,38 @@ class Actions {
...
+        activePipeline,
+        state.pipelines,
+        1,
+        0,
+      );
+      if (response.data) {
+        commit('setRightDataset', {
+          rightDataset: addLocalUniquesToDataset(response.data),
+        });
+      }
```
davinov, line 136: `addLocalUniquesToDataset` have the purpose of creating a list of unique values for each column.
This is useful to display these values for filtering, in the column menu.

As we don't display the right dataset in a data table, nor want to filter it, I think you can skip this step.

src/store/actions.ts
```diff
@@ -105,6 +105,38 @@ class Actions {
...
+      if (rightPipelineLabel in state.pipelines) {
+        activePipeline = state.pipelines[rightPipelineLabel];
+      } else {
+        activePipeline = [
+          {
+            name: 'domain',
+            domain: rightPipelineLabel,
+          },
+        ];
+      }
```
davinov, line 124: Good to have thought of the two cases: a domain and another pipeline :100:

src/store/state.ts
```diff
@@ -43,6 +43,11 @@ export interface VQBState {
    */
   pipelines: { [name: string]: Pipeline };
 
+  /**
+   * the right dataset to be joined.
+   */
+  rightDataset: DataSet;
```
davinov, line 49: I think having the `rightDataset` will matter only during the join step, when one must choose the column of the right dataset to join on.
So keepjng that info in the store may not be useful, as no other component will ever need it.

My recomendation, to keep the data as close as where we need it, would be to only keep the `rightDataset` in the local state (`data`) of the `JoinStepForm`.

What do you think of this idea?

src/store/actions.ts
```diff
@@ -105,6 +105,38 @@ class Actions {
...
+
+      const response = await state.backendService.executePipeline(
+        activePipeline,
+        state.pipelines,
+        1,
```
davinov, line 129: And another good idea to fetch only one row :+1:

tests/unit/store.spec.ts
```diff
@@ -701,6 +701,54 @@ describe('mutation tests', () => {
...
+      ],
+      {
+        coco_l_asticot: [],
+        dataset1: [],
+        dataset2: [],
+      },
+      1,
+      0,
+    );
+  });
```
davinov, line 750: I think it'll be more useful to have test cases for the scenarios inside the `JoinStepForm`:
- when someone choose a right dataset, we should fetch the first row fo the selected right dataset and fill the `rightOn` options with it
- when someone changes the right dataset, we should update these options
- when someone edits an existing JoinStep, we should retrieve the right dataset right away

This may enlights some intermediate cases:
- do we need a spinner on the rightOn dropdown while we wait for the results?
- if we can't fetch the right dataset, should we default on a free input?
- what if someone selects a column, then changes dataset and this column doesn't exist in the newly selected dataset?

davinov (changes requested):
Thanks for this initiative @CharlesRngrd 
It'd be a very useful feature!

Here are some recommendations and questions. Don't hesitate to reach out to get help or discuss the feedback :)

tests/unit/store.spec.ts
CharlesRngrd: Hi @davinov,

Thank you for your feedback.
Could you call me this afternoon ?

Have a nice day,

Charles

davinov: Hi! Today is a bit packed and I'm off monday, but Tuesday could work!
Let's find a time together on slack :)

davinov:
I've made a PR to address the requested changes: https://github.com/CharlesRngrd/weaverbird/pull/1

davinov:
Merge + modifs in  #1249

PR #1037, opened 2021-08-30
refactor(popover): prefix css class to avoid collision

austil:
The popover is using a simple `.popover` class.
This name is so generic that collision have been reported by customers using Weaverbird via Toucan Embeds.
Let's avoid that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


austil:
[Intégration VQB en webcomponent](https://trello.com/c/Ku7tIYIm/2029-int%C3%A9gration-vqb-en-webcomponent)

davinov (approved):

PR #1041, opened 2021-08-30
feat: sql append step

raphaelvignes:
What: 
Adding the Append Step. 
This Step is a bit complex as it might have to resolve a provided pipeline and give a relevant subquery name based on index.
Moreover, the step as to handle metadata in a specific way: 
When appending, use first dataset columns names & fill with first available column name from appended dataset if needed. 
In case when appended dataset as not enough column, add a NULL column.


Sanix-Darker (approved):
Looks good to me !

PR #1047, opened 2021-09-01
feat(font-awesome): use svg icons

alice-sevin:
We transform all (possible) font-awesome css icons to svg icons

**Adds and updates**

Add font awesome svg packages
Add a list of all icons used in app (prefixed with style 'far' if there is duplicated icons with different styles) (assets file)
Load icons in main and storybook one time at init (lib file)
Add a new component FAIcon to display a font-awesome using the referent (assets file) icon name

Replaced all available occurencies of `<i>` with `<FAIcon/>`

:warning:  **Limits and important updates**
Some icons can't be modified (multiselect) because it's an outside package
Some v-html needs have been modified to utf8 icons because v-html doesnt support vue component insertion (DateRangeInput)
Some components have been modified from functional to classes to enable child component usage (parent and listeners keeped for now, we will modified logic in another PR because a lot of file are impacted ex: StepFormHeader is called in every step, remove the $parent call will impact a lot of files)

Some css properties have been keeped with usage of fa class (using only a ::v-deep to access FAIcon component), we may clean it in another PR

### DIFF
--- a/src/assets/FA-ICONS.ts
+++ b/src/assets/FA-ICONS.ts
@@ -62,8 +62,6 @@ const FA_ICONS = {
   'far trash-alt': faTrashAltFar,
 };
 
-/* istanbul ignore next */
 export const FA_ICONS_PACK = Object.values(FA_ICONS);
-/* istanbul ignore next */
 export const FA_ICONS_NAMES = Object.keys(FA_ICONS);
 export type FAIconName = typeof FA_ICONS_NAMES[number];
--- a/src/assets/FA-ICONS.ts
+++ b/src/assets/FA-ICONS.ts
@@ -1,0 +1,5 @@
+import { library } from '@fortawesome/fontawesome-svg-core';
+import { far } from '@fortawesome/free-regular-svg-icons'; // TODO: find a better way that including all icons
+import { fas } from '@fortawesome/free-solid-svg-icons'; // TODO: find a better way that including all icons
+
+library.add(far, fas);
--- a/src/assets/FA-ICONS.ts
+++ b/src/assets/FA-ICONS.ts
@@ -1,4 +1,57 @@
-const FA_ICONS = {};
+import { faCalendar, faTrashAlt } from '@fortawesome/free-regular-svg-icons';
+import {
+  faAngleDown,
+  faAngleLeft,
+  faAngleRight,
+  faArrowRight,
+  faCalculator,
+  faCalendarAlt,
+  faCheck,
+  faCodeBranch,
+  faCog,
+  faDrawPolygon,
+  faExclamation,
+  faExclamationCircle,
+  faExclamationTriangle,
+  faFilter,
+  faFont,
+  faGripVertical,
+  faMagic,
+  faObjectGroup,
+  faPlus,
+  faQuestionCircle,
+  faSearch,
+  faTimes,
+  faTrash,
+} from '@fortawesome/free-solid-svg-icons';
+
+const FA_ICONS = {
+  'angle-down': faAngleDown,
+  'angle-left': faAngleLeft,
+  'angle-right': faAngleRight,
+  'arrow-right': faArrowRight,
+  calculator: faCalculator,
+  calendar: faCalendar,
+  'calendar-alt': faCalendarAlt,
+  check: faCheck,
+  'code-branch': faCodeBranch,
+  cog: faCog,
+  'draw-polygon': faDrawPolygon,
+  exclamation: faExclamation,
+  'exclamation-circle': faExclamationCircle,
+  'exclamation-triangle': faExclamationTriangle,
+  filter: faFilter,
+  font: faFont,
+  'grip-vertical': faGripVertical,
+  magic: faMagic,
+  'object-group': faObjectGroup,
+  plus: faPlus,
+  'question-circle': faQuestionCircle,
+  search: faSearch,
+  times: faTimes,
+  trash: faTrash,
+  'trash-alt': faTrashAlt,
+};
 
 export const FA_ICONS_PACK = Object.values(FA_ICONS);
 export const FA_ICONS_NAMES = Object.keys(FA_ICONS);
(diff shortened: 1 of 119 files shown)


tests/unit/step-form-header.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    it('should display the backend error', () => {
+      expect(wrapper.find('.step-edit-form__error').exists()).toBe(true);
+    });
+
+    it('should display the correct backend error message', () => {
+      expect(wrapper.find('.step-edit-form__error').text()).toBe('error');
+    });
+  });
+});
```
davinov, line 52: Thanks for adding these tests!

src/assets/FA-ICONS.ts
```diff
@@ -0,0 +1,66 @@
+import {
```
davinov, line 1: Maybe we should add a comment here to indicate that every icon from font awesome should be imported here?

src/components/QueryBuilder.vue
```diff
@@ -173,7 +175,7 @@ export default class QueryBuilder extends Vue {
     text-decoration: underline;
   }
 
-  .fas.fa-question-circle {
+  ::v-deep .fa-question-circle {
```
davinov, line 178: Instead of using the v-deep, maybe you can add a class to the FAIcon component above ?
`<FAIcon class="documentation-help__icon" icon="question-circle" />`
and then `.documentation-help__icon { ...}` in the stylesheet

src/components/FAIcon.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <FontAwesomeIcon :icon="iconWithStyle" />
```
davinov, line 2: Should we use aria-hidden on these SVGs? or is it OK not to?

Edit: the docs pointed me that it's automatic! https://fontawesome.com/v5.15/how-to-use/on-the-web/other-topics/accessibility#svg-decorative So nothing to change here

tests/unit/pipeline.spec.ts
```diff
@@ -159,7 +160,7 @@ describe('Pipeline.vue', () => {
       ];
       const store = setupMockStore(buildStateWithOnePipeline(pipeline));
       dispatchSpy = jest.spyOn(store, 'dispatch');
-      wrapper = mount(PipelineComponent, { store, localVue });
+      wrapper = mount(PipelineComponent, { store, localVue, stubs: { FAIcon: '<div />' } });
```
davinov, line 163: Tiny detail: maybe it's best to stub this component with an empty <svg/> as it'll more close to the actual result

davinov: Also, to avoid having to mock FAIcon repeatidly, you can write a stub component in a `__mocks__` folder just beside FAIcon.vue, and then you'll only have to write once `jest.mock('@/components/FAIcon.vue')` per test file.
And it may also ease following tests writing :)

src/components/DeleteConfirmationModal.vue
```diff
@@ -1,13 +1,9 @@
-<template functional>
```
davinov, line 1: Note for future readers: we could have make the FAIcon component function (as the FontAwesomeIcon is functional) but we decided not ot use them as they will be no more in vue 3 https://v3.vuejs.org/guide/migration/functional-components.html

src/components/Step.vue
```diff
@@ -289,7 +294,7 @@ export default class Step extends Vue {
   cursor: move;
 }
 
-.query-pipeline-step__action i {
+.query-pipeline-step__action ::v-deep .fa-cog {
```
davinov, line 297: Same here, you may be able ot use a class in the FAIcon component instead

src/lib/dates.ts
```diff
@@ -42,8 +42,7 @@ export const DEFAULT_DURATIONS: DurationOption[] = [
   { label: 'Days ago', value: 'day' },
 ];
 
-export const CUSTOM_DATE_RANGE_LABEL_SEPARATOR =
-  '<i aria-hidden="true" class="fa fa-arrow-right"></i>';
+export const CUSTOM_DATE_RANGE_LABEL_SEPARATOR = '<span>&#8594;</span>'; //use utf8 arrow symbol
```
davinov, line 45: A simpler refactoring suggestion:
we could just use space, arrow, space (`&nbsp;&#8594;&nbsp;`)
and remove the styling around the arrow, and drop the v-html

It would differ minimally from the mock-up (the color the arrow would be the same than the dates) though. Do you think it's worth the simplification?

davinov, line 45:
```suggestion
export const CUSTOM_DATE_RANGE_LABEL_SEPARATOR = '<span>&#8594;</span>'; // arrow symbol: →
```

davinov (commented):
Thanks! Huge work !

Here are a few comments but nothing really blocking. Tell me which ones you'd like to implement and feel free to dismiss others.

src/components/Step.vue
alice-sevin: As said in description, will handle all the v-deep .fa logic in another PR, this one is too long already

src/components/FAIcon.vue
alice-sevin: Not necessary, it's done automatically yes

src/lib/dates.ts
alice-sevin: Let's try it :wink:

alice-sevin: Tried it but we cant' because the arrow is 'rikiki' so we need some style :cry:

tests/unit/pipeline.spec.ts
alice-sevin: Je l'ai dev ici https://github.com/ToucanToco/weaverbird/pull/1047/commits/d3ba6f1f6783ec2a5e86dc1e6cfdd8446842c9ad, par contre j'ai du l'importer dans la quasi totalité des fichier (mais ça clean toutes les console error qui trainaient surement, c'est pas plus mal. Pas besoin de tout review t'as plus de 60 fichiers d'impactées :sweat_smile: )

davinov (approved):
Thanks for the changes! Let's go for merge :)

PR #1050, opened 2021-09-01
fix: update sql_query_executor function

raphaelvignes:
* Update steps signature to receive one describer function & one executor function 

Backend was updated accordingly: https://github.com/ToucanToco/laputa/pull/4685


server/weaverbird/backends/sql_translator/metadata.py
```diff
@@ -183,6 +187,10 @@ def remove(self) -> TableMetadata:
         self.delete = True
         return self
 
+    def remove_column_alias(self, column_name: str) -> TableMetadata:
+        self.columns[column_name].remove_alias()
```
Sanix-Darker, line 191: Not sure but, maybe we need to raise an error here when a column doesn't exist ?

Sanix-Darker (commented):
Thanks for this fix, i left some comments !

server/weaverbird/backends/sql_translator/metadata.py
raphaelvignes: good one.

Sanix-Darker (approved):

PR #1053, opened 2021-09-02
fix(vqb): top-n-rows

Sanix-Darker:
# What :
- top-nrows step was not displaying other columns even without the group-by

# Before the fix : 
Peek 2021-09-02 17-28

# After the fix :
fix-top-n

### DIFF
--- a/server/weaverbird/backends/sql_translator/steps/top.py
+++ b/server/weaverbird/backends/sql_translator/steps/top.py
@@ -62,7 +62,7 @@ def translate_top(
         query_name=query_name,
         transformed_query=f"""{query.transformed_query}, {query_name} AS"""
         f""" (SELECT {completed_fields}"""
-        f""" FROM {query.query_name} {group_by_query} ORDER BY {step.rank_on} {step.sort} LIMIT {step.limit}) """,
+        f""" FROM {query.query_name} {group_by_query} ORDER BY {step.rank_on} {step.sort} LIMIT {step.limit})""",
         selection_query=build_selection_query(
             query.metadata_manager.retrieve_query_metadata_columns(), query_name
         ),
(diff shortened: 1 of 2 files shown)


Sanix-Darker:
[(2) [Snowflake SQL VQB] - Top N rows step [3]](https://trello.com/c/lxERrYJ5/1178-2-snowflake-sql-vqb-top-n-rows-step-3)

raphaelvignes (approved):
Thanks for this fix!

server/weaverbird/backends/sql_translator/steps/top.py
```diff
@@ -40,20 +40,22 @@ def translate_top(
...
-    # to handle upper and lower cases
-    # we refresh query metadatas with columns to not remove
-    cols = [g.upper() for g in step.groups] + [step.rank_on.upper()]
-    cols_to_remove = [
-        col
-        for col in query.metadata_manager.retrieve_query_metadata_columns_as_list()
-        if col not in cols
-    ]
-    query.metadata_manager.remove_query_metadata_columns(cols_to_remove)
+    # We remove other columns not in the group-by
```
hachichaud, line 47: In the pandas implementation, columns arent removed.
So you keep all columns of the top rows.

hachichaud (dismissed):
even with groupby, you dont remove columns

server/weaverbird/backends/sql_translator/steps/top.py
raphaelvignes: Good point, because if we use `Group By` in SQL we'll lose columns anway. So to have the same behavior as in Pandas, we'll need to perform an auto-join a remove non aggregated columns.

PR #1054, opened 2021-09-02
fix(sql append): use union all in append step

raphaelvignes:
To have the same behavior as the Pandas append step, we need to use UNION ALL keyword in Snowflake instead of just Union.


Sanix-Darker (approved):
Looks Good to Me !

raphaelvignes:
[[Snowflake SQL VQB] - Append Step (5) [5]](https://trello.com/c/r3d6z5hJ/1188-snowflake-sql-vqb-append-step-5-5)

PR #1060, opened 2021-09-07
feat(sql vqb): add fillna step

raphaelvignes:
https://user-images.githubusercontent.com/71276735/132383762-40b9bbc1-1acc-4b4f-a734-27e72fbbc167.mp4

* Add sql vqb fillna step 
weird behavior with dates, as the front end directly fills null date values.

### DIFF
--- a/server/weaverbird/backends/sql_translator/steps/fillna.py
+++ b/server/weaverbird/backends/sql_translator/steps/fillna.py
@@ -1,5 +1,6 @@
 from weaverbird.backends.sql_translator.steps.utils.query_transformation import (
     build_selection_query,
+    sanitize_input,
 )
 from weaverbird.backends.sql_translator.types import (
     SQLPipelineTranslator,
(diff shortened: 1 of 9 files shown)


raphaelvignes:
[[Snowflake SQL VQB] - Fill null step (3)](https://trello.com/c/5OLwXgN9/1212-snowflake-sql-vqb-fill-null-step-3)

server/weaverbird/backends/sql_translator/steps/fillna.py
```diff
@@ -0,0 +1,40 @@
...
+    sql_query_executor: SQLQueryExecutor = None,
+    sql_translate_pipeline: SQLPipelineTranslator = None,
+    subcall_from_other_pipeline_count: int = None,
+) -> SQLQuery:
+    query_name = f'FILLNA_STEP_{index}'
+
+    unchanged_colums = query.metadata_manager.retrieve_query_metadata_columns_as_str(
+        columns_filter=step.columns
+    )
+    fill_value = f"'{step.value}'" if isinstance(step.value, str) else step.value
```
Sanix-Darker, line 29: This can be problematic if we have in `step.value` a single quotes for the final query...
I don't know... but it could be a good idea todo some replace here... just to be sure ?
```python
# we escape weird quotes here
step.value = step.value.replace('"', "'").replace("'", "\\'")

fill_value = f"'{step.value}'" if isinstance(step.value, str) else step.value
```

if this rmq is relevant... i thing we need to add tests for that behaviour too ? 🤔

Sanix-Darker (approved):
Thanks for this step !
I leave a comment for the escape of quotes !

Not sure if it's realy relevant but for now, the whole PR looks Good !

server/weaverbird/backends/sql_translator/steps/fillna.py
raphaelvignes: :+1:  good point. 
I'll add an util for that.

PR #1063, opened 2021-09-08
feat(sql custom): sql custom step

raphaelvignes:
Custom Step for sql vqb.

https://user-images.githubusercontent.com/71276735/132564769-5e0ff902-934c-4ebf-90c4-dc5a02dfadbd.mp4


raphaelvignes:
[[Snowflake SQL VQB] Discovery - Custom step POC (3)](https://trello.com/c/J4qCslkl/1236-snowflake-sql-vqb-discovery-custom-step-poc-3)

server/weaverbird/pipeline/steps/custom.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import Field
+
+from weaverbird.pipeline.steps.utils.base import BaseStep
+
+
+class CustomStep(BaseStep):
```
Sanix-Darker, line 6: For me i suggest `CustomSqlStep` instead of `CustomStep`.... as the class name, because this can be confusing with the customStep alreadyavailable on the frontend but with mongo... i don't know 🤔 same for `CUSTOM_STEP` to `CUSTOMSQL_STEP` ?
What do you think ?

server/weaverbird/backends/sql_translator/steps/custom.py
```diff
@@ -0,0 +1,52 @@
...
+
+    log.debug(
+        '############################################################'
+        f'query_name: {query_name}\n'
+        '------------------------------------------------------------'
+        f'step: {step}\n'
+        f'query.transformed_query: {query.transformed_query}\n'
+        f'query.metadata_manager.query_metadata: {query.metadata_manager.retrieve_query_metadata()}\n'
+    )
+    injected_query = step.query.replace('##PREVIOUS_STEP##', query.query_name).replace(';', '')
```
Sanix-Darker, line 36: - Maybe we should make a trim for extras spaces in this incoming query
- Maybe we should check if the length of the query is upper than a minimal value before proceed it, otherwise, we just replay the precedent step ? 🤔 
- We should add a validator here i think... even if it's not a perfect one... (for the snowflake synthax, this is a [source code](https://gist.github.com/tamiroze/dc0bdd596ed2c6b70fe921061401e739) for converting normal sql to snowflake sql... maybe we can take some regex here to add to our validator ? 🤔  )

src/components/stepforms/CustomSqlStepForm.vue
```diff
@@ -0,0 +1,56 @@
...
+@Component({
+  name: 'custom-sql-step-form',
+  components: { CodeEditorWidget },
+})
+export default class CustomSqlStepForm extends BaseStepForm<CustomSqlStep> {
+  stepname: PipelineStepName = 'customsql';
+
+  @Prop({
+    type: Object,
+    default: () => ({ name: 'customsql', query: '' }),
```
Sanix-Darker, line 42: To help the user.... we can add a placeholder here... something like :
```
SELECT * FROM  ##PREVIOUS_STEP##
```
Just for the user to see how it can be presented.

src/lib/translators/snowflake.ts
```diff
@@ -128,4 +131,25 @@ export class SnowflakeTranslator extends BaseTranslator {
...
     return step;
   }
+
+  validate(customEditedStep: S.CustomSqlStep): ValidationError[] | null {
+    try {
+      if (
+        !customEditedStep.query.toLowerCase().includes('select') &&
+        !customEditedStep.query.toLowerCase().includes('##previous_step##')
+      ) {
+        throw new Error('Invalid Query: should use SELECT and ##PREVIOUS_STEP## keywords');
```
Sanix-Darker, line 141: 👍🏾

Sanix-Darker (commented):
Thanks for this PR, i left some comments !

server/weaverbird/pipeline/steps/custom.py
raphaelvignes: Good remark, I update that.

server/weaverbird/backends/sql_translator/steps/custom.py
raphaelvignes: 1. Yes, good idea I'll sanitize more 
2. This check is performed by the frontend's validation
3. I Guess we'll handle that by an enhanced frontend validation

src/components/stepforms/CustomSqlStepForm.vue
raphaelvignes: yes, good idea.

Sanix-Darker (approved):

hachichaud (approved):
ok for the front-end and using a specific widget for sql, in case we want to add more sql logic later on.
However I think we can still add some customization through the codeeditor config (see doc here https://github.com/ToucanToco/weaverbird/blob/master/docs/_docs/tech/custom-code-editor.md#L15:L15

and tucana project https://github.com/ToucanToco/tucana/blob/master/src/scripts/utils/weaverbird.js#L74:L75)

PR #1066, opened 2021-09-14
fix(dates): use default toString

austil:
instead UTC or ISO (json) ones,
as we now overload the default Date.toString
beforehand to globaly control how dates are
displayed inside Toucan Toco

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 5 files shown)


austil:
[Timezoneless dates : affichage dans Weaverbird & option "cast dates to string" des requester](https://trello.com/c/E7NfeJA2/2509-timezoneless-dates-affichage-dans-weaverbird-option-cast-dates-to-string-des-requester)

davinov (approved):
:+1:

PR #1067, opened 2021-09-14
fix(pandas): support millisecond timestamps instead of nanosecond

austil:
I spotted a inconsistency between the Mongo & Pandas backends : when converting dates to int, Mongo return millisecond timestamps while Pandas return nanosecond ones.

```
input :  2020-01-01T00:00:00.000Z
mongo  : 1577836800000
pandas : 1577836800000000000
```

This conversion can be useful for applying a timezone offset.
I chose to align Pandas with Mongo as it require less changes than supporting second timestamp on both & nanosecond timestamp are clearly not part of our use cases.


austil:
[Timezoneless dates : affichage dans Weaverbird & option "cast dates to string" des requester](https://trello.com/c/E7NfeJA2/2509-timezoneless-dates-affichage-dans-weaverbird-option-cast-dates-to-string-des-requester)

server/weaverbird/backends/pandas_executor/steps/todate.py
```diff
@@ -10,5 +10,8 @@ def execute_todate(
     domain_retriever: DomainRetriever = None,
     execute_pipeline: PipelineExecutor = None,
 ) -> DataFrame:
-    datetime_serie = to_datetime(df[step.column], format=step.format, errors='coerce')
+    timestamp_unit = 'ms' if df[step.column].dtype == 'int64' else None
+    datetime_serie = to_datetime(
+        df[step.column], format=step.format, errors='coerce', unit=timestamp_unit
+    )
```
luc-leonard, line 16: Je me demande si y a pas moyen de factoriser ce code, et https://github.com/ToucanToco/weaverbird/pull/1067/files#diff-b4d0dcf9be156b4d95773586a88d698e7b1286daf8c1aab8f607ea4dbc78ae83R24-R27

austil: Sûrement :shrug: 
Aussi j'ai l'impression que la step "convert" vers dates n'est jamais utilisée car la UI n'a l'air de pointer que vers la step "todate". Les deux sont un peut redondante d'ailleurs. La seule différence c'est que "todate" accepte une string pour spécifier du format de la source en plus.

server/tests/steps/test_convert.py
```diff
@@ -19,3 +19,46 @@ def test_benchmark_convert(benchmark):
...
+    """
+    df = pandas.DataFrame(
+        {
+            'date': [pandas.to_datetime('21/03/2006 21:50:00')],
+        }
+    )
+    step = ConvertStep(name='convert', columns=['date'], data_type='integer')
+    result = execute_convert(step, df)
+    timestamp_result = result.at[0, 'date']
+    assert str(timestamp_result) == '1142977800000'
```
davinov, line 64: Good! Could you also create the json files so these tests can easily be executed by other and future backends?
See the "how to create tests fixtures" doc

server/weaverbird/backends/pandas_executor/steps/utils/cast.py
```diff
@@ -6,10 +6,13 @@ def cast_to_float(s: Series) -> Series:
...
 def cast_to_int(s: Series) -> Series:
-    s = cast_to_float(s)
+    s_float = cast_to_float(s)
     # Since .astype(int) would fail on NaN values, treat them separately:
-    s_integer = s[s.notna()].astype(int)
-    s_nan = s[s.isna()]
+    s_integer = s_float[s_float.notna()].astype(int)
+    s_nan = s_float[s_float.isna()]
+    # Pandas operate on NanoSecond Timestamp but we want to output MilliSecond Timestamps
+    if s.dtype == 'datetime64[ns]':
```
davinov, line 14: Maybe we should us a startswith here, because if dates would arrive with timezone info, it would be reflected here I think
(https://pandas.pydata.org/docs/reference/api/pandas.DatetimeTZDtype.html#pandas.DatetimeTZDtype)

server/weaverbird/backends/pandas_executor/steps/utils/cast.py
```diff
@@ -18,7 +21,10 @@ def cast_to_str(s: Series) -> Series:
 
 
 def cast_to_datetime(s: Series) -> Series:
-    return to_datetime(s, errors='coerce')  # cast errors will result in NaT values
+    timestamp_unit = 'ms' if s.dtype == 'int64' else None
```
davinov, line 24: Not clear to me why only when the type is `int64`. What cases does this type represent and what other cases could happen?

davinov (changes requested):
Thanks! Good to have noted that :s !

Just a few notes:

server/weaverbird/backends/pandas_executor/steps/utils/cast.py
austil: This type represent the "converting timestamp to dates" case,other case

austil: Gonna try being a little more verbose and will add comments if necessary

server/tests/steps/test_convert.py
austil: Btw : some tests with nano second timestamp already exist but nothing broke on the CI, why ?

davinov (approved):
:+1: great, thanks for adding these fixtures

PR #1079, opened 2021-09-21
feat(sql/totals step): add totals step for sql translator

luc-leonard:


raphaelvignes (changes requested):
Merci pour cette step ! :tada: 
Il me semble qu'il manque deux choses: 
- l'ajout de la fixture de test pour le test d'intégration
- il faut aussi rajouter la step dans `src/lib/translators/snowflake.ts`

server/weaverbird/backends/sql_translator/steps/totals.py
```diff
@@ -0,0 +1,102 @@
...
+        for aggregation in aggregations
+        if aggregation.agg_function not in ['first', 'last']
+    ]:
+
+        for col, new_col in zip(aggregation.columns, aggregation.new_columns):
+            if aggregation.agg_function == 'count distinct':
+                aggregated_cols.append(f'COUNT(DISTINCT {col}) AS {new_col}')
+            else:
+                aggregated_cols.append(f'{aggregation.agg_function.upper()}({col}) AS {new_col}')
+    return aggregated_cols
```
Sanix-Darker, line 43: Instead of creating/using this function, it could be better to adapt the one in [utils/aggregations.py](https://github.com/ToucanToco/weaverbird/blob/master/server/weaverbird/backends/sql_translator/steps/utils/aggregation.py#L13) 
The actual function get as input an `aggregate step` and the `aggregate type`, i suggest you to adapt that function to get as input `aggregations` and `agg_type` instead !


That's mean, in line  51 of this file it should be `get_aggs_columns(step.aggregations)`, not `select_aggregate(step.aggregations)` and then adapt tests for that function too 🙏🏾  !

server/weaverbird/backends/sql_translator/steps/totals.py
```diff
@@ -0,0 +1,102 @@
...
+
+    group_sets = []
+    for combination in combinations(total_dimensions):
+        group_sets.append('(' + ','.join(combination) + ')')
+    group_sets.append('()')
+    group_by = step.groups + ['GROUPING SETS(' + (', '.join(group_sets)) + ')']
+
+    query = f"""SELECT {', '.join(selects)} FROM {from_table} GROUP BY {', '.join(group_by)};"""
+
+    return query
```
Sanix-Darker, line 63: Since this is an independent function, we need to make sure we control inputs and outputs for futures refactos/updates, so, can we have unit tests for it ?

server/weaverbird/backends/sql_translator/steps/totals.py
```diff
@@ -0,0 +1,102 @@
...
+from weaverbird.utils.iter import combinations
+
+
+def select_total_dimensions(total_dimensions: List[TotalDimension]) -> List[str]:
+    selects = []
+    for dimension in total_dimensions:
+        selects.append(
+            f"CASE WHEN GROUPING({dimension.total_column}) = 0 THEN {dimension.total_column} ELSE '{dimension.total_rows_label}' END"
+        )
+    return selects
```
Sanix-Darker, line 27: Since this function is used in only one place, and it's just a loop, is it possible to merge that loop directly inside the function `make_totals_query` at line 46 ? 🤔 
That will prevent to have to write unit tests for this small function that is use only in one place !

server/weaverbird/backends/sql_translator/steps/totals.py
```diff
@@ -0,0 +1,102 @@
...
+        f'query_name: {query_name}\n'
+        '------------------------------------------------------------'
+        f'step: {step}\n'
+        f'query.transformed_query: {query.transformed_query}\n'
+        f'query.metadata_manager.query_metadata: {query.metadata_manager.retrieve_query_metadata()}\n'
+    )
+
+    new_query = SQLQuery(
+        query_name=query_name,
+        transformed_query=f"{query.transformed_query}, {query_name} AS({make_totals_query(step, query.query_name)})",
```
Sanix-Darker, line 89:
```suggestion
        transformed_query=f"{query.transformed_query}, {query_name} AS ({make_totals_query(step, query.query_name)})",
```

server/tests/steps/sql_translator/test_totals.py
```diff
@@ -0,0 +1,48 @@
...
+CASE WHEN GROUPING(PRODUCT) = 0 THEN PRODUCT ELSE 'All products' END,
+SUM(VALUE_1) AS VALUE_1-SUM, SUM(VALUE_2) AS VALUE_2-SUM, AVG(VALUE_1) AS VALUE_1-AVG, YEAR
+FROM SELECT_STEP_0
+GROUP BY
+YEAR,
+GROUPING SETS((COUNTRY), (PRODUCT), (COUNTRY,PRODUCT), ());""".replace(
+            '\n', ' '
+        )
+        + ')'
+    )
```
Sanix-Darker, line 46: In unit tests for the VQB we try to make the output as readable, explicit and clear as possible, that's why is suggest:
- to remove the `.replace` function
- write the string output of `query.transformed_query` in the most explicit and readable way

```suggestion
    expected_query = (
        'WITH SELECT_STEP_0 AS (SELECT * FROM products), TOTALS_STEP_1 AS('
        'SELECT CASE WHEN GROUPING(COUNTRY) = 0 THEN COUNTRY ELSE \'All countries\' END,'
        'CASE WHEN GROUPING(PRODUCT) = 0 THEN PRODUCT ELSE \'All products\' END,'
        'SUM(VALUE_1) AS VALUE_1-SUM, SUM(VALUE_2) AS VALUE_2-SUM, AVG(VALUE_1) '
        'AS VALUE_1-AVG, YEAR FROM SELECT_STEP_0 GROUP BY YEAR, '
        'GROUPING SETS((COUNTRY), (PRODUCT), (COUNTRY,PRODUCT), ()))'
    )
```

server/tests/steps/sql_translator/test_totals.py
```diff
@@ -0,0 +1,48 @@
...
+
+    assert new_query.transformed_query == expected_query
```
Sanix-Darker, line 48: There is 3 missing tests for this unit test:
- the selection query
- the new query name
- the metadatas columns


you can take [test_translate_argmax](https://github.com/ToucanToco/weaverbird/blob/69216b72fa464065c3dc2b725052a5fe61097aa8/server/tests/steps/sql_translator/test_argmax.py#L6) test as reference !

server/tests/steps/sql_translator/test_totals.py
```diff
@@ -0,0 +1,48 @@
+from weaverbird.backends.sql_translator.steps.totals import translate_totals
+from weaverbird.pipeline.steps import TotalsStep
+
+
+def test_totals(query):
```
Sanix-Darker, line 5:
```suggestion
def test_translate_totals(query):
```

Sanix-Darker (changes requested):
Thanks for this PR !
I left some comments and suggestions for you !
As @raphaelvignes said, we need also fixtures tests for each step, you can have some examples [here](https://github.com/ToucanToco/weaverbird/tree/69216b72fa464065c3dc2b725052a5fe61097aa8/server/tests/backends/fixtures)

server/weaverbird/backends/sql_translator/steps/totals.py
luc-leonard: they dont do the same thing. The function `get_aggs_columns` returns a list of tuple(old_column, new_column), not sure what to do with this.
This function returns a list of select string. 

It we wanted to refactor, there is the same code here:
https://github.com/ToucanToco/weaverbird/blob/69216b72fa464065c3dc2b725052a5fe61097aa8/server/weaverbird/backends/sql_translator/steps/utils/aggregation.py#L148

server/weaverbird/backends/sql_translator/steps/totals.py
luc-leonard: Its is not because a function exists that it should be unit tested imo. This function is scoped to the totals module, that is tested. I'm making this clearer by making it an inner func

server/weaverbird/backends/sql_translator/steps/totals.py
luc-leonard: Same answer as above. It is not because a function exists that it must be unit tested. I'd rather keep this function separated for better readability.

server/weaverbird/backends/sql_translator/steps/totals.py
luc-leonard: :+1:

server/tests/steps/sql_translator/test_totals.py
luc-leonard: :+1:

server/tests/steps/sql_translator/test_totals.py
luc-leonard: Thanks for the reference !

server/tests/steps/sql_translator/test_totals.py
luc-leonard: :+1:

server/weaverbird/backends/sql_translator/steps/totals.py
Sanix-Darker: ok

server/weaverbird/backends/sql_translator/steps/totals.py
Sanix-Darker: ok

server/weaverbird/backends/sql_translator/steps/totals.py
Sanix-Darker: ok

server/weaverbird/backends/sql_translator/steps/totals.py
davinov: Good point! The rule would be "if a function is exposed as a public API of the component unit, then it must be uni tested".
@luc-leonard one way to ensure everyone know it's not exposed would be to tag it "private' (even if the concept is just a convention: in python, one would prefix it by an underscore)

https://www.python.org/dev/peps/pep-0008/#designing-for-inheritance
> Public attributes should have no leading underscores.

Sanix-Darker:
[[Snowflake SQL VQB] - Add total rows (3)](https://trello.com/c/mBTcB4ys/1248-snowflake-sql-vqb-add-total-rows-3)

raphaelvignes (approved):
Thanks for this new step !

Sanix-Darker (approved):
Thanks for this Step !
Looks Good to Me !

PR #1087, opened 2021-09-27
feat(vqb): kill residuals tables + job to kill all old tables

Sanix-Darker:
# Why
- Because we were having too much tables when somebody was skipping/interrupting tests
- Too much residuals tables with unused datasets

# How
- Added a kind of job to delete all old tables with more than 24h of creation
- Automatically catch the keyboard interrupt and try to kill all tables created

### DIFF
--- a/server/tests/backends/sql_translator/test_snowflake_translator_steps.py
+++ b/server/tests/backends/sql_translator/test_snowflake_translator_steps.py
@@ -1,4 +1,5 @@
 import json
+import time
 from os import environ
 from random import randint
 from typing import Dict, NamedTuple, Optional, Union
@@ -22,6 +23,8 @@ WAREHOUSE = 'toucan_test'
 DATABASE = 'toucan_test'
 ROLE = 'toucan_test'
 SCHEMA = 'toucan_test'
+
+SNOWFLAKE_TABLES_TESTS = []
 
 
 # Update this method to use snowflake connection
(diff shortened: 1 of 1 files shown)


server/tests/backends/sql_translator/test_snowflake_translator_steps.py
```diff
@@ -91,62 +94,131 @@ def snowflake_query_executor(domain: str, query_string: str = None) -> Union[Tes
...
 
+def _drop_table(table_array: list):
+    global SNOWFLAKE_TABLES_TESTS
+
+    """
+    This method will drop tables
+    As given parameter, we have list of tables we want to drop
+
+    """
+    print("[x] Cleaning tables in progress, please wait...")
```
raphaelvignes, line 105: Thanks for this logging effort ! 
Can you use a logger instead of a print :pray:

server/tests/backends/sql_translator/test_snowflake_translator_steps.py
```diff
@@ -91,62 +94,131 @@ def snowflake_query_executor(domain: str, query_string: str = None) -> Union[Tes
...
+                json.dumps(spec['expected_sql']), orient='table'
+            )
+        else:
+            result_expected: pd.DataFrame = pd.read_json(
+                json.dumps(spec['expected']), orient='table'
+            )
 
-    assert_dataframes_equals(result_expected, result)
+        assert_dataframes_equals(result_expected, result)
+    except KeyboardInterrupt as es:
```
raphaelvignes, line 221: NICE :+1:

server/tests/backends/sql_translator/test_snowflake_translator_steps.py
```diff
@@ -91,62 +94,131 @@ def snowflake_query_executor(domain: str, query_string: str = None) -> Union[Tes
...
+
 # Translation from Pipeline json to SQL query
 @pytest.mark.parametrize('case_id, case_spec_file_path', test_cases)
 def test_sql_translator_pipeline(case_id, case_spec_file_path, get_engine):
-    # TO prevent conflict on snowflake tables when testing on multiple terminals
-    case_id = case_id.replace("/", "") + str(randint(1, 100000))
+    global SNOWFLAKE_TABLES_TESTS
+
+    # Let's clean some old residuals tables
+    clean_too_old_residuals_tables()
```
raphaelvignes, line 154: Thanks for this cleaning function ! 
Let's see if it doesn't slow too much the CI tests. 
If it's the case, perhaps we should try to do the same thing directly with a scheduled job in Snowflake.

raphaelvignes (changes requested):
Thanks for this improvement !
I have just one recommendation, can you use a logger instead of printing stuff to the standard output ? 
Thanks !

server/tests/backends/sql_translator/test_snowflake_translator_steps.py
Sanix-Darker: Updated !
Noticed, the actual function is been called multiple time in a same session tests
To fix that, i added a boolean checker that will allow the function to be executed only when it is False.

server/tests/backends/sql_translator/test_snowflake_translator_steps.py
Sanix-Darker: Updated !

server/tests/backends/sql_translator/test_snowflake_translator_steps.py
Sanix-Darker: Updated !

raphaelvignes (approved):

PR #1090, opened 2021-09-28
fix(sql pivot): clean selected cols

raphaelvignes:
When using SQL's `Pivot` step, we need to preselect columns from columns to keep, value column & pivot column to avoid keeping them in the transformation.


Sanix-Darker (approved):
Looks Good to me !

PR #1091, opened 2021-09-28
fix(sql pivot): handle pivot with illegal characters in value

raphaelvignes:
In SQL `Pivot` step we might have values with `-`, ` `, or starting with a number in value column to pass as column names.
This PR cleans such illegal characters.

### DIFF
--- a/server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
+++ b/server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
@@ -383,7 +383,9 @@ def get_query_for_date_extract(
 
 
 def sanitize_input(value: str) -> str:
-    return value.replace('"', '\\"').replace("'", "\\'")
+    if value[0].isdigit():
+        value = f'_{value}'
+    return value.replace('"', '\\"').replace("'", "\\'").replace('-', '_').replace(' ', '_')
 
 
 def build_aggregated_columns(aggregations):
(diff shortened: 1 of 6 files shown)


server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
```diff
@@ -386,6 +386,12 @@ def sanitize_input(value: str) -> str:
     return value.replace('"', '\\"').replace("'", "\\'")
 
 
+def sanitize_column_name(col: str) -> str:
```
Sanix-Darker, line 389: Maybe we should handle also:  
- spaces ?
- single/double quotes ?
- not alphanumerics characters

Sanix-Darker (changes requested):
Thanks for this PR, i left some comments !

server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
raphaelvignes: Thanks for this remark. 
* Whitespace is covered 
* It might be dangerous to replace quotes as it is valid to have quotes in column names : https://docs.snowflake.com/en/sql-reference/identifiers-syntax.html
* Alphanumeric caracters are valid if not at first position of the column name (but I improve the replacement)

I add a unit test :+1:

Sanix-Darker (approved):

PR #1095, opened 2021-10-05
feat(front): make date-range input fixed/relative tabs hideable

alice-sevin:
Add a new prop enableRelativeDate to date range input
I false, it will hide the tabs 'Fixed'/'Dynamic' and use Fixed tab as default tab
Add specific label to custom variable list custom button if enableRelativeDate is false

### DIFF
--- a/src/components/stepforms/widgets/DateComponents/DateRangeInput.vue
+++ b/src/components/stepforms/widgets/DateComponents/DateRangeInput.vue
@@ -12,11 +12,13 @@
           class="widget-date-input__editor-side"
           :availableVariables="availableVariables"
           :selectedVariables="selectedVariables"
+          :enableRelativeDate="enableRelativeDate"
           @selectCustomVariable="editCustomVariable"
           @input="selectVariable"
         />
         <div class="widget-date-input__editor-content" v-show="isCustom" ref="custom-editor">
           <Tabs
+            v-if="enableRelativeDate"
             class="widget-date-input__editor-header"
             :tabs="tabs"
             :selectedTab="selectedTab"
@@ -103,6 +105,9 @@ export default class DateRangeInput extends Vue {
   @Prop({ default: () => ({ start: '', end: '' }) })
   variableDelimiters!: VariableDelimiters;
 
+  @Prop({ default: () => true })
+  enableRelativeDate!: boolean;
+
   isEditorOpened = false;
   isEditingCustomVariable = false; // force to expand custom part of editor
   alignLeft: string = POPOVER_ALIGN.LEFT;
@@ -182,6 +187,10 @@ export default class DateRangeInput extends Vue {
     } else if (isRelativeDateRange(this.value)) {
       this.tabsValues.Dynamic = this.value;
       this.selectTab('Dynamic');
+    }
+    // force fixed tab by default if relative date is not enabled
+    if (!this.enableRelativeDate) {
+      this.selectTab('Fixed');
     }
   }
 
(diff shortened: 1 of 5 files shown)


tests/unit/date-range-input.spec.ts
```diff
@@ -334,6 +334,25 @@ describe('Date range input', () => {
...
+    });
+    it('should have no tabs', () => {
+      expect(wrapper.find('Tabs-stub').exists()).toBe(false);
+    });
+    it('should always use "Fixed" as selected tab', () => {
+      expect(wrapper.find('Tabs-stub').props().selectedTab).toBe('Fixed');
+    });
+    it('should pass disabled relative date disabled to custom variable list', () => {
+      expect(wrapper.find('CustomVariableList-stub').props().enableRelativeDate).toBe(false);
+    });
```
austil, line 353: Double Disabled :grimacing: 
Alternative text : 'should pass down disbled relative date props to custom variable list'

tests/unit/date-range-input.spec.ts
```diff
@@ -334,6 +334,25 @@ describe('Date range input', () => {
...
+        variableDelimiters: { start: '{{', end: '}}' },
+        enableRelativeDate: false,
+      });
+    });
+    it('should have no tabs', () => {
+      expect(wrapper.find('Tabs-stub').exists()).toBe(false);
+    });
+    it('should always use "Fixed" as selected tab', () => {
+      expect((wrapper.vm as any).selectedTab).toBe('Fixed');
+    });
```
austil, line 350: May be we should test what we expect to be displayed relative to this internal state ? Here the `RangeCalendar` ?

src/components/stepforms/widgets/DateComponents/DateRangeInput.vue
```diff
@@ -12,11 +12,13 @@
           class="widget-date-input__editor-side"
           :availableVariables="availableVariables"
           :selectedVariables="selectedVariables"
+          :enableRelativeDate="enableRelativeDate"
           @selectCustomVariable="editCustomVariable"
           @input="selectVariable"
         />
         <div class="widget-date-input__editor-content" v-show="isCustom" ref="custom-editor">
           <Tabs
+            v-if="enableRelativeDate"
```
austil, line 21: I see below that `RangeCalendar` is a child of this tabs component. So if we stop showing tabs this component will never show a `RangeCalendar` but only the `CustomVariableList` ?

austil (commented):
We're almost there ! But I have some doubt on the expected behavior.
Can you double check the storybook and provide a screenshot in the meantime ? :upside_down_face:

tests/unit/date-range-input.spec.ts
alice-sevin: :+1:

src/components/stepforms/widgets/DateComponents/DateRangeInput.vue
alice-sevin: Nop it's not a child, it's depedns only on selectedTab state property, if the component is display or not, so it has no impact to hide the tabs component, it will just remove navigation

tests/unit/date-range-input.spec.ts
alice-sevin: https://github.com/ToucanToco/weaverbird/pull/1095/commits/a628048ee9b6e6f5aeb05bc451fdb2635162ba9b

tests/unit/date-range-input.spec.ts
alice-sevin: https://github.com/ToucanToco/weaverbird/pull/1095/commits/a628048ee9b6e6f5aeb05bc451fdb2635162ba9b

austil (approved):
:green_circle:

PR #1103, opened 2021-10-06
feat(mongo): relative date support in filter step (mongo 5+) TCTC-1197

davinov:
Introduce a new `mongo50` translator, which knows how to translate `RelativeDate`s.

An important point: the format 
```{
  $match: {
    column: {
      $operator: value
    }
  }
}
```
does not work with `$dateAdd` expressions, so we had to use [`$expr`](https://docs.mongodb.com/manual/reference/operator/query/expr/).

Other note: in the `RelativeDate` type, there is no reference date, so I guessed this was going to be the current moment (`$$NOW`)


davinov:
Still needs some work, because if we replace all filters by $expr, they will be performance implications:
- indexes will no be used for mongo before 5
- and even in mongo 5 there is some limitations

raphaelvignes:
> indexes will no be used for mongo before 5
and even in mongo 5 there is some limitations 
:(

piotch:
@davinov cool!

I had a question which is beyond the scope of the PR : when you said it is relative to `$$NOW` I guess it is the UTC value, is it obvious  because we make sure all dates passing thru these filters are UTC as well ? I am not sure where we stand on this. Even if it is obvious maybe it would be worth documenting in the relevant confluence page that it is how it works.

src/lib/translators/mongo.ts
```diff
@@ -2261,26 +2261,26 @@ export class Mongo36Translator extends BaseTranslator {
   /**
    * Translate any value expression into one mongo can evaluate.
    *
-   * @private
+   * @protected
    */
-  private translateValue(value: any): any {
+  protected translateValue(value: any): any {
```
alice-sevin: pourquoi on passe a du protected plutot qu'à du private ici ?

alice-sevin (approved):
LGTM! A prévoir aujourd'hui on renvoit toujours une quantity négative pour une relative date je crois. Il faudra penser a corriger ça car j'imagine que fonctionnelement pour l'instant le $add ne marchera pas comme espéré

PR #1121, opened 2021-10-13
doc(dates): rework preparation of dates operations

davinov:
Precision have been added to the specification of or filtering capabilities with dates:
- an `is` operator should filter on a whole day
- `until` should be inclusive of the selected day

As their behavior is a bit different than simple `eq` or `le`, I suggest making dedicated operators, and therefore require a dedicated support in each backend. What do you think?


davinov:
Rework again the doc:
- remove the is operator for now (we decided it's ok to do "from a day" and "until a day" as it'll give the sale results
- describe the operators in relativeDates (@austil I'd like you're review on this)

PR #1164, opened 2021-11-02
feat(ui): reset button for DateRangeInput TCTC-1097

davinov:
A really gorgeous button to reset the selection to `undefined` (therefore the default value)
Screenshot from 2021-11-02 16-44-07


alice-sevin (approved):
Nothing to say :fire:

PR #1165, opened 2021-11-02
chore(build): storybook reload css

JeremyPinhel:
The storybook rollup config is based on the weaverbird config. So the extract option of postCss was kept at the launch of storybook which prevented the reload of the css when there was a change in a component.

So we override the postCss config by removing the extract option when we launch storybook.


austil (approved):
:green_circle:

PR #1166, opened 2021-11-03
fix(range calendar): clamp selection to bounds when switching granularity TCTC-1031

austil:
In the "Fixed" tab of the Date Range Calendar, when switching between tabs, with bounds configured, the selected range could end up out of bounds because : 

- on "larger than day" granularity we allow selection of any period that completely OR partially without configured bounds
- bounds where not taken into account when retrieving the appropriate value when arriving on a new calendar

This led to : 

- a wrong range being displayed as selected
- and no range appearing as selected at all in the case of the "day" calendar

This PR fix that.

## How

Make end calendars clamp the selected range to bounds before determining their corresponding value
and emit the end result back which is guaranteed to not be completely out of bound.

First tried to do this on their common parent `TabbedRangeCalendar` but wasn't appropriate. 
It was fine for clamping the selected range to current bound but not for determining the right value for the
current selected tab.

## Screenshots

Before, with bound set from February 2nd to October 3rd, when selecting Q4 and switching to the day calendar : 
 
image

After:

image

---

Before, with bound set from February 2nd to October 3rd, when selecting Q1 and switching to the month calendar : 

image

After:

image

### DIFF
--- a/src/components/DatePicker/Calendar.vue
+++ b/src/components/DatePicker/Calendar.vue
@@ -1,6 +1,6 @@
 <template>
   <DatePicker
-    :value="value"
+    :value="boundedValue"
     :availableDates="availableDates"
     :attributes="highlights"
     :select-attribute="selectedDatesStyle"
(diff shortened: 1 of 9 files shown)


src/components/DatePicker/Calendar.vue
```diff
@@ -84,6 +98,10 @@ export default class Calendar extends Vue {
     if (this.shouldUpdateDefaultDate) {
       this.defaultDate = this.availableDates?.start ?? '';
     }
+
+    if (!_isEqual(this.boundedValue, this.value)) {
+      this.onInput(this.boundedValue);
+    }
```
alice-sevin, line 104: Pour le duration 'day' / changement de granularity avec instanciation sur un autre calender ? Bien vu merci! Peut être qu'on peux rajouter un petit commit sur la raison de ce onInput car ca peut ne pas être clair pour tout le monde :pray: ?

PR #1167, opened 2021-11-03
fix(custom-granularity-calendar): update navigation based on selected option when switching between tabs TCTC-1031

alice-sevin:
When we switch between granulrity tabs of calendars we want to retrieve options that fit the current selected option.
Day calendar is already ok because it has a from property, just needed to update navStart in customgranulrity calendar on granulrity watcher

Before:
before

After:
after


davinov (approved):
Buen vu ! :clap:

PR #1175, opened 2021-11-10
build(playground): enable docker deloyment with a single python server TCTC-1139

davinov:
Retire the old node server in favor of a unique simple python server (playground.py) that will be able to us route between multiple back-ends (today pandas an mongo).
With this dockerfile, I will be able to enable the deployment of master and each PR to an online service (onrender.com). I showcased this on my personnal fork: https://github.com/davinov/weaverbird/pull/1 / https://weaverbird-playground.onrender.com/

Notable change in the server : I removed poetry, which was really long to resolve python dependencies, in favor of a more standard `setup.cfg` with [setuptools](https://setuptools.pypa.io/en/latest/).


davinov:
Once this gets merged, I'll use our github service account to setup render, and decommission the personal for I experimented the deployment on. I'll also setup payment using so that the services will always be up (not the case with the free plan)

luc-leonard (approved):
> I removed poetry,

Agreed with this decision. Poetry is nice to create env, but it is too strict on dependencies versions to be really useable, and the lack of `pip install -e .`  is kinda annoying.

thanks you for this PR. it is clean, well documented and will be useful !

PR #1177, opened 2021-11-15
refactor(relative date range): use a proper "operator" property TCTC-1183

austil:
Following lessons learned in https://github.com/ToucanToco/weaverbird/pull/1173
Refactor our `RelativeDateRange` type to explicitly include an "operator" instead of deducing it from quantity's sign.

### DIFF
--- a/src/components/DatePicker/transform-value-to-date-or-range.ts
+++ b/src/components/DatePicker/transform-value-to-date-or-range.ts
@@ -1,6 +1,12 @@
 import { DateTime } from 'luxon';
 
-import { DateRange, isDateRange, RelativeDate, RelativeDateRange } from '@/lib/dates';
+import {
+  DateRange,
+  isDateRange,
+  RELATIVE_DATE_RANGE_OPERATORS,
+  RelativeDate,
+  RelativeDateRange,
+} from '@/lib/dates';
 import { retrieveVariable, VariableDelimiters, VariablesBucket } from '@/lib/variables';
 
 /*
@@ -82,7 +88,8 @@ export const transformRelativeDateRangeToDateRange = (
   const target = DateTime.fromJSDate(targetDateTime, { zone: 'UTC' }).toJSDate();
 
   // if quantity is negative, target will arrive before base
-  const [start, end] = relativeDateRange.operator === 'from' ? [base, target] : [target, base];
+  const quantityIsPositive = RELATIVE_DATE_RANGE_OPERATORS[relativeDateRange.operator].sign > 0;
+  const [start, end] = quantityIsPositive ? [base, target] : [target, base];
   return { start, end };
 };
 
--- a/src/components/DatePicker/transform-value-to-date-or-range.ts
+++ b/src/components/DatePicker/transform-value-to-date-or-range.ts
@@ -67,7 +67,9 @@ export const transformRelativeDateRangeToDateRange = (
       // Base day should be included in full
       // so if the range starts with the base day, its time should be 00:00
       // but if the range ends with the base day, its time should be 23:59
-      relativeDateRange.quantity > 0 ? {} : { hour: 23, minute: 59, second: 59, millisecond: 999 },
+      relativeDateRange.operator === 'from'
+        ? {}
+        : { hour: 23, minute: 59, second: 59, millisecond: 999 },
     )
     .toJSDate();
 
@@ -80,7 +82,7 @@ export const transformRelativeDateRangeToDateRange = (
   const target = DateTime.fromJSDate(targetDateTime, { zone: 'UTC' }).toJSDate();
 
   // if quantity is negative, target will arrive before base
-  const [start, end] = relativeDateRange.quantity >= 0 ? [base, target] : [target, base];
+  const [start, end] = relativeDateRange.operator === 'from' ? [base, target] : [target, base];
   return { start, end };
 };
 
--- a/src/components/DatePicker/transform-value-to-date-or-range.ts
+++ b/src/components/DatePicker/transform-value-to-date-or-range.ts
@@ -67,6 +67,8 @@ export const transformRelativeDateRangeToDateRange = (
   )?.value;
   if (!(value instanceof Date)) return;
 
+  const operator = RELATIVE_DATE_RANGE_OPERATORS[relativeDateRange.operator];
+
   // pass base date to UTC
   const base = DateTime.fromJSDate(value, { zone: 'UTC' })
     .set(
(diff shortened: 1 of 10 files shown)


src/lib/dates.ts
```diff
@@ -28,11 +28,36 @@ export type RelativeDate = {
...
+ * and "Monday 15th" as an exemple date & "2 weeks" as an exemple delta
+ *
+ * We identified 4 possible operators:
+ * - until  = [date - delta ; date] => upper bound is included <=> [Tuesday 2nd  ; Monday 15th]
+ * - before = [date - delta ; date[ => upper bound is excluded <=> [Monday 1st   ; Sunday 14th]
+ * - from   = [date ; date + delta] => lower bound is included <=> [Monday 15th  ; Sunday 28th]
+ * - after  = ]date ; date + delta] => lower bound is excluded <=> [Tuesday 16th ; Monday 29th]
+ *
+ * However, only two of them have been implemented yet. We may or may not tweak those two or even
+ * implement the four of them after getting more user feedback regarding the relevance of current ones.
```
davinov, line 45: great explanation, thanks

src/lib/dates.ts
```diff
@@ -28,11 +28,36 @@ export type RelativeDate = {
...
+
+export type RelativeDateRangeOperator = keyof typeof RELATIVE_DATE_RANGE_OPERATORS;
+
 // Should be included in RelativeDate
 export type RelativeDateRange = {
   date: string;
-  quantity: number; // can be negative or positive
   duration: Duration;
+  operator: RelativeDateRangeOperator;
+  // Quantity can be negative or positive but the sign is ultimately dictated by the operator
```
davinov, line 59: :ok:

src/components/stepforms/widgets/DateComponents/RelativeDateRangeForm.vue
```diff
@@ -99,21 +104,24 @@ export default class RelativeDateRangeForm extends Vue {
...
+    return [RELATIVE_DATE_RANGE_OPERATORS.until, RELATIVE_DATE_RANGE_OPERATORS.from];
   }
 
-  get rangeDirection() {
-    return this.value.quantity >= 0 ? this.directions[1] : this.directions[0];
+  get operator() {
+    // Current value should have an operator but we introduced it without at first,
+    // relying entierly on quantity's sign to encode it.
+    // We keep this fallback as a reminder of this mistake until someone decide that comments
+    // is not the place for keeping track of our stuttering torward Clean Code :tm:
```
davinov, line 115: :trollface:

src/components/stepforms/widgets/DateComponents/RelativeDateRangeForm.vue
```diff
@@ -99,21 +104,24 @@ export default class RelativeDateRangeForm extends Vue {
...
+    const fallbackOperator = this.value.quantity >= 0 ? this.operators[1] : this.operators[0];
```
davinov, line 116: Detail: Using directly `RELATIVE_DATE_RANGE_OPERATORS.until` and `RELATIVE_DATE_RANGE_OPERATORS.from` could be a bit more readable (avoid relying on the order of operators and need to check this order above)

src/components/stepforms/widgets/DateComponents/RelativeDateRangeForm.vue
```diff
@@ -99,21 +104,24 @@ export default class RelativeDateRangeForm extends Vue {
     this.$emit('input', { ...this.value, date: value });
   }
 
-  get directions() {
-    return [
-      { label: 'until', value: -1 },
-      { label: 'from', value: +1 },
-    ];
+  get operators() {
```
davinov, line 107: So this is only the implemented operators, or available operators. Maybe we should name this with a prefix to mark the difference with the whole list of operators?

src/components/DatePicker/transform-value-to-date-or-range.ts
```diff
@@ -80,7 +82,7 @@ export const transformRelativeDateRangeToDateRange = (
   const target = DateTime.fromJSDate(targetDateTime, { zone: 'UTC' }).toJSDate();
 
   // if quantity is negative, target will arrive before base
```
davinov, line 84: This comment should be changed?

tests/unit/transform-value-to-date-or-range.spec.ts
```diff
@@ -92,6 +92,7 @@ describe('transformRelativeDateRangeToDateRange', () => {
       date: '{{toto}}',
       quantity: -2,
```
davinov, line 93: :warning: Should we remove in tests all traces of cases where the quantity is negative?
As the operator is now used to determine the order of base and target, these case will produce intervals where the end precedes the beginning?

davinov (commented):
Thanks! I have just a question about the negative quantities( see :warning: ), the rest is great :+1:

tests/unit/transform-value-to-date-or-range.spec.ts
austil: > As the operator is now used to determine the order of base and target

Full reasoning is "the operator is now used to determine the quantity sign which is used to determine the order of base and target.
I'm taking shortcut in the code but add a intermediate line to make it clear.

I don't see why this should impact tests in the end :shrug:

davinov: I didn't see any modifications in `https://github.com/ToucanToco/weaverbird/pull/1177/files#diff-f9f2b6ed3933ed5ca25d0627cfbad7ecc4e88799f09904ee7eba4afd3471daffR77-R80` (https://github.com/ToucanToco/weaverbird/pull/1177/files#diff-f9f2b6ed3933ed5ca25d0627cfbad7ecc4e88799f09904ee7eba4afd3471daffR77-R80) in your PR, so I assume that the sign is still used to compute the target date.

By "the operator is now used to determine the quantity sign ", I was expecting a modification in this function that computes the target date, but reading again I realize this is done way before, before the value is emitted from the component (`Math.sign(operator.sign`).

So that means we have a redundant info between the operator and the sign of the quantity. Is that OK and easy to understand for later readers?

austil: > Is that OK and easy to understand for later readers?

Depend of the reader :grimacing: 
I get what you mean. Because the info is redundant, today is does not matter if we use the quantity sign or the operator sign.

Getting rid of this redundancy would mean that quantity is a positive integer that always need to have the sign applied to it. We can go this way :shrug: 

As TypeScript don't have any positive integer type, it will still be a precondition only "enforced" here and there.
Properly enforcing it require a type constructor & private members aka a good old Class.
We almost never use Classes but should consider them more often IMHO.

davinov: > Getting rid of this redundancy would mean that quantity is a positive integer that always need to have the sign applied to it. We can go this way shrug

I would feel this matches the most the UI provided (we never ask our users to enter a negative quantity, but instead they do select an operator).
Still, I get that this clarity is somewhat subjective and I won't enforce you to do it before getting this PR approved.

Thanks for the clear answer and understanding :)

davinov (approved):

tests/unit/transform-value-to-date-or-range.spec.ts
austil: > this matches the most the UI

the saved object also read better : 
`{ quantity: -2, duration: 'month', operator: 'until', date: '{{today}}' }` feel a little wrong compared to
`{ quantity: 2, duration: 'month', operator: 'until', date: '{{today}}' }`

So currently trying to refactor it more, but deep down I know that this will probably never matter :no_mouth:

PR #1181, opened 2021-11-17
fix(calendars): reset value when selected range is out of updated bounds TCTC-1248

alice-sevin:
Avoid selected range error when upadting bounds and value is now out of bounds.
Reset value instead
Also add a new behaviour to adapt nav to new selected bounds if there is no value selected (avoid user to stay on the same calendar page and have no available options)

:warning: There is a bug with day calendar on my chrome version exclusively don't keep attention on it

after

### DIFF
--- a/tests/unit/custom-granularity-calendar.spec.ts
+++ b/tests/unit/custom-granularity-calendar.spec.ts
@@ -116,11 +116,27 @@ describe('CustomGranularityCalendar', () => {
     });
   });
 
-  describe('range out of bounds', () => {
+  describe('with bounds', () => {
     const bounds = {
       start: new Date('11/15/2021'),
       end: new Date('12/15/2021'),
-      duration: 'day',
+    };
+
+    beforeEach(() => {
+      createWrapper({
+        granularity: 'month',
+        bounds,
+      });
+    });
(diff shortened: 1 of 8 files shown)


tests/unit/custom-granularity-calendar.spec.ts
```diff
@@ -368,48 +415,13 @@ describe('CustomGranularityCalendar', () => {
...
-          start: new Date('2021-05-01'), // start of may 2021
-          end: new Date('2049-01-01'), // start of 2049
-        },
-      });
-      expect(wrapper.find('.header-btn__previous').classes()).toContain(
-        'custom-granularity-calendar__header-btn--disabled',
-      );
-      expect(wrapper.find('.header-btn__next').classes()).not.toContain(
-        'custom-granularity-calendar__header-btn--disabled',
-      );
```
austil, line 411: Why are we deleting all those assertions ? They are unnecessary ?

alice-sevin: Nop there are just not relevant since we move nav to bound start (it was not moving before)

austil (approved):
Well tested :muscle: 
That's a :green_circle:

PR #1184, opened 2021-11-18
fix(sql): fix filter step on date [TCTC-1242]

Sanix-Darker:
## What
- Fix Snowflake VQB:
   - [x] Filter Step
      - [x] Fix filter on date (is greater than/is lower than...) : 
      - [x] Unit Tests
      - [x] Fixtures tests

## Demo
Peek 2021-11-18 13-45

### DIFF
--- a/server/tests/backends/fixtures/filter/dates.json
+++ b/server/tests/backends/fixtures/filter/dates.json
@@ -1,5 +1,5 @@
 {
-  "exclude": [],
+  "exclude": ["sql"],
   "step": {
     "pipeline": [
       {
--- a/server/tests/backends/fixtures/filter/dates.json
+++ b/server/tests/backends/fixtures/filter/dates.json
@@ -1,7 +1,5 @@
 {
-  "exclude": [
-    "sql"
-  ],
+  "exclude": [],
   "step": {
     "pipeline": [
       {
--- a/server/tests/backends/fixtures/filter/dates.json
+++ b/server/tests/backends/fixtures/filter/dates.json
@@ -7,7 +7,7 @@
         "condition": {
           "column": "CREATED",
           "operator": "gt",
-          "value": "2020-01-01 01:03:30"
+          "value": "date: 2020-01-01 01:03:30"
         }
       }
     ]
--- a/server/tests/backends/fixtures/filter/dates.json
+++ b/server/tests/backends/fixtures/filter/dates.json
@@ -1,0 +1,74 @@
+{
+  "exclude": [
+    "sql"
+  ],
+  "step": {
+    "pipeline": [
+      {
+        "name": "filter",
+        "condition": {
+          "column": "CREATED",
+          "operator": "gt",
+          "value": "2020-01-01 01:03:30"
+        }
+      }
+    ]
+  },
+  "input": {
+    "schema": {
+      "fields": [
+        {
+          "name": "NAME",
+          "type": "string"
+        },
+        {
+          "name": "AGE",
+          "type": "integer"
+        },
+        {
+          "name": "CREATED",
+          "type": "datetime"
+        }
+      ],
+      "pandas_version": "0.20.0"
+    },
+    "data": [
+      {
+        "NAME": "foo",
+        "AGE": 42,
+        "CREATED": "2021-01-01 01:03:30"
+      },
+      {
+        "NAME": "bar",
+        "AGE": 43,
+        "CREATED": "2019-01-01 01:03:30"
+      }
+    ]
+  },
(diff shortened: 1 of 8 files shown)


davinov (changes requested):
Thanks! I have just a few comments:

server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
```diff
@@ -42,8 +43,14 @@
 def apply_condition(condition: Condition, query: str) -> str:
     if isinstance(condition, ComparisonCondition):
         try:
-            float(condition.value)
-            query += f'{condition.column} {SQL_COMPARISON_OPERATORS[condition.operator]} {condition.value}'
+            if isinstance(condition.value, datetime.datetime):
+                query += (
+                    f'{condition.column} {SQL_COMPARISON_OPERATORS[condition.operator]} '
+                    f'to_timestamp(\'{condition.value}\')'
```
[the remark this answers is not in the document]
Sanix-Darker: Good point !
Updated !

server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
```diff
@@ -42,8 +43,14 @@
 def apply_condition(condition: Condition, query: str) -> str:
     if isinstance(condition, ComparisonCondition):
         try:
-            float(condition.value)
-            query += f'{condition.column} {SQL_COMPARISON_OPERATORS[condition.operator]} {condition.value}'
+            if isinstance(condition.value, datetime.datetime):
```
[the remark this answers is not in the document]
Sanix-Darker: Good point, updated !

server/tests/backends/fixtures/filter/dates.json
```diff
@@ -0,0 +1,74 @@
+{
+  "exclude": [
+    "sql"
+  ],
```
davinov, line 4: should we not exclude the sql translator?

server/tests/backends/fixtures/filter/dates.json
```diff
@@ -0,0 +1,74 @@
...
+  "step": {
+    "pipeline": [
+      {
+        "name": "filter",
+        "condition": {
+          "column": "CREATED",
+          "operator": "gt",
+          "value": "2020-01-01 01:03:30"
```
davinov, line 12: hm.... how do we know this will be translated as datetime.datetime when we read this json in the tests?

server/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
```diff
@@ -43,10 +43,12 @@
...
         try:
-            if isinstance(condition.value, datetime.datetime):
+            if isinstance(condition.value, datetime.datetime) or isinstance(
+                condition.value, datetime.date
+            ):
                 query += (
-                    f'{condition.column} {SQL_COMPARISON_OPERATORS[condition.operator]} '
-                    f'to_timestamp(\'{condition.value}\')'
+                    f'to_timestamp({condition.column}) {SQL_COMPARISON_OPERATORS[condition.operator]} '
+                    f'to_timestamp(\'{condition.value.isoformat()}\')'
```
davinov, line 51: Good idea to turn both to timestamps! I guess it could also work without but that should definitely work :+1:

server/tests/backends/fixtures/filter/dates.json
Sanix-Darker: We know that because of the type of the column here :  https://github.com/ToucanToco/weaverbird/blob/67d2f0b802694615912c0237af26fde761c3301e/server/tests/backends/fixtures/filter/dates.json#L30 , if we change that value to a none datetime string the fixture will failed and we should have this kind of error : 
Screenshot from 2021-11-18 17-47-15

server/tests/backends/fixtures/filter/dates.json
Sanix-Darker: yeah, true, i forgot that !
Updated !

server/tests/backends/fixtures/filter/dates.json
davinov: That's true for the data, but does that also work when we read the pipeline?

server/tests/backends/fixtures/filter/dates.json
davinov: My bad! I din't get that sql != snowflake
So it's OK to exclude it.
Sorry!

davinov (approved):

PR #1185, opened 2021-11-19
chore(bump): from 0.8.5 to 0.8.8

Sanix-Darker:
## What
   - Bump of the weaverbird back : 
         - Fix on replace step
         - Fix on Filter on Date
   - Update the Makefile for installing build package

## Status
   - [x] Unit Tests


server/Makefile
```diff
@@ -43,6 +43,7 @@ all: test lint
 
 .PHONY: build
 build:
+	pip install build
```
davinov, line 46: :pray:

davinov (approved):

davinov:
@Sanix-Darker I got it!
Thing is: when you execute `pytest`, it doesn't add the rootdir (current folder to your PYTHONPATH). But when you use `python -m pytest`, it does! https://docs.pytest.org/en/6.2.x/usage.html#cmdline

PR #1186, opened 2021-11-19
chore(playground): add dates test case in playground TCTC-1192

davinov:
Also prepare the way for having a feature flag to test the new date selection component


davinov:
This PR is ready! I had to tackle the transmission of dates via JSON over the network. It was already OK from server to client (see annotateDataset / autocastDataset) but not the other way.

alice-sevin (approved):
LGTM, nothin to say, thanks :pray:

PR #1187, opened 2021-11-19
fix(variable-input-base): avoid variable to replace single string when selecting a variable in multiple mode TCTC-1047

alice-sevin:
Before:
before

After:
after


austil (approved):
Perfect ! LGTM :green_circle:

PR #1190, opened 2021-11-22
feat(new-date-input): enable to use advanced variable TCTC-1195

alice-sevin:
advanced variable

Didn't change other components behaviour


davinov (approved):
Very clear :+1: 
Thanks fr taking the time to even polish the design of the variable tag!

PR #1192, opened 2021-11-23
feat(playground): provide a mongo 5 container with data from playground TCTC-1197

davinov:
Necessary for working with relative dates.
This will be deployed on render, like the playground server.


davinov:
I switched the mongo used by the playground to the new one. Everything should continue to work as expected :)

alice-sevin (approved):

PR #1191, opened 2021-11-23
feat(filter-step): use specific operator for date columns TCTC-1193

alice-sevin:
When selecting a date column, transform the operators to have specific one : 'until'/'from'/'isnull'/'notnull' other column type refers to unchanged older types.

operators

PR #1193, opened 2021-11-23
feat(filter-step): replace date widget with new date input TCTC-1194

alice-sevin:
:warning: Need to review https://github.com/ToucanToco/weaverbird/pull/1191 first.

Replace date widget with new date input when a date column is selected in filter step
Also filter variables to return only dates one

Before:
operators

After:
date-input

**Design:**

design

Before:
before-1
before

After:
after-1
after

:warning: Optional (added a two last commit to use variable tag for variable and display value for variables in list, not sure if we want to keep them ? Useful for conceptor to know value of variable right ? )
There is a strange behaviour with tooltip when closing the dropdown, ( solved in the commit using !isEditorOpened, video has been made before :wink: )
hover

### DIFF
--- a/src/components/stepforms/widgets/DateComponents/NewDateInput.vue
+++ b/src/components/stepforms/widgets/DateComponents/NewDateInput.vue
@@ -35,6 +35,7 @@
           :availableVariables="availableVariables"
           :selectedVariables="selectedVariables"
           :enableCustom="enableCustom"
+          :showOnlyLabel="!isEditorOpened"
           @selectCustomVariable="editCustomVariable"
           @input="selectVariable"
           @addAdvancedVariable="openAdvancedVariableModal"
@@ -352,11 +353,11 @@ export default class NewDateInput extends Vue {
   height: 100%;
 }
 .widget-date-input__editor-side {
-  width: 200px;
-  min-width: 200px;
+  width: 285px;
+  min-width: 285px;
   height: 100%;
   max-height: 400px;
-  flex: 1 200px;
+  flex: 1 285px;
   overflow-x: hidden;
   overflow-y: auto;
 }
--- a/src/components/stepforms/widgets/DateComponents/NewDateInput.vue
+++ b/src/components/stepforms/widgets/DateComponents/NewDateInput.vue
@@ -1,5 +1,10 @@
 <template>
-  <div class="widget-date-input">
+  <div
+    class="widget-date-input"
+    :class="{
+      'widget-date-input--opened': isEditorOpened,
+    }"
+  >
     <div class="widget-date-input__container" @click.stop="openEditor">
       <VariableTag
         class="widget-date-input__advanced-variable"
@@ -295,8 +300,19 @@ export default class NewDateInput extends Vue {
   max-width: 400px;
   position: relative;
 }
+
+.widget-date-input--opened {
+  .widget-date-input__container {
+    border-color: $active-color;
+  }
+  .widget-date-input__icon {
+    background-color: $active-color-faded-2;
+    color: $active-color;
+  }
+}
+
 .widget-date-input__container {
-  border: 1px solid $grey-light;
+  border: 1px solid #ddd;
   display: flex;
   align-items: center;
   justify-content: space-between;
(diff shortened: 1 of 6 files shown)


src/components/stepforms/widgets/FilterSimpleCondition.vue
```diff
@@ -203,6 +205,12 @@ export default class FilterSimpleConditionWidget extends Vue {
     return this.featureFlags?.RELATIVE_DATE_FILTERING === 'enable';
   }
 
+  get useDateInput(): boolean {
+    return Boolean(
+      this.hasDateSelectedColumn && this.enableRelativeDateFiltering && this.inputWidget,
+    );
+  }
```
davinov, line 212: Do you use this computed?

src/components/stepforms/widgets/FilterSimpleCondition.vue
```diff
@@ -211,6 +211,14 @@ export default class FilterSimpleConditionWidget extends Vue {
     );
   }
 
+  get filteredAvailableVariables(): VariablesBucket | undefined {
```
davinov, line 214: This would be better named `dateAvailableVariables`

src/components/stepforms/widgets/FilterSimpleCondition.vue
```diff
@@ -211,6 +211,14 @@ export default class FilterSimpleConditionWidget extends Vue {
...
+    if (!this.useDateInput || !this.availableVariables) {
```
davinov, line 215: I think it may be easier to have just two set of variables whatever the case:
`availableVariables` and `dateAvailableVariables`, and then use the `inputWidget` to choose which one to use:
```
availableVariablesForInputWidget(inputWidget) {
  switch(inputWidget) {
    case NewInputWidget:
      return this.dateAvailableVariables;
   default:
     return this.availableVariables
}
```

This way, you do not need a computed (useDateInput) that needs to replicate and stay in sync with the widget actually chosen.

src/components/stepforms/widgets/DateComponents/NewDateInput.vue
```diff
@@ -8,7 +8,7 @@
     <div class="widget-date-input__container" @click.stop="openEditor">
       <VariableTag
         class="widget-date-input__advanced-variable"
-        v-if="advancedVariable"
+        v-if="variable || advancedVariable"
```
davinov, line 11: Good catch!

src/components/stepforms/widgets/DateComponents/NewDateInput.vue
```diff
@@ -35,6 +35,7 @@
           :availableVariables="availableVariables"
           :selectedVariables="selectedVariables"
           :enableCustom="enableCustom"
+          :showOnlyLabel="!isEditorOpened"
```
davinov, line 38: So the value of variables is displayed when we open the dropdown, but then hides when "custom" is selected?
Neat!

davinov (commented):
:100: 
Apart for a computed that I feel duplicates some operator logic, it's ready to merge:

src/components/stepforms/widgets/FilterSimpleCondition.vue
davinov: found in the next commit!

src/components/stepforms/widgets/DateComponents/NewDateInput.vue
alice-sevin: Yes it avoid the strange behaviour of tooltip going to top left screen when closing the popover :wink:

davinov:
While testing on the playground, I noted that the label used for the step doesn't seem to be ok for relative dates:
Screenshot from 2021-11-24 10-56-05
Screenshot from 2021-11-24 10-56-10

alice-sevin:
> While testing on the playground, I noted that the label used for the step doesn't seem to be ok for relative dates: Screenshot from 2021-11-24 10-56-05 Screenshot from 2021-11-24 10-56-10

Strange it take a look at it. It worked like a charm in local

davinov (approved):

PR #1194, opened 2021-11-24
feat(mongo): support from/until operators for fixed dates TCTC-905

davinov:


alice-sevin (approved):
Awesome, cristal clear :pray: !

PR #1195, opened 2021-11-24
fix(filter-step): keep relative date selected when switching from/until operators TCTC-1193

alice-sevin:
Keep a relative date selected when switching between relative date operators.
Before:
Relative date was transformed to a simple date because we didn't authorize relative date in Date validator
before

After:
Add a new condition to enable relative date, keep it unchanged if operator is from/until
validation

UAT:
Apply feature flag for relative date as enabled
Create a filter step on a date column
Select from operator
Select a relative date as value
Switch to until operator (should keep relative date value unchanged)

### DIFF
--- a/src/lib/dates.ts
+++ b/src/lib/dates.ts
@@ -160,6 +160,11 @@ export const relativeDateRangeToString = (
   return `${relativeDateLabel} ${relativeDateRange.operator}${CUSTOM_DATE_RANGE_LABEL_SEPARATOR}${baseDateLabel}`;
 };
 
+export const isRelativeDate = (value: any): value is RelativeDateRange => {
+  if (!(value instanceof Object)) return false;
+  return _has(value, 'duration') && _has(value, 'quantity');
+};
+
 export const isRelativeDateRange = (
   value: string | CustomDateRange | undefined,
 ): value is RelativeDateRange => {
--- a/src/lib/dates.ts
+++ b/src/lib/dates.ts
@@ -30,7 +30,10 @@ export type RelativeDate = {
 
 export function isRelativeDate(val: any): val is RelativeDate {
   return (
-    typeof val == 'object' && typeof val.quantity == 'number' && typeof val.duration == 'string'
+    val &&
+    typeof val == 'object' &&
+    typeof val.quantity == 'number' &&
+    typeof val.duration == 'string'
   );
 }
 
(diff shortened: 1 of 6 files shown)


src/lib/dates.ts
```diff
@@ -154,6 +154,11 @@ export const relativeDateRangeToString = (
   return `${relativeDateLabel} ${relativeDateRange.operator}${CUSTOM_DATE_RANGE_LABEL_SEPARATOR}${baseDateLabel}`;
 };
 
+export const isRelativeDate = (value: any): value is RelativeDateRange => {
+  if (!(value instanceof Object)) return false;
+  return _has(value, 'duration') && _has(value, 'quantity');
+};
```
davinov, line 160: Hihi I think this duplicates https://github.com/ToucanToco/weaverbird/pull/1103/files#diff-be9722f143d0d492db199700788ead0bc38d9b1b4bb9401e82f0f1434d8da1ceR31-R35
They are doing approximately the same thing we just need to be careful to remove one of the two when merging the second one.

tests/unit/lib/dates.spec.ts
```diff
@@ -167,6 +168,16 @@ describe('relativeDateRangeToString', () => {
...
 
+describe('isRelativeDate', () => {
+  it('should return false if value is not a relative date', () => {
+    expect(isRelativeDate('{{today}}')).toBe(false);
+    expect(isRelativeDate({ start: new Date() })).toBe(false);
+  });
+  it('should return true if value is a relative date', () => {
+    expect(isRelativeDate({ quantity: 2, duration: 'year' })).toBe(true);
+  });
+});
```
davinov, line 179: Thanks for testing this utility function!

src/lib/helpers.ts
```diff
@@ -110,6 +111,15 @@ export function keepCurrentValueIfCompatibleDate(
   return isDate(value) || typeof value === 'string' ? value : defaultValue;
 }
 
+export function keepCurrentValueIfCompatibleRelativeDate(
```
davinov, line 114: Should we name it whtout the "relative" part?
`keepCurrentValueIfCompatibleDate`
Because it works also for fixed dates

davinov (approved):
Great :100: Thanks identifying this bug and adding clear tests

One optional renaming suggestion:

src/lib/dates.ts
alice-sevin: Totally, i just thought the same when i was reviewing the other PR :wink:

src/lib/helpers.ts
alice-sevin: keepCurrentValueIfCompatibleDate already exists and need to be keeped until we have the featue flag (because relative date should not be enabled when feature flag is set to false ;) )

PR #1198, opened 2021-11-25
fix(filter-step): update invalid operator when switching relative date feature flag TCTC-1193

alice-sevin:
When relative date feature flag is switched, operators change so it may happen that some previously selected operator missed. We want to replace them with correct available operator:

until -> le
from -> ge
ge/gt -> from
le/lt -> until

If the operator is available fallback to it
If the operator is not in the list (ex: eq) fallback to first available operator

Also fixed the filter step behaviour because it was reassigning default value after the migration of operator


davinov (approved):
:+1: Nickel

PR #1201, opened 2021-11-29
feat(dates): use relative date with variables in new date input and clean relative date interfaces TCTC-1321

alice-sevin:
Use RelativeDateRangeForm in new date input
Fix referent methods (mongo/labeller/Transform value) to fit new behaviour of date input using relative date range

Cleaning:

After updating all methods, we consider that RelativeDate interface and RelativeDateForm component (as well as referent methods) are unused so we delete it and rename all RelativeDateRange interfaces to RelativeDate


src/lib/translators/mongo5.ts
```diff
@@ -1,17 +1,17 @@
...
 
-  protected translateRelativeDate(value: RelativeDate): object {
+  protected translateRelativeDate(value: RelativeDateRange): object {
     return {
       $dateAdd: {
         startDate: {
           // Base date does not include any hour information
           $dateTrunc: {
-            date: '$$NOW',
+            date: '$$NOW', // TODO: adapt date with selected variable
```
davinov, line 14: Thanks for this clear TODO :+1:

src/components/DatePicker/transform-value-to-date-or-range.ts
```diff
@@ -38,17 +37,34 @@ export const transformRelativeDateObjectToDate = (
...
-  // retrieve date from luxon
-  return transformRelativeDateObjectToDate({
-    ...relativeDate,
-    date,
+export const transformRelativeDateToDate = (
+  relativeDateRange: RelativeDateRange,
+  relativeAvailableVariables: VariablesBucket = [],
+  variableDelimiters: VariableDelimiters = { start: '', end: '' },
+): Date | undefined => {
+  // start is always a date as relativeAvailableVariables are not customizable
```
davinov, line 42: Not sure I got this comment on first read.
Could you confirm if this is correct : relativeAvailableVariables always contains Dates, and they can be changed by a user, in the UI.

src/lib/dates.ts
```diff
@@ -129,28 +128,22 @@ export const dateRangeToString = (dateRange: DateRange, locale?: LocaleIdentifie
...
   const identifier = extractVariableIdentifier(relativeDateRange.date, variableDelimiters);
-  const baseDateLabel = availableVariables.find(v => v.identifier === identifier)?.label;
-  const relativeDate = _pick(relativeDateRange, ['quantity', 'duration']);
-  const relativeDateLabel = relativeDateToString(relativeDate, { before: '', after: '' });
+  const baseDateLabel =
+    availableVariables.find(v => v.identifier === identifier)?.label ?? identifier;
+  const duration: string | undefined = DEFAULT_DURATIONS.find(
+    d => d.value === relativeDateRange.duration,
+  )?.label;
+  const suffix = ' ' + (relativeDateRange.quantity < 0 ? 'ago' : '');
```
davinov: I think the "ago" suffix would only apply when talking from a base date of "now". So it's best to remove this logic IMHO.
("3 days ago before today" isn't correct I think, but "3 days before today" is)

src/lib/dates.ts
```diff
@@ -129,28 +128,22 @@ export const dateRangeToString = (dateRange: DateRange, locale?: LocaleIdentifie
...
+  const relativeDateLabel = `${Math.abs(
+    relativeDateRange.quantity,
+  )} ${duration?.toLowerCase()}${suffix.trimEnd()}`;
   return `${relativeDateLabel} ${relativeDateRange.operator}${CUSTOM_DATE_RANGE_LABEL_SEPARATOR}${baseDateLabel}`;
```
davinov: Not really relative to this PR, but I also think we could remove the `CUSTOM_DATE_RANGE_LABEL_SEPARATOR`:
"3 days before today" reads best than "3 days before - today"
Do you agree?

src/components/DatePicker/transform-value-to-date-or-range.ts
```diff
@@ -40,58 +35,52 @@ export const transformRelativeDateObjectToDate = (
...
   return DateTime.fromJSDate(targetDateTime, { zone: 'UTC' }).toJSDate();
 };
 
 /*
-Read a `RelativeDateRange` and find the corresponding `Date` relative to the variable passed as `date` property.
+Read a `RelativeDate` and find the corresponding `Date` relative to the variable passed as `date` property.
 */
-export const transformRelativeDateRangeToDateRange = (
-  relativeDateRange: RelativeDateRange,
+export const transformRelativeDateToDateRange = (
```
davinov, line 65: Do we need to update Tucana as well? Maybe this function was used to compute the simple date selector values

src/lib/dates.ts
```diff
@@ -64,6 +64,11 @@ export const DEFAULT_DURATIONS: DurationOption[] = [
   { label: 'Days', value: 'day' },
 ];
 
+export const DEFAULT_RELATIVE_VARIABLES = [
+  { identifier: 'today', value: '$$NOW', label: 'TODAY' },
```
davinov, line 68: Another todo would be to adapt this "NOW" to each backend ($$NOW will not work in pandas or snowflake)

src/lib/dates.ts
```diff
@@ -64,6 +64,11 @@ export const DEFAULT_DURATIONS: DurationOption[] = [
...
+  // TODO: add enabled relative variables here
```
davinov, line 69: Actually we're not that sure we're gonna "hardcode" the relative date variables here (cf discussion with @austil last friday).
So let's keep that on hold and remove this TODO.

However, the playground is already providing a "now" variable, so it should still be testable :)
We can also add another variable (like "the first day of this year") to the playground if we cant to test more cases

src/components/Step.vue
```diff
@@ -257,6 +257,7 @@ export default class Step extends Vue {
   white-space: nowrap;
   text-overflow: ellipsis;
   overflow: hidden;
+  padding-right: 5px; //avoid label to be truncate due to emphasis style
```
davinov, line 260: Dark magic ahaha

davinov: Is this supposed to be the size of the ellipsis character? In that case, putting `1em` (size of 1 character) would be a it better because it would adapt to the font-size.

davinov (changes requested):
Excellent! A big chunk of work and cleaning, congrats :tada: 

The only "blocking" one I think is the question about the exported function that are modified: we don't want to break any usage of these, so we may need a PR to adapt Tucana too.
The rest are mostly questions

src/components/DatePicker/transform-value-to-date-or-range.ts
alice-sevin: I will had a proper explanation: it not relevant since calculated date can be in the past or the future now.
Before we always have start date as reference for start so it was always the selected variable and this variable comes from the relative available variables which is an array that conceptor cannot customize ( unless availableVariables that are customizable, but we don't really care about this information here)

src/components/DatePicker/transform-value-to-date-or-range.ts
```diff
@@ -16,7 +10,7 @@ If the `exclusive` option is set, add/remove 1 millisecond to the computed date,
 */
 
 export const transformRelativeDateObjectToDate = (
```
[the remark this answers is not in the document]
alice-sevin: It's still relevant (used by both function (transformRelativeDateToDate and transformRelativeDateToDateRange), it calculates the date adding quantity and duration (share util), it's not use in Tucana, we just export it to test it.

But indeed transformRelativeDateToDateRange has been modifin in his naming so we will need to rename it also in Tucana

src/lib/dates.ts
alice-sevin: Yes we will do it in another pr :+1:

src/lib/dates.ts
alice-sevin: Yes indeed, we should do it in another PR with the removal of CUSTOM_DATE_RANGE_LABEL_SEPARATOR

src/components/DatePicker/transform-value-to-date-or-range.ts
alice-sevin: Yes indeed

src/components/Step.vue
alice-sevin: i do it!

src/lib/dates.ts
alice-sevin: Not sure to understand, will ask you to explain me dircectly because it's not clear for me

davinov (approved):
Thanks for the answers and updates :+1:

PR #1202, opened 2021-11-29
feat(mongo): truncate dates to compare only days for from/until operators TCTC-1322

davinov:
- A new dataset with hours in the playground for tests
- Dates chosen from the calendar are always set at lmidnight UTC
- Change util/from comparison logic : use $dateTrunc for mongo 5+


src/components/DatePicker/Calendar.vue
```diff
@@ -105,17 +106,27 @@ export default class Calendar extends Vue {
...
+      const newDateRange: DateRange = { ...value, duration: 'day' };
+      if (newDateRange.start) {
+        newDateRange.start = setDateToMidnight(newDateRange.start);
+      }
+      if (newDateRange.end) {
+        newDateRange.end = setDateToMidnight(newDateRange.end);
+      }
+      this.$emit('input', newDateRange);
     }
   }
```
alice-sevin, line 125: C'est pas déjà le cas, on a un paramètre UTC sur le composant donc il nous retourne toujours des heures a minuit par défault due a v-calendar non ?

src/components/DatePicker/transform-value-to-date-or-range.ts
```diff
@@ -99,6 +99,10 @@ export const transformRelativeDateRangeToDateRange = (
   return { start, end };
 };
 
+export function setDateToMidnight(value: Date): Date {
+  return new Date(new Date(value).setUTCHours(0, 0, 0, 0));
+}
+
 export const setDateRangeHours = (value: DateRange | string | undefined): DateRange | undefined => {
   if (!isDateRange(value) || !value.start || !value.end) return;
   const startHours = { hour: 0, minute: 0, second: 0, millisecond: 0 };
```
alice-sevin, line 108: note: on devrait peut être renommer ce fichier j'ai l'impression qu'il ne fait pas uniquement de la transformation de valeur du coup :wink:

src/lib/translators/mongo.ts
```diff
@@ -2330,8 +2330,8 @@ export class Mongo36Translator extends BaseTranslator {
       return {
         $expr: {
           [operatorMapping[cond.operator]]: [
-            this.truncateColumnToDay(cond.column),
-            isRelativeDate(cond.value) ? this.translateRelativeDate(cond.value) : cond.value,
+            this.truncateDateToDay($$(cond.column)),
```
alice-sevin, line 2333: Je comprends pas pourquoi on applique le truncate aux deux (colonne/value) ici ?

alice-sevin (approved):
LGTM sur l'implem par contre je comprends pas tout, je connais certainement pas assez Mongo pour tout saisir mais on peux peut etre rajouter quelques commentaires aussi ? Pour qu'une personne non initiée puisse un peu mieux comprendre :pray: ?

tests/unit/mongo.spec.ts
```diff
@@ -606,12 +606,28 @@ describe.each(['36', '40', '42', '50'])(`Mongo %s translator`, version => {
           $and: [
             {
               $expr: {
-                $gte: ['$DateFrom', new Date('2021-11-24T00:00:00.000Z')],
+                $gte: [
+                  {
+                    $dateTrunc: {
+                      unit: 'day',
+                      date: '$DateFrom',
```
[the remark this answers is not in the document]
davinov: Good question! It's the name of the column we wish to compare.
This object means: use the `DateFrom` column, truncated to the day (i.e. without any hours - at midnight UTC)

src/components/DatePicker/Calendar.vue
davinov: Hm... good point
When testing this on the playground yesterday, we saw dates not at midnight but at the time we were testing.
Maybe we've missed setting a prop

davinov: Indeed! It seems we should use `timeAdjust` to do this (setting the timezone is not enough).

See https://vcalendar.io/datepicker.html#time-adjust

I'll do this, it's easier than recoding it ourselves!

src/components/DatePicker/transform-value-to-date-or-range.ts
davinov: I'm gonna remove this function and use `ŧimeAdjust`

src/lib/translators/mongo.ts
davinov: Good point! I'm gonna add comments

The idea is that we can never be so sure of the time of the variables that are gonna be passed to weaverbird.
Imagine we have a variable "now" that means `2021-30-11T10:32`, and someone makes a query with the filter step:
`{ column: 'dateCol', operator: 'until', value: '<%= now %>' }`.
Then the query translated in mongo will be roughly: 
`{ $match: { $expr: $lte: [ '$dateCol',  '<%= now %>']}}`.
If we replace the variablein this and execute the query, we will have the wrong hours (dateCol <= 2021-30-11T10:32).
So we need mongo to remove the hours on both side to make sure we comapre only days.

davinov: d8c8b76

src/components/DatePicker/transform-value-to-date-or-range.ts
davinov: 2cd56bb

src/components/DatePicker/Calendar.vue
davinov: 2cd56bb

src/components/DatePicker/Calendar.vue
```diff
@@ -60,6 +60,19 @@ export default class Calendar extends Vue {
...
+      timeAdjust: '00:00:00',
+      start: {
+        timeAdjust: '00:00:00',
+      },
+      end: {
+        timeAdjust: '23:59:59',
+      },
+    };
+  }
+
```
alice-sevin, line 75: Awesome!

alice-sevin (approved):
It's a lot clearer, tahnks a lot :pray:

PR #1203, opened 2021-11-29
feat(playground): improve the debug panel display

CharlesRngrd:
Hi Toucan,

Here a some UI improvements :

- Move the debug panel to a separated sheet.
- Move the debug button to a toolbar.

I don't have access to the main product components. So the code may be different than the main product ;)

test


davinov (approved):
Great ! Thanks for this :)

PR #1206, opened 2021-11-30
fix(relative-date): simplify label for relative date removing suffix and separator

alice-sevin:
Remove suffix and separator for relative date label going from:

2 months ago until - Last week
to
2 months until Last week


davinov (approved):
Great! Thanks!

PR #1207, opened 2021-11-30
feat(mongo): enable to use variable as base date for relative date filtering TCTC-1321

alice-sevin:
1. Retrieve relative variables based on date variables in availableVariables in NewDateInput
2. Use selected variable as date base for relative date filtering (mongo)


davinov (approved):
Perfect :100: thx!

PR #1210, opened 2021-11-30
fix(ui): component InputNumber return value as number

alice-sevin:
We returned value as string in input number, it should be a number.
Update statics steps because parseInt is not necessary anymore


davinov (approved):
Way clearer than before! Thanks for fixing this

PR #1211, opened 2021-11-30
feat(backend service):  configurable "preview source" rows subset TCTC-1324

austil:
Not the definitive design, just a quick but still functional implementation.

image


davinov (approved):
Cristal clear :gem: 

One question before merging: should we put this into weaverbird? or letting that responsabiliy to its host?

PR #1212, opened 2021-11-30
feat(new-date-input): send analytics event when user select a (advanced or base) variable TCTC-899

alice-sevin:
Add a new util to populate a sendAnalytics method from outside of weaverbird
Add a case in playground for UAT
Use it to send events when user select a variable or advanced variable in NewDateInput

Example in playground with method populated with a console.debug only:
analytics

### DIFF
--- a/src/lib/send-analytics.ts
+++ b/src/lib/send-analytics.ts
@@ -8,9 +8,9 @@
   - value: any (the data refering to this event that can be of any type depending on the context)
 */
 
-type SendAnalyticsManager = (_name: string, _value: any) => void;
+type SendAnalyticsManager = (_params: any) => void;
 
-let sendAnalytics: SendAnalyticsManager = (_name: string, _value: any) => {};
+let sendAnalytics: SendAnalyticsManager = (_params: any) => {};
 
 function defineSendAnalytics(sendAnalyticsManager: SendAnalyticsManager) {
   sendAnalytics = sendAnalyticsManager;
--- a/src/lib/send-analytics.ts
+++ b/src/lib/send-analytics.ts
@@ -1,0 +1,19 @@
+/*
+  This module enable to update the method used when the sendAnalytics method is called in the app
+  It would do nothing by default
+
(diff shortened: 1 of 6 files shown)


src/lib/send-analytics.ts
```diff
@@ -0,0 +1,19 @@
...
+  - name: string (the name of the referent event. ex: 'Save step xxx'
+  - value: any (the data refering to this event that can be of any type depending on the context)
+*/
+
+type SendAnalyticsManager = (_name: string, _value: any) => void;
+
+let sendAnalytics: SendAnalyticsManager = (_name: string, _value: any) => {};
+
+function defineSendAnalytics(sendAnalyticsManager: SendAnalyticsManager) {
+  sendAnalytics = sendAnalyticsManager;
```
davinov, line 16: Good idea to use a singleton!

tests/unit/lib/send-analytics.spec.ts
```diff
@@ -0,0 +1,15 @@
+import { defineSendAnalytics, sendAnalytics } from '@/lib/send-analytics';
+
+describe('defineSendAnalytics', () => {
+  it('should populate the send analytics method', () => {
+    const sendAnalyticsManager = (name: string, value: any) =>
+      `i will do anything you want! ${name} :: ${value}`;
+    defineSendAnalytics(sendAnalyticsManager);
+    // verify that method has been populate
```
davinov, line 8: `// verify that the exported sendAnalytics method has been changed`

davinov (commented):
Great! 
The singleton pattern is a good idea :+1: 
What would you thinkf of modifying a bit the `sendAnalytics` signature? Instead of using two arguments, we could use a single object argument ? This seems more "extendable" in the future, if we need more things than name and value.
As an example, instead of using `sendAnalytics('Date input - Select variable', value);` we could use `sendAnalytics({component: 'Date input', action: 'Select variable', value });`.

davinov (approved):

PR #1214, opened 2021-12-02
feat(sql,pandas): support from/until operators (comparison of days) TCTC-906 TCTC-907

davinov:
Implement the same functionnality than #1194 #1195 #1202 for pandas and SQL backends

Checklist:
- [x] set up test cases
- [x] support fixed dates in SQL translator 
- [x] support fixed dates in pandas executor 
- [x] ensure we do not use time in comparisons
- [x] support relative dates in SQL translator 
- [x] support relative dates in pandas executor

### DIFF
--- a/server/src/weaverbird/backends/pandas_executor/steps/utils/dates.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/utils/dates.py
@@ -18,5 +18,6 @@ def evaluate_relative_date(relative_date: RelativeDate) -> datetime:
         raise NotImplementedError
 
     return operation(
-        relative_date.date, relativedelta(**{relative_date.duration + 's': relative_date.quantity})
+        relative_date.date,
+        relativedelta(**{relative_date.duration + 's': relative_date.quantity}),  # type: ignore
     )
--- a/server/src/weaverbird/backends/pandas_executor/steps/utils/dates.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/utils/dates.py
@@ -17,7 +17,14 @@ def evaluate_relative_date(relative_date: RelativeDate) -> datetime:
     else:
         raise NotImplementedError
 
+    if relative_date.duration == 'quarter':
+        quantity = relative_date.quantity * 3
+        duration = 'months'
+    else:
+        quantity = relative_date.quantity
+        duration = relative_date.duration + 's'
+
     return operation(
         relative_date.date,
-        relativedelta(**{relative_date.duration + 's': relative_date.quantity}),  # type: ignore
+        relativedelta(**{duration: quantity}),  # type: ignore
     )
--- a/server/src/weaverbird/backends/pandas_executor/steps/utils/dates.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/utils/dates.py
@@ -1,0 +1,25 @@
+import operator
+from datetime import datetime
+
+from dateutil.relativedelta import relativedelta
+
+from weaverbird.pipeline.dates import RelativeDate
+
+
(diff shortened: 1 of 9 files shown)


davinov:
Ready! We can test it on the playground with Pandas, I'll make a second PR to add Snowflake to the playground

davinov:
Not ready yet actually, I missed the implementaion of 2 operators in relative dates

server/src/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
```diff
@@ -9,6 +9,7 @@
     Condition,
     ConditionComboAnd,
     ConditionComboOr,
+    DateBoundCondition,
```
alice-sevin, line 12: Pourquoi DateBound plutôt que RelativeDate ?

server/tests/backends/fixtures/filter/dates_from_until_strip_hours.json
```diff
@@ -0,0 +1,104 @@
+{
+  "description": "From/until operators should always include the whole selected day. Users don't select hours (and shouldn't care about them in most cases).",
```
alice-sevin, line 2: :+1:

server/src/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
davinov: This type correspond to the whole condition, so column + operator + value.
The value maybe either fixed or relative (so not always relative)

server/tests/backends/fixtures/filter/relative_dates.json
```diff
@@ -1,6 +1,6 @@
 {
   "description": "Dates from/until can be specified relative to another date.",
-  "exclude": ["sql", "snowflake"],
+  "exclude": ["sql"],
```
alice-sevin, line 3: Hum... le titre de la PR dit que c'est appliqué à sql aussi mais il est toujours exclu pour le moment non ?

alice-sevin (approved):
LGTM, a part le sql qui est toujours dans les exclude ?

davinov:
@alice-sevin actually, it's normal to have `exclude: ["sql"]`
`sql` is for testing against mysql, while `snowflake` is for tetsing against snowflake.
We may put some effort to become compatible with mysql and postgresql, but for now these engines are not run in tests.

server/src/weaverbird/backends/pandas_executor/steps/utils/dates.py
```diff
@@ -17,7 +17,14 @@ def evaluate_relative_date(relative_date: RelativeDate) -> datetime:
     else:
         raise NotImplementedError
 
+    if relative_date.duration == 'quarter':
+        quantity = relative_date.quantity * 3
+        duration = 'months'
+    else:
+        quantity = relative_date.quantity
+        duration = relative_date.duration + 's'
+
```
alice-sevin, line 26: Est-ce qu'on a besoin de faire la même chose pour les semestre aussi @davinov ?

davinov: Je crois qu'on pas les semestres dans la spec des dates relatives, si ?

davinov: Check: pas de semestres ici: https://github.com/ToucanToco/weaverbird/blob/master/server/src/weaverbird/pipeline/dates.py#L11 https://github.com/ToucanToco/weaverbird/blob/master/src/lib/dates.ts#L19

PR #1218, opened 2021-12-07
fix(mongo): always truncate time to day in date extract step TCTC-1283

austil:
Some user noticed that some option of the step "date extract" were keeping the source date time while other were not (and that's what we want).

Example with : 
image

`firstDayOfMonth` truncate the time while `firstDayOfIsoWeek` does not.

image

Affected date info were : 

- firstDayOfWeek
-  firstDayOfIsoWeek
- previousDay
- firstDayOfPreviousWeek
- firstDayOfPreviousIsoWeek


src/lib/translators/mongo.ts
```diff
@@ -8,6 +8,7 @@ import { $$, combinations, escapeForUseInRegExp } from '@/lib/helpers';
 import { OutputStep, StepMatcher } from '@/lib/matcher';
 import * as S from '@/lib/steps';
 import { BaseTranslator, ValidationError } from '@/lib/translators/base';
+import { ADVANCED_DATE_EXTRACT_MAP, DATE_EXTRACT_MAP } from '@/lib/translators/mongo_dates';
```
davinov: Nitpick: our js/ts source files are generally named with dashes instead of underscores

src/lib/translators/mongo_dates.ts
```diff
@@ -179,3 +179,34 @@ export const ADVANCED_DATE_EXTRACT_MAP: Record<
     $isoWeek: { $subtract: [$$(step.column), 7 * 24 * 60 * 60 * 1000] },
   }),
 };
+
+type MongoStep = Record<string, any>;
+
+export const transformDateExtractFactory = (advancedDateExtractMap = ADVANCED_DATE_EXTRACT_MAP) => {
```
davinov, line 212: Nice idea to have a factory. I personally found it easier to read to have protected methods or attributes that would get overridden, but it's just a matter of style.

src/lib/translators/mongo_dates.ts
```diff
@@ -180,6 +175,58 @@ export const ADVANCED_DATE_EXTRACT_MAP: Record<
...
+  firstDayOfWeek: step => ({
+    // [ ] TO FIX using Mongo 5
+    // We subtract to the target date a number of days corresponding to (dayOfWeek - 1)
+    $subtract: [
+      $$(step.column),
+      {
+        $multiply: [{ $subtract: [{ $dayOfWeek: $$(step.column) }, 1] }, 24 * 60 * 60 * 1000],
+      },
+    ],
+  }),
```
davinov: For Mongo 5, the easiest is to use the $dateTrunc operator, which does exactly what you need with the unit `week`.
https://docs.mongodb.com/manual/reference/operator/aggregation/dateTrunc/
Beware that `startOfWeek` defaults to sunday, so we may want to precise monday for our use case.

And actually, the same operator can be leverage for other extractions in mongo 5+. Maybe it's worth using it not only for weeks but for first day of month, first day of quarter, etc.

src/lib/translators/mongo_dates.ts
```diff
@@ -180,51 +190,21 @@ export const ADVANCED_DATE_EXTRACT_MAP_MONGO_5: Record<
...
-    // We subtract to the target date a number of days corresponding to (isoDayOfWeek - 1)
-    $subtract: [
-      { $subtract: [$$(step.column), 7 * 24 * 60 * 60 * 1000] },
-      {
-        $multiply: [{ $subtract: [{ $isoDayOfWeek: $$(step.column) }, 1] }, 24 * 60 * 60 * 1000],
-      },
-    ],
-  }),
+  firstDayOfWeek: step => {
+    return truncateDateToDay(ADVANCED_DATE_EXTRACT_MAP['firstDayOfWeek'](step));
```
davinov, line 194: I you use the $dateTrunc operator on a unit > day (see previous comment), calling it again with the unit day will not make any difference.

tests/unit/mongo.spec.ts
```diff
@@ -2771,163 +2771,155 @@ describe.each(['36', '40', '42', '50'])(`Mongo %s translator`, version => {
...
-          firstDayOfIsoWeek: {
-            // We subtract to the target date a number of days corresponding to (isoDayOfWeek - 1)
-            $subtract: [
-              '$foo',
-              {
-                $multiply: [{ $subtract: [{ $isoDayOfWeek: '$foo' }, 1] }, 24 * 60 * 60 * 1000],
-              },
+        ],
+      },
+      previousDay: { $subtract: ['$foo', 24 * 60 * 60 * 1000] },
```
davinov, line 2836: Instead of substracting the duration (this may be false for days bwith daylight savings), prefer using the $dateSubstract operator in mongo5, which is specifically done to handle these cases
https://docs.mongodb.com/manual/reference/operator/aggregation/dateSubtract/

davinov (changes requested):
Thanks for handling this :)
Here a few comments:

Main suggestion is to use $dateTrun and $dateSubstract when possible, as they are much better than substracting seconds manually or recreating a date from parts.

davinov (approved):
Thanks to have taken this <3

PR #1219, opened 2021-12-08
feat(new-date-input): add analytics when saving a relative date TCTC-899

alice-sevin:
Add analytics when saving a relative date for date filtering


austil (approved):
:green_circle:

PR #1222, opened 2021-12-09
fix(sql): align SnowFlake date extract results with other backends TCTC-1283

austil:
When checking if some date extract part were returning non-truncate date like in https://github.com/ToucanToco/weaverbird/pull/1218, we noticed that this step was returning wrong result on many operators.

We did our best to get similar result than what Pandas return. However SnowFlake (our main SQL target) "week based" date functions depend on the sessions parameters `WEEK_START` & `WEEK_OF_YEAR_POLICY` (https://docs.snowflake.com/en/sql-reference/parameters.html#week-of-year-policy) and the default value are inappropriate regarding some of our goals.


server/src/weaverbird/backends/sql_translator/steps/utils/query_transformation.py
```diff
@@ -341,51 +341,61 @@ def get_query_for_date_extract(
     This method will get as input the date type and return a query with
     the appropriate function of that date type on snowflake, it can be a simple function or a whole expression
 
-
+    Notes on Snowflake SQL:
```
raphaelvignes, line 344: Thx for this doc :+1:

raphaelvignes (commented):
If I understand correctly, We adapted the fixture to have passing test based on start of the week = monday right ? 
What if this param is set to Sunday in the session ?

austil:
> What if this param is set to Sunday in the session ?

Well, we'll get the expected result for "non Iso Week based operations" (compared to other Weaverbird backend).
We could use iso based operator everywhere to have "reliably broken non Iso Week based operations" :sweat_smile: or, at some point, always alter those variable ourselves to set them on Sunday.

austil:
> We adapted the fixture to have passing test based on start of the week = monday right ?

yup, this for operation that should have returned result based on start of the week = sunday.
And there is also this second difference where the [Python week number](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes) consider that

> All days in a new year preceding the first Sunday are considered to be in week 0.

while Snowflake goes for (by default) for an : 

> equivalent to the ISO semantics, in which a week belongs to a given year if at least 4 days of that week are in that year.

but can be configured to : 

> January 1 is included in the first week of the year and December 31 is included in the last week of the year.

:exploding_head:

raphaelvignes (approved):
Thanks for the explanations. 
Let's close this topic.

PR #1226, opened 2021-12-13
test(server): some tests were not played

davinov:
Like dateextract tests because they were called date_extract in python
files of backends...


austil (approved):
:+1:

PR #1227, opened 2021-12-13
fix(pandas): some dateextract were failing if some dates were duplicated TCTC-1379

davinov:
This is very edge case found by adding a "missing dates" step with some
duplicate dates but not in the same group.

While fixing this, I noticed that the tests of the dateextract step were not executed because f a filename mismatch! I fixed this in #1226, so this PR needs to be merged first for the test case to be run.


austil (approved):
:green_circle: :checkered_flag:

PR #1229, opened 2021-12-15
chore: export date range to string method TCTC-1398

alice-sevin:
export method to use it outside of weaverbird


alice-sevin:
Not sure we need it, keep it open until we verify

alice-sevin:
Finalement on en a bien besoin :laughing:

austil (approved):
:green_circle:

PR #1235, opened 2021-12-20
feat(filter-step): rename label for from/until date operators TCTC-1443

alice-sevin:
Update labels of from/until operator in filter step when filtering on a date


davinov (approved):
:+1:

PR #1238, opened 2021-12-20
fix(stepforms): change placeholders color [TCTC-1380]

raphaelvignes:
# What
Making placeholders in step forms appear different than actual values.

# Why
https://toucantoco.atlassian.net/browse/TCTC-1380 

# How ?
Well ... css

### DIFF
--- a/src/components/stepforms/widgets/InputText.vue
+++ b/src/components/stepforms/widgets/InputText.vue
@@ -88,7 +88,7 @@ export default class InputTextWidget extends Mixins(FormWidget) {
   }
 
   &::placeholder {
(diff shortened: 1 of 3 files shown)


davinov:
It looks really really light, doesn't it?
Screenshot from 2021-12-20 17-09-25

raphaelvignes:
let's go for `$grey-medium: #aaaaaa;` ?

davinov (approved):
Way better! Thx

Before:
Screenshot from 2021-12-20 18-25-26
After:
Screenshot from 2021-12-20 18-25-03

PR #1012, opened 2021-08-20
RollupStep : Make the aggregation optional (usefull for requesters)

CharlesRngrd:
Hi Toucan,

As a conceptor, I use the rollup step to generate the requesters.
The aggregation is required, that is confusing...

So here is my PR to avoid putting a dummy column in the aggregation form ;)

Have a nice day,

Charles

PR #1245, opened 2022-01-12
feat(rollup): no longer needing aggregations TCTC-1552

davinov:
Reopening #1012 with a few changes (such as having an empty aggregation array instead of no aggregation property at all, and rework of the pandas implementation to avoid creating a temporary column)


davinov:
Oops! The SQL version still creates an aggregate column by default... Let me correct this ^

raphaelvignes (approved):
Thx !

PR #992, opened 2021-08-17
feat(mongo): division by zero and null in formulas

CharlesRngrd:
Hi Toucan,

As a conceptor, I need to divide 2 columns but it may crash during data refresh because of division by zero or null.
So here is my PR to allow division by zero or null.

Notes :
- In Pandas, the code was working but I reshaped the unit test

Have a nice day,

Charles

PR #1247, opened 2022-01-12
feat(mongo): division by zero and null in formulas TCTC-1552

davinov:
#992 reopened to resolve conflicts
Contrib by @CharlesRngrd (thanks a lot!)


src/lib/translators/mongo.ts
```diff
@@ -1946,13 +1946,52 @@ export class Mongo36Translator extends BaseTranslator {
...
+          }
+          if (typeof obj[key] === 'object') {
+            iterate(obj[key]);
+          }
+        });
+      }
     };
+
+    // The first element was set for type reason
+    denominators.shift();
```
austil, line 1970: #HowToTypeScript ? :sweat_smile: 
Can we get rid of this or are we really stuck with this code ?

src/lib/translators/mongo.ts
```diff
@@ -1946,13 +1946,52 @@ export class Mongo36Translator extends BaseTranslator {
...
+
+    iterate(mongoFormulaTree);
```
austil, line 1972: Could be renamed "collectDenominator" or something more explicit ?

src/lib/translators/mongo.ts
```diff
@@ -1946,13 +1946,52 @@ export class Mongo36Translator extends BaseTranslator {
...
+        $addFields: {
+          [step.new_column]: mongoFormulaTree,
+        },
+      };
+    } else {
+      const conditions = [];
+
+      for (const d in denominators) {
+        conditions.push({ $in: [denominators[d], [0, null]] });
+      }
```
austil, line 1985:
```suggestion
      const conditions = denominators.map((d) => ({ $in: [d, [0, null]] }));
```

Bit weird to use `for...in` for iterating over arrays, right ? :sweat_smile:

davinov:
@austil I've taken your review as an opportunity to rework how we collect denominators so it doesn't introduce another iteration logic over formula trees twice.

2c72f2d 

and merged to fix changelog conflicts

austil (approved):
:green_circle:

PR #1248, opened 2022-01-12
feat(mongo): concatenate won't fail on non-string columns TCTC-1552

davinov:
#990 + code review


src/lib/translators/mongo.ts
```diff
@@ -703,10 +703,13 @@ function transformCumSum(step: Readonly<S.CumSumStep>): MongoStep {
 }
 
 /** transform a 'concatenate' step into corresponding mongo steps */
-function transformConcatenate(step: Readonly<S.ConcatenateStep>): MongoStep {
-  const concatArr: string[] = [$$(step.columns[0])];
+export function transformConcatenate(
+  step: Readonly<S.ConcatenateStep>,
+  convertColumnName: (c: string) => string | object = $$, // allow to override the conversion function of each column
+): MongoStep {
```
austil, line 709: When reading this first, I understood it was about renaming the column for some purpose.
But it's apparently here to add a conversion instruction (or anything really) on each targeted column content before concat. Do we have a better param name to describe that ?

Also, can we add to the comment that this is used to introduce different behavior between mongo versions ? So the day we decide to only support the latest, we could get rid of it easely :wink:

davinov: Good idea :+1:

davinov: Here is a proposal for a implemention using override of a protected method: df4ad061
(It's the pattern we used to write the mongo 4.2 and 5 translators, and it makes clearer that a behavior is supported only from that version)

austil (approved):
That's perfect ! :green_circle:

PR #1249, opened 2022-01-12
feat(ui): select columns from the right dataset in the join step TCTC-1552

davinov:
#1008 + modifications about where the column names are stored
Contrib by @CharlesRngrd


alice-sevin (approved):
LGTM

PR #1251, opened 2022-01-17
fix(ui): cumsumstep migration should delete old properties TCTC-1552

davinov:
The check introduced in #999 / #1246 for mongo translator to be able to handle legacy cumsum step format need the properties 'valueColumn' and 'newColumn' not be present (not just present to undefined) to interpret correctly new steps with the 'toCumSum' attribute.

This PR fixes the format produced by the CumSumStepForm so these two old attributes are not present anymore.

Before:
Screenshot from 2022-01-17 15-19-13

After:
Screenshot from 2022-01-17 15-18-44


raphaelvignes (approved):
Seems clean, thanks for the fix.

PR #1252, opened 2022-01-17
chore(ui): add data cy for date components, pagination and source subset preview TCTC-1515

alice-sevin:
Only add data-cy in order to add e2e tests for missing features


davinov (approved):
:+1:

PR #1253, opened 2022-01-18
feat: domains can be replaced by references to other queries  TCTC-1548

davinov:
These references will not resolved by weaverbird, but can be used by its host to create dependencies and enable query reuse and factorization.

I've added a few layers of protections to ensure the  (front-end) mongo translator will fail if we ever try to pass references to it. These references are intended to be resolved back-end side.

I the process, I cleaned the way domain was handled in the UI store. (There is no need for it to have a special treatment anymore, and it leads to code simplification/deletion ace6347 )

### DIFF
--- a/server/src/weaverbird/pipeline/steps/utils/combination.py
+++ b/server/src/weaverbird/pipeline/steps/utils/combination.py
@@ -1,3 +1,12 @@
-from typing import List, Union
+from typing import List, Literal, Union
+
+from pydantic import BaseModel
+
+
+class Reference(BaseModel):
+    type: Literal['ref']
+    uid: str
+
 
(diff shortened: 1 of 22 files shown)


server/src/weaverbird/pipeline/steps/utils/combination.py
```diff
@@ -1,3 +1,12 @@
-from typing import List, Union
+from typing import List, Literal, Union
+
+from pydantic import BaseModel
+
+
+class Reference(BaseModel):
```
raphaelvignes, line 6: :+1:

raphaelvignes (commented):
Ok pour moi la partie Back :+1: 
Est-ce que ça appelle une PR Laputa pour mettre à jour la méthode `retrieve_domain` ?

alice-sevin (approved):

PR #1258, opened 2022-01-20
fix(step): avoid display previous error when creating a new step TCTC-1343

alice-sevin:
Before:
before

After:
after

When we create a new step we don't entre the editStep method so we don't update backendErrorMessage display.
The only way to hide it is to base our display on if the step is a newly created one or not. So i added a computed for this case.

:information_source: Didn't find a way to test it, feel free to help me on it if you have an idea :pray:


davinov (approved):
:+1:

PR #1262, opened 2022-01-25
feat(python backends): add new mongo translator [TCTC-1747, TCTC-1748]

luc-leonard:
This PR create a new translator in the backend, for mongoDB.

It is quite barebone at the moment: 
* only some steps are implemented: `domain`, `text` ,`filter`, `argmin` and `argmax`
* test runner is functionnal, and will spawn a mongoDB in docker

PR is quite big, but most of it is excluding the new engine from existing tests


davinov:
Cool! It seems that pymongo is missing from the python packages requirements to run the tests. Can you add it?

server/src/weaverbird/backends/mongo_translator/mongo_pipeline_translator.py
```diff
@@ -0,0 +1,25 @@
...
+from weaverbird.pipeline import Pipeline, PipelineStep
+
+
+def translate_pipeline(
+    pipeline_to_translate: Pipeline,
+) -> list:
+    mongo_pipeline = []
+    for index, step in enumerate(pipeline_to_translate.steps):
+        try:
+            mongo_pipeline.append(mongo_step_translator[step.name](step))
```
davinov, line 11: Does append works like concat when we have lists ?
(To avoid having a list of lists in the end, which needs flattening)

server/src/weaverbird/backends/mongo_translator/mongo_pipeline_translator.py
```diff
@@ -0,0 +1,25 @@
...
+        except Exception as e:
+            raise PipelineTranslationFailure(step, index, e) from e
```
davinov, line 13: :+1:

server/src/weaverbird/backends/mongo_translator/mongo_pipeline_translator.py
```diff
@@ -0,0 +1,25 @@
+from weaverbird.backends.mongo_translator.steps import mongo_step_translator
+from weaverbird.pipeline import Pipeline, PipelineStep
+
+
+def translate_pipeline(
+    pipeline_to_translate: Pipeline,
+) -> list:
```
davinov, line 7: `list[dict[str, Any]]` to be more precise?

davinov: This will make mypy check that each step function return a compatible type too :)

server/src/weaverbird/backends/mongo_translator/steps/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+from weaverbird.backends.mongo_translator.steps.argmax import translate_argmax
+from weaverbird.backends.mongo_translator.steps.argmin import translate_argmin
+from weaverbird.backends.mongo_translator.steps.domain import translate_domain
+from weaverbird.backends.mongo_translator.steps.filter import translate_filter
+from weaverbird.backends.mongo_translator.steps.text import translate_text
+from weaverbird.pipeline.steps import BaseStep
+
+# I would like to have a better type for the callable, but I don't know how to do it.
+# each of this function take a particular step as input
+# so it is not possible to use `Dict[str, Callable[[BaseStep], list]]
```
davinov, line 12: :'( it's because it says BaseStep is not compatible with each concrete step? wierd!

server/src/weaverbird/backends/mongo_translator/steps/filter.py
```diff
@@ -0,0 +1,96 @@
...
+    'nin': '$nin',
+    'isnull': '$eq',
+    'notnull': '$ne',
+    'from': '$gte',
+    'until': '$lte',
+}
+
+
+def translate_relative_date(value: RelativeDate):
+    RELATIVE_DATE_OPERATORS: Dict[str, int] = {'until': -1, 'from': +1}
```
davinov, line 33: May be defined outsisde this function?

server/src/weaverbird/backends/mongo_translator/steps/filter.py
```diff
@@ -0,0 +1,96 @@
...
+    sign = RELATIVE_DATE_OPERATORS[value.operator]
+    return {
+        '$dateAdd': {
+            'startDate': {
+                '$dateTrunc': {
+                    'date': {'$toDate': value.date},
+                    'unit': 'day',
+                },
```
davinov, line 41: You can use your function `truncate_to_day` here?

server/tests/backends/mongo_translator/test_mongo_translator_steps.py
```diff
@@ -0,0 +1,80 @@
...
+                casted_row[key] = value
+        casted_data.append(casted_row)
+    return casted_data
+
+
+@pytest.mark.parametrize('case_id,case_spec_file_path', test_cases)
+def test_mongo_translator_pipeline(mongo_collection, case_id, case_spec_file_path):
+    # insert in mongoDB
+    spec = get_spec_from_json_fixture(case_id, case_spec_file_path)
+    data = cast_to_schema(spec['input'])
```
davinov, line 66: Would it be easier to use pandas to read the schema, and then .to_dict() before sending it to pymongo? (I haven't tried, so it's just a suggestion to avoid having to maintain the cast_to_schema implementation)

davinov (commented):
Great! Here a minor comments:

(I haven't had the time to finish reviewing, will do later!)

server/src/weaverbird/backends/mongo_translator/steps/__init__.py
luc-leonard: yes. translate_filter does not take a `BaseStep` as an argument, but a `FilterStep` I am not sure how to express this in python, in java it would be something like `Function<? extends BaseStep, List>

davinov (approved):
LGTM! :+1:

PR #1263, opened 2022-01-25
fix(ui): trully load all unique column values TCTC-1688

austil:
The "load all values" button bottom of the unique column values list was a lie, it was only loading unique values found in the first 10K rows. Some users noticed dat. Let's fix it !

image


davinov (approved):
:+1:

PR #1276, opened 2022-03-08
feat:python mongo append step [TCTC-1829]

raphaelvignes:
* Move Append step from front mongo translator to Back.

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/steps/append.py
+++ b/server/src/weaverbird/backends/mongo_translator/steps/append.py
@@ -1,0 +1,40 @@
+import logging
+from typing import List
+
+from weaverbird.pipeline import Pipeline, steps
+from weaverbird.pipeline.steps import AppendStep, DomainStep
+
+logger = logging.getLogger(__name__)
+
+
+def translate_append(step: AppendStep) -> List:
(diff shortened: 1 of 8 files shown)


server/src/weaverbird/backends/mongo_translator/steps/append.py
```diff
@@ -0,0 +1,40 @@
...
+                }
+            }
+        )
+        pipelines_names.append(f'$_vqbPipelineToAppend_{i}')
+    return [
+        {'$group': {'_id': None, '_vqbPipelineInline': {'$push': '$$ROOT'}}},
+        *lookups,
+        {'$project': {'_vqbPipelinesUnion': {'$concatArrays': pipelines_names}}},
+        {'$unwind': '$_vqbPipelinesUnion'},
+        {'$replaceRoot': {'newRoot': '$_vqbPipelinesUnion'}},
```
davinov, line 39: Very clear to read it like that :+1:

davinov (approved):

server/src/weaverbird/backends/mongo_translator/steps/append.py
raphaelvignes: thx

PR #1280, opened 2022-03-11
feat(mongo back): rename step TCTC-1830

raphaelvignes:
* Rename Step in Mongo Back.

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/steps/rename.py
+++ b/server/src/weaverbird/backends/mongo_translator/steps/rename.py
@@ -5,6 +5,12 @@ from weaverbird.pipeline.steps import RenameStep
 
 def translate_rename(step: RenameStep) -> List:
     return [
-        {'$addFields': {renaming[1]: f'${renaming[0]}'} for renaming in step.to_rename},
-        {'$project': {renaming[0]: 0} for renaming in step.to_rename},
+        {'$addFields': {to_rename[1]: f'${to_rename[0]}' for to_rename in step.to_rename}},
+        {
+            '$project': {
+                to_rename[0]: 0
+                for to_rename in step.to_rename
+                if to_rename[0] not in [to_rename[1] for to_rename in step.to_rename]
+            }
+        },
     ]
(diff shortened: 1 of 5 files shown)


server/src/weaverbird/backends/mongo_translator/steps/rename.py
```diff
@@ -0,0 +1,10 @@
+from typing import List
+
+from weaverbird.pipeline.steps import RenameStep
+
+
+def translate_rename(step: RenameStep) -> List:
+    return [
+        {'$addFields': {renaming[1]: f'${renaming[0]}'} for renaming in step.to_rename},
```
davinov, line 8: I think there is a subtle mistake here: the `for` seems outside of the dict comprehension, I think it should be inside.
Check the difference in output:
```python
>>> {'$addFields': {renaming[1]: f'${renaming[0]}'} for renaming in [['a', 'b'], ['c', 'd']]}
{'$addFields': {'d': '$c'}}
>>> {'$addFields': {renaming[1]: f'${renaming[0]}' for renaming in [['a', 'b'], ['c', 'd']]}}
{'$addFields': {'b': '$a', 'd': '$c'}}
```

To be sure not to miss that, what about enriching the test case with multiple columns being renamed at once?

server/src/weaverbird/backends/mongo_translator/steps/rename.py
```diff
@@ -0,0 +1,10 @@
...
+        {'$project': {renaming[0]: 0} for renaming in step.to_rename},
```
davinov, line 9: Same note as above, with a twist: if we want to rename column A to B and column B to C, then we'll end up with B being removed (only C in the output).
Even if this is a subtle edge case that should honestly not happen a lot, we can manage it quite easily iof we add a condition to the suppression: 
`for renaming in step.to_rename if renaming[1] not in new_column_names` with `new_column_names` being `[renaming[1] for renaming in setp.to_rename]`

davinov (changes requested):
Thanks for this! I may have a few concerns though:

(it was not that easy after all :p)

server/src/weaverbird/backends/mongo_translator/steps/rename.py
raphaelvignes: Okay, I missed the fact that we could have multiple renamings.

davinov (approved):

PR #1281, opened 2022-03-11
feat(mongo back): aggregate step mongo back TCTC-1830

raphaelvignes:
Create the aggregate step in Mongo's backend translator.

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/steps/aggregate.py
+++ b/server/src/weaverbird/backends/mongo_translator/steps/aggregate.py
@@ -1,0 +1,51 @@
+from typing import Any, Dict, List, Union
+
+from weaverbird.pipeline.steps import AggregateStep
+
+
+def translate_aggregate(step: AggregateStep) -> List[Dict[str, Any]]:
+    idblock = {col: f'${col}' for col in step.on}
+    group: Dict[str, Dict] = {}
+    project: Dict[str, Union[str, int]] = {}
+    add_fields = {}
+    group['_id'] = idblock
+
+    for agg_f_step in step.aggregations:
+        cols = agg_f_step.columns
+        new_cols = agg_f_step.new_columns
+
+        if agg_f_step.agg_function == 'count':
+            for i in range(len(cols)):
+                group[new_cols[i]] = {'$sum': 1}
+        elif agg_f_step.agg_function == 'count distinct':
+            for i in range(len(cols)):
+                group[new_cols[i]] = {'$addToSet': f'${cols[i]}'}
+                add_fields[new_cols[i]] = {'$size': f'${new_cols[i]}'}
+        else:
+            for i in range(len(cols)):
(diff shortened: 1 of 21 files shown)


server/src/weaverbird/backends/mongo_translator/steps/aggregate.py
```diff
@@ -0,0 +1,51 @@
+from typing import Any, Dict, List, Union
+
+from weaverbird.pipeline.steps import AggregateStep
+
+
+def translate_aggregate(step: AggregateStep) -> List[Dict[str, Any]]:
```
davinov, line 6: The other steps are hinted with the return type `list` only.
In the front-end version, we had an alias called `MongoStep`. Maybe it could be semantically nice to have an TypeAlias here too (to make sure all steps have a similar signature).

server/src/weaverbird/backends/mongo_translator/steps/aggregate.py
```diff
@@ -0,0 +1,51 @@
...
+    idblock = {col: f'${col}' for col in step.on}
+    group: Dict[str, Dict] = {}
+    project: Dict[str, Union[str, int]] = {}
```
davinov, line 9: Maybe that type will work for this specific step, but `$project` in general  can take other forms than just `str` or `int`. It can hold other `dict`s, for nester fields or expressions (like `$cond`, `$substr`, etc.).
We may be more safe with just hinting `Dict[str, Any]` here (that's what it as in the front-end version)

server/src/weaverbird/backends/mongo_translator/steps/aggregate.py
```diff
@@ -0,0 +1,51 @@
...
+    add_fields = {}
+    group['_id'] = idblock
+
+    for agg_f_step in step.aggregations:
+        cols = agg_f_step.columns
+        new_cols = agg_f_step.new_columns
+
+        if agg_f_step.agg_function == 'count':
+            for i in range(len(cols)):
```
davinov, line 18: What do you thunk of keeping the original comments found in the mongo.ts translator?
I think this one was useful:
```python
// There is no `$count` operator in Mongo, we have to `$sum` 1s to get
// an equivalent result
```

davinov (approved):
Thanks for this efficient port, and great to see the tests case are working as intended :tada:
Here are very little type and comments considerations, optional before merge:

raphaelvignes:
Thanks for the review. 
I'll update the PR with your comments before merging

server/src/weaverbird/backends/mongo_translator/steps/aggregate.py
raphaelvignes: :+1:

server/src/weaverbird/backends/mongo_translator/steps/aggregate.py
raphaelvignes: :+1:

PR #1283, opened 2022-03-15
fix(sql): inclusion condition with a single value TCTC-1605

austil:
A pipeline with a filter step using the 'in' operator but with a single value like `column='origin', operator='in', value=['france']` was translated to `WHERE origin IN ('france',)` instead of `WHERE origin IN ('france')` (notice the useless trailing comma that's invalid SQL).


raphaelvignes (approved):
Thx pour le fix !

PR #1286, opened 2022-03-16
fix(convert-int-to-date): enable to convert year in int to date from types menu [TCTC-1895]

alice-sevin:
Avoid to have error when trying to convert a year in int to a date.

Before:
We convert to date using convert step it results to an error because translators are not abled to transform a int to a date (it think that it is a timestamp)
before

After:
We convert to date using to date step (same behaviour than for str columns conversion). We add an automatic str conversion in pandas and mongo translator to transform int year to str before casting to date
after

### DIFF
--- a/server/src/weaverbird/backends/pandas_executor/steps/todate.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/todate.py
@@ -11,11 +11,15 @@ def execute_todate(
     execute_pipeline: PipelineExecutor = None,
 ) -> DataFrame:
     format = step.format
-    timestamp_unit = 'ms' if df[step.column].dtype == 'int64' else None
-    # By default int are understood as timestamps, we prefer to convert it to %Y format if values are < 10_000
-    if format is None and df[step.column].dtype == 'int64' and df[step.column].max() < 10_000:
-        format = '%Y'
-        timestamp_unit = None
+    timestamp_unit = None
+
+    if format is None and df[step.column].dtype == 'int64':
+        # By default int are understood as timestamps, we prefer to convert it to %Y format if values are < 10_000
+        if df[step.column].max() < 10_000:
+            format = '%Y'
+        else:
+            # Timestamps are expected in ms (not in ns, which is pandas' default)
+            timestamp_unit = 'ms'
 
     datetime_serie = to_datetime(
         df[step.column], format=format, errors='coerce', unit=timestamp_unit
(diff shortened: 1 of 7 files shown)


server/src/weaverbird/backends/pandas_executor/steps/todate.py
```diff
@@ -11,11 +11,15 @@ def execute_todate(
...
-    if format is None and df[step.column].dtype == 'int64' and df[step.column].max() < 10_000:
-        format = '%Y'
-        timestamp_unit = None
+    timestamp_unit = None
+
+    if format is None and df[step.column].dtype == 'int64':
+        # By default int are understood as timestamps, we prefer to convert it to %Y format if values are < 10_000
+        if df[step.column].max() < 10_000:
+            format = '%Y'
+        else:
```
alice-sevin, line 20: :pray:

austil (commented):
Thx for initiating this PR !

- Pandas runtime look good
- Doubt on the Mongo implem, seems to be way different than Pandas, does it still handle the millisecond timestamp to date conversion ?
- SQL translation is missing, is this voluntary ?
- Overall, we should add E2E tests fixtures for those, I can show you how this work

alice-sevin:
> Thx for initiating this PR !
> 
> * Pandas runtime look good
> * Doubt on the Mongo implem, seems to be way different than Pandas, does it still handle the millisecond timestamp to date conversion ?
> * SQL translation is missing, is this voluntary ?
> * Overall, we should add E2E tests fixtures for those, I can show you how this work

SQL translation, viewed with @davinov , it's out of scope, to date step doesn't seems to work well anyway :cry: 
Mongos only convert int to str if exist (it doesn't work has panda so we didn't update the logic of to date himself, nothing change, it still work perfectly)!

You can add e2e tests if necessary, indeed i don't know how to add it myself!

src/lib/translators/mongo4.ts
```diff
@@ -44,6 +44,14 @@ function transformToDate(step: Readonly<ToDateStep>): MongoStep[] {
     format = format.replace('%B', '%b');
   }
 
+  const input = {
+    $cond: [
+      { $eq: [{ $type: $$(step.column) }, 'int'] },
```
davinov, line 49: @austil has a good point, we could implement here the same heuristic as we added in pandas.
Meaning:
```
$cond: [
  { $and: [ { $eq: [{ $type: $$(step.column) }, 'int'] }, { [step.column]: { $lt: 10_000 } } ] }
```

That may ensure same results for both backends regarding timestamps. What do you thinks ?

alice-sevin: Alright je corrige ça :pray:

alice-sevin: https://github.com/ToucanToco/weaverbird/pull/1286/commits/8d97f24cad1fbcd4614eb1e49869db2d8ad7e09c

PR #1287, opened 2022-03-17
feat(mongo back): if then else

raphaelvignes:
Port mongo front if then else step's to mongo back.

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/utils.py
+++ b/server/src/weaverbird/backends/mongo_translator/utils.py
@@ -1,7 +1,90 @@
+import datetime
 from re import sub
+from typing import List, Union
+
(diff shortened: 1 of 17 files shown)


server/src/weaverbird/backends/mongo_translator/utils.py
```diff
@@ -1,7 +1,93 @@
...
+                            truncate_to_day(f'${cond.column}'),
+                            truncate_to_day(
+                                translate_relative_date(cond.value)
+                                if isinstance(cond.value, RelativeDate)
+                                else cond.value
+                            ),
+                        ]
+                    }
+                }
+            return cond_expression
```
jGundermann, line 93: can we make it simplify here ?
i would suggest splitting in little function here.

PR #1289, opened 2022-03-17
feat(playground): flag to test the mongo backend (python implementation)

davinov:
Now that we've started to work on the python implementation of mongo translator, we should be able to test it in the playground!
Add `?mongo-python=enable` to the URL to test it.

Note: The `/mongo-translated` route continue to execute queries that are already translated by the front-end. This route will be deleted when we retire the front-end translator.


raphaelvignes (approved):
Seems good to me. 
Needs a rebase though

PR #1290, opened 2022-03-18
tests(snowflake): avoid connecting multiple times

davinov:
Snowflake tests are really slow, this is attempt to speed them up a bit.


raphaelvignes (approved):
So Nice thanks !

PR #1291, opened 2022-03-18
fix(mongo): support years as strings in python implem TCTC-1831

davinov:
This was working with the TypeScript implementation, so I added a test case and supported it in python one.

Also include a small fix to handle cases where years are in type "long" and not "int"

### DIFF
--- a/server/tests/backends/fixtures/todate/str_as_year_with_automatic_guess.json
+++ b/server/tests/backends/fixtures/todate/str_as_year_with_automatic_guess.json
@@ -1,0 +1,54 @@
+{
+  "exclude": [
+    "snowflake",
+    "mysql",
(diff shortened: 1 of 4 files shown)


server/tests/backends/fixtures/todate/str_as_year_with_automatic_guess.json
```diff
@@ -0,0 +1,54 @@
+{
```
raphaelvignes, line 1: Thx for adding the fixture :+1:

jGundermann (approved):
LGTM

raphaelvignes (approved):
OK! Nice.

raphaelvignes:
js unit tests need to be updated :S

PR #1296, opened 2022-04-19
feat(mongo back): unpivot step

fspot:
Implementation of the `unpivot` step in mongo backend translator (logic copied from `translators/mongo.ts`)


lukapeschke (approved):
LGTM

PR #1298, opened 2022-04-19
feat(mongo_translator): Implemented concatenate step

lukapeschke:
Translates the `concatenate` step from `../src/lib/translators/mongo.ts`. 

NOTE: The `$convert` operator is used here, which makes us dependent on mongo>=4.0


fspot (approved):
:+1:

PR #1299, opened 2022-04-19
feat(playground): Added a docker-compose spec allowing to run the playground with mongo

lukapeschke:


fspot (approved):
Merci pour ça :ok_hand:

davinov (approved):
Good idea! We could also do the same for postgres
:100: 

Note: this is great to test weaverbird, but as the images are built (the code not linked), it's not suited very well for development

lukapeschke:
> Good idea! We could also do the same for postgres 100

You're right, I'll add it.

> Note: this is great to test weaverbird, but as the images are built (the code not linked), it's not suited very well for development

I wasn't really sure about how to link the code since `yarn build` is quite long... Or I can just add a volume with the python sources, run `pip install` in the entrypoint and specify in the docs that the image needs to be rebuilt in order to test any frontend change ?

Sanix-Darker:
i think the idea with 'volumes' is good IMO !

lukapeschke:
@Sanix-Darker added the make targets you requested

Sanix-Darker (approved):

PR #1300, opened 2022-04-19
feat(mongo_translator): Implemented convert step [TCTC-1832]

fspot:
Translates the concatenate step from `../src/lib/translators/mongo.ts`

:warning: I had to handle some edge cases in order to make it iso with the pandas executor. There are probably still missing cases because type conversion in mongo and pandas have different rules. I also fear that adding these edge cases might decrease performances.

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/steps/convert.py
+++ b/server/src/weaverbird/backends/mongo_translator/steps/convert.py
@@ -1,0 +1,34 @@
+from typing import List, Union
+
+from weaverbird.backends.mongo_translator.steps.types import MongoStep
+from weaverbird.pipeline.steps import ConvertStep
+
+TYPE_MAP = {
+    'boolean': 'bool',
+    'date': 'date',
+    'float': 'double',
+    'integer': 'long',  # better to cast into 64-bit integers. 32-bit integers ('int') cannot be casted to dates
+    'text': 'string',
+}
+
+
+def translate_convert(step: ConvertStep) -> List[MongoStep]:
+    mongo_add_fields = {}
+    mongo_type = TYPE_MAP[step.data_type] if step.data_type in TYPE_MAP else ''
+    for column in step.columns:
+        # Mongo cannot cast integers into date but only long into date, so we
+        # manage the cast from int to long when needed.
+        _input: Union[dict, str]
+        if mongo_type == 'date':
+            _input = {
+                '$cond': [
+                    {'$eq': [{'$type': f'${column}'}, 'int']},
+                    {'$toLong': f'${column}'},
+                    f'${column}',
+                ]
+            }
+        else:
+            _input = f'${column}'
+        mongo_add_fields[column] = {'$convert': {'input': _input, 'to': mongo_type}}
+
+    return [{'$addFields': mongo_add_fields}]
--- a/server/src/weaverbird/backends/mongo_translator/steps/convert.py
+++ b/server/src/weaverbird/backends/mongo_translator/steps/convert.py
@@ -13,6 +13,9 @@ TYPE_MAP = {
 
 
 def translate_convert(step: ConvertStep) -> List[MongoStep]:
+    # /!\ we want to get the same results than with the mongo typescript
+    # translator, not than with the pandas executor.
+
     mongo_add_fields = {}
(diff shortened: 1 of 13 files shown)


Sanix-Darker (approved):
LGTM !

server/tests/backends/mongo_translator/test_mongo_translator_steps.py
```diff
@@ -69,5 +70,11 @@ def test_mongo_translator_pipeline(mongo_database, case_id, case_spec_file_path)
     df = pd.DataFrame(result)
     if '_id' in df:
         df.drop('_id', axis=1, inplace=True)
-    expected_df = pd.read_json(json.dumps(spec['expected']), orient='table')
+    expected_df = pd.read_json(
+        json.dumps(
+            spec[f'expected_{exec_type}' if f'expected_{exec_type}' in spec else 'expected']
```
davinov, line 75: I guess this is to have specific expectations for mongo in fixtures? But I don't see `expected_mongo` property anywhere. Are they created by the `retrieve_case` function?
If yes, have I have miss any mongo specific declaration in the JSON fixtures?

server/src/weaverbird/backends/mongo_translator/steps/convert.py
```diff
@@ -0,0 +1,87 @@
+from typing import List, Optional, Union
+
+from weaverbird.backends.mongo_translator.steps.types import MongoStep
+from weaverbird.pipeline.steps import ConvertStep
+
+TYPE_MAP = {
+    'boolean': 'bool',
+    'date': 'date',
+    'float': 'double',
+    'integer': 'long',  # better to cast into 64-bit integers. 32-bit integers ('int') cannot be casted to dates
```
davinov, line 10: :+1:

davinov (commented):
This is an interesting step to port!
Today, we have not the same results for the convert step in mongo and in pandas. Question is: do we want to have exact matching results, or is it OK if we have specific fixtures for this step for each backend.
I have no strong opinion on this :shrug:

server/tests/backends/mongo_translator/test_mongo_translator_steps.py
fspot: I modified this code because I thought I would end up adding a `expected_mongo` property, but in the end I managed to get the same results than with pandas.

server/src/weaverbird/backends/mongo_translator/steps/convert.py
```diff
@@ -0,0 +1,87 @@
...
+                        # Mongo cannot cast integers into date but only long into date,
+                        # so we manage the cast from int to long when needed:
+                        {
+                            'case': {'$eq': [{'$type': f'${column}'}, 'int']},
+                            'then': {'$toLong': f'${column}'},
+                        },
+                    ],
+                    'default': f'${column}',
+                }
+            }
```
davinov, line 46: Since last version, we use `todate` step instead of the `convert` step when someone tries to cast a column to date types. This makes this branch less critical to implement exactly right.

server/src/weaverbird/backends/mongo_translator/steps/convert.py
```diff
@@ -0,0 +1,87 @@
...
+        elif mongo_type == 'string':
+            # First, convert NaN to null:
+            _input = {
+                '$switch': {
+                    'branches': [{'case': {'$eq': [f'${column}', float('NaN')]}, 'then': None}],
+                    'default': f'${column}',
+                }
+            }
+            # Then, null will be converted to the literal string 'nan':
+            _on_null = 'nan'
```
davinov, line 74: This is a specific pandas' behavior to use the string 'nan'.
Maybe we would think than `null` / `None` is more appropriate? (because we can filter on it quickly after, and it denotes some missing data)
Maybe that's something we'd like to do differently than pandas?

server/tests/backends/mongo_translator/test_mongo_translator_steps.py
davinov: :ok_hand: I was afraid of missing something, I understand better now how this ended up there :+1:

davinov (approved):
:+1:

PR #1301, opened 2022-04-19
feat(mongo back): replace Step [TCTC-1832]

Sanix-Darker:
# What
Mongo-Back Replace Step from `../src/lib/translators/mongo.ts`


fspot (approved):

lukapeschke (approved):
:+1:

PR #1302, opened 2022-04-19
feat(mongo back): KeepColumn (a.k.a select) Step [TCTC-1832]

Sanix-Darker:
# What
Mongo-Back KeepColumn (a.k.a select) Step from `../src/lib/translators/mongo.ts`


fspot (approved):

lukapeschke (approved):
LGTM, thankf for the fixture!

PR #1304, opened 2022-04-26
[Proposition] feat(server/backends/pandas): Implement geographic dissolve step

lukapeschke:
POCing support for geographic transformation steps in the pandas executor


lukapeschke:
Everything here has been implemented

PR #1310, opened 2022-05-02
feat(python-mongo-translator): rank step [TCTC-2656]

Sanix-Darker:
# What
Implemented the rank step in python for the mongo-translator, following the logic from `./src/lib/translators/mongo.ts`

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/steps/rank.py
+++ b/server/src/weaverbird/backends/mongo_translator/steps/rank.py
@@ -5,6 +5,21 @@ from weaverbird.pipeline.steps import RankStep
 
 
 def translate_rank(step: RankStep) -> List[MongoStep]:
+
+    """
+    Here we define the order variable that will be used in the '$reduce' step
+    defined below. The order definition depends on the ranking method chosen.
+
+    Example of ranking output depending on method chosen:
+
+     - standard: [10, 15, 15, 15, 20, 20, 22] => [1, 2, 2, 2, 5, 5, 7]
+     - dense: [10, 15, 15, 15, 20, 20, 22] => [1, 2, 2, 2, 3, 3, 4]
+
+    Notes on special Mongo variables used in the '$reduce' step defined below:
+
+     - '$$this' refers to the element being processed
+     - '$$value' refers to the cumulative value of the expression
+    """
     vqb_var_order: MongoStep = {}
 
     if step.method == 'dense':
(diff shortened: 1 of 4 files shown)


server/src/weaverbird/backends/mongo_translator/steps/rank.py
```diff
@@ -0,0 +1,91 @@
+from typing import List
+
+from weaverbird.backends.mongo_translator.steps.types import MongoStep
+from weaverbird.pipeline.steps import RankStep
+
+
+def translate_rank(step: RankStep) -> List[MongoStep]:
+    vqb_var_order: MongoStep = {}
+
+    if step.method == 'dense':
```
lukapeschke, line 10: yould you pelase add the comments that were present in the typescript translator here ?

lukapeschke, line 10: yould you pelase add the comments that were present in the typescript translator here ?

server/src/weaverbird/backends/mongo_translator/steps/rank.py
```diff
@@ -0,0 +1,91 @@
...
+        vqb_var_order = {
+            '$cond': [
+                {'$ne': [f'$$this.{step.value_col}', '$$value.prevValue']},
```
lukapeschke, line 13: doesn't the double `$$` in a JS backtick string escape to a single `$` ?

lukapeschke, line 13: doesn't the double `$$` in a JS backtick string escape to a single `$` ?

server/src/weaverbird/backends/mongo_translator/steps/rank.py
Sanix-Darker: Yeah, you're right, it's better with those comments !

server/src/weaverbird/backends/mongo_translator/steps/rank.py
Sanix-Darker: If am not wrong, yes it escape to a singe $, but in this case it was `$$this` not `$${this}` so if we take an example to see well the difference, it could look like this : 
Screenshot from 2022-05-02 19-59-15

Sanix-Darker: https://github.com/ToucanToco/weaverbird/pull/1310#discussion_r863061239

lukapeschke (approved):

PR #1311, opened 2022-05-02
Mongo translator: implement `addmissingdates` and `dateextract` steps TCTC-2491

lukapeschke:
`addmissingdates` was required by test fixtures.

Note that this only targets mongo 5.

Note that the behvaiour here is ISO with the mongo translator, BUT there are few issues/inconsistencies that would need to be adressed imo:

-  mongo's `$week` operator returns the week of the year for a date, starting at 0 (ending at 52, included). Other backends (and the commonly used form) start at 1 and end at 53. Maybe this should be adressed later on to be consistent throughout all backends ?
- Several steps using mongo's `$switch` statements have a weird behaviour with null values: Since `{$lte:  [null, anyPositiveInteger]}` is `true` in mongo 5, we get `1` for `quarter` with `null`, `4` for `previousQuarter` and so on... This could probably easily be fixed by checking first if the value is not null

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/steps/addmissingdates.py
+++ b/server/src/weaverbird/backends/mongo_translator/steps/addmissingdates.py
@@ -1,0 +1,215 @@
+from typing import List
+
+from weaverbird.backends.mongo_translator.steps.types import MongoStep
+from weaverbird.pipeline.steps.addmissingdates import AddMissingDatesStep, DatesGranularity
+
+
+def _add_missing_dates_year(step: AddMissingDatesStep) -> List[MongoStep]:
+    groups = step.groups or []
+    # add missing dates by looping over the unique dates array and adding a given date
+    # if it's not already present in the original dataset
+    add_missing_years_as_dates: MongoStep = {
+        '$map': {
+            # loop over a sorted array of all years between min and max year
+            'input': {'$range': ['$_vqbMinYear', {'$add': ['$_vqbMaxYear', 1]}]},
+            # use a variable "currentYear" as cursor
+            'as': 'currentYear',
+            # and apply the following expression to every "currentYear"
+            'in': {
+                '$let': {
+                    # use a variable yearIndex that represents the index of the "currentYear"
+                    # cursor in the original array
+                    'vars': {
+                        'yearIndex': {'$indexOfArray': ['$_vqbArray._vqbYear', '$$currentYear']},
+                    },
+                    'in': {
+                        '$cond': [
+                            # if "currentYear" is found in the original array
+                            {'$ne': ['$$yearIndex', -1]},
+                            # just get the original document in the original array
+                            {'$arrayElemAt': ['$_vqbArray', '$$yearIndex']},
+                            # else add a new document with the missing date (we convert the year back to a date object)
+                            # and the group fields (every other field will be undefined)
+                            {
+                                **{col: f'$_id.{col}' for col in groups},
+                                step.dates_column: {'$dateFromParts': {'year': '$$currentYear'}},
+                            },
+                        ],
+                    },
+                }
+            },
+        }
+    }
+
+    return [
+        # Extract the year of the date in a new column
+        {'$addFields': {'_vqbYear': {'$year': f'${step.dates_column}'}}},
+        # Group by logic to create an array with all original documents, and get the
+        # min and max date of all documents confounded
+        {
+            '$group': {
+                '_id': {col: f'${col}' for col in groups} if groups else None,
+                '_vqbArray': {'$push': '$$ROOT'},
+                '_vqbMinYear': {'$min': '$_vqbYear'},
+                '_vqbMaxYear': {'$max': '$_vqbYear'},
+            },
+        },
+        # That at this stage that all the magic happens (cf. supra)
+        {'$addFields': {'_vqbAllDates': add_missing_years_as_dates}},
+    ]
+
+
+def _generate_date_from_parts(mongo_date: str, granularity: DatesGranularity) -> MongoStep:
+    date_from_parts = {'year': {'$year': mongo_date}}
+    if granularity in ('day', 'month'):
+        date_from_parts['month'] = {'$month': mongo_date}
+    if granularity == 'day':
+        date_from_parts['day'] = {'$dayOfMonth': mongo_date}
+
+    return date_from_parts
+
+
+def _add_missing_dates_day_or_month(step: AddMissingDatesStep) -> List[MongoStep]:
+    groups = step.groups or []
+    # Create a sorted array of all dates (in days) ranging from min to
+    # max dates found in the whole dataset. At this stage for the
+    # month granularity, we will get duplicate dates because we need
+    # to first generate all days one by one to increment the calendar
+    # safely, before converting every day of a given  month to the 1st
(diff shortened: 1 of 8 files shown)


server/src/weaverbird/backends/mongo_translator/steps/date_extract.py
```diff
@@ -0,0 +1,320 @@
+from copy import deepcopy
+from typing import List
+
+from weaverbird.backends.mongo_translator.steps.types import MongoStep
+from weaverbird.pipeline.steps import DateExtractStep
+
+
+def _truncate_date_to_day(expr: dict) -> MongoStep:
```
Sanix-Darker, line 8: On js : 
```javascript
export function truncateDateToDay(dateExpr: string | object): object {
```

i think, the translation should be : 
```suggestion
def _truncate_date_to_day(expr: Union[str, dict]) -> MongoStep:
```

What do you think ?

server/src/weaverbird/backends/mongo_translator/steps/date_extract.py
```diff
@@ -0,0 +1,320 @@
...
+    date_info: list
+    # For retrocompatibility
+    if step.operation:
+        date_info = [step.operation] if step.operation else step.date_info
+        new_columns = [
+            step.new_column_name if step.new_column_name else f"{step.column}_{step.operation}"
+        ]
+    else:
+        date_info = step.date_info.copy()
+        new_columns = deepcopy(step.new_columns)
```
Sanix-Darker, line 310: Not a blocking point, but just for curiosity, why not using the `.copy()` also here instead of `deepcopy` ? 🤔

server/tests/backends/mongo_translator/test_mongo_translator_steps.py
```diff
@@ -44,12 +46,27 @@ def mongo_database(mongo_server_port):
...
 
 
+def _sanitize_nulls(data: List[dict], inplace: bool = True) -> List[dict]:
+    """Replaces NaTs and NaNs with None"""
+    d = data if inplace else deepcopy(data)
+    for record in d:
+        for k, v in record.items():
+            # cannot use isinstance(v, pd.NaT) here, as pd.NaT is not
+            # a type
+            if v is pd.NaT:
```
Sanix-Darker, line 56: And for this : 
```
isinstance(fg, type(pd.NaT))
```


PS : not tested !

Sanix-Darker (commented):
Thanks for this amazing an LONG PR, i left some comments/suggestions, most of them are not blocking (just recommendations and questions) !

server/src/weaverbird/backends/mongo_translator/steps/addmissingdates.py
```diff
@@ -0,0 +1,215 @@
...
+                        '$cond': [
+                            # if "currentYear" is found in the original array
+                            {'$ne': ['$$yearIndex', -1]},
+                            # just get the original document in the original array
+                            {'$arrayElemAt': ['$_vqbArray', '$$yearIndex']},
+                            # else add a new document with the missing date (we convert the year back to a date object)
+                            # and the group fields (every other field will be undefined)
+                            {
+                                **{col: f'$_id.{col}' for col in groups},
+                                step.dates_column: {'$dateFromParts': {'year': '$$currentYear'}},
```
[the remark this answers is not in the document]
lukapeschke: I think black will automatically remove the parentheses :smile:

server/src/weaverbird/backends/mongo_translator/steps/addmissingdates.py
```diff
@@ -0,0 +1,215 @@
...
+                            # if "date" is found in the original array
+                            {'$ne': ['$$dateIndex', -1]},
+                            # just get the original document in the original array
+                            {'$arrayElemAt': ['$_vqbArray', '$$dateIndex']},
+                            # else add a new document with the missing
+                            # date and the group fieldds (every other
+                            # field will be undefined)
+                            {
+                                **{col: f'$_id.{col}' for col in groups},
+                                step.dates_column: '$$date',
```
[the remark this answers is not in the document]
lukapeschke: Same here

server/src/weaverbird/backends/mongo_translator/steps/addmissingdates.py
```diff
@@ -0,0 +1,215 @@
...
+                    '$divide': [{'$subtract': ['$_vqbMaxDay', '$_vqbMinDay']}, 60 * 60 * 24 * 1000],
+                },
+            },
+        },
+        # That at this stage that all the magic happens (cf. supra)
+        {'$addFields': {'_vqbAllDates': add_missing_dates}},
+    ]
+
+
+def translate_addmissingdates(step: AddMissingDatesStep) -> List[MongoStep]:
```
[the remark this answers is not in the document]
lukapeschke: I agree, but all other python step translation functions are called `translate_x`, so I wanted to be consistent with them

server/src/weaverbird/backends/mongo_translator/steps/addmissingdates.py
```diff
@@ -0,0 +1,215 @@
...
+        _add_missing_dates_year(step)
+        if step.dates_granularity == 'year'
+        else _add_missing_dates_day_or_month(step)
+    ) + [
+        # Get back to 1 row per document
+        {'$unwind': '$_vqbAllDates'},
+        # Change the root to get back to the document granularity
+        {'$replaceRoot': {'newRoot': '$_vqbAllDates'}},
+        # Remove remaining temporary column
+        {'$project': {'_vqbYear' if step.dates_granularity == 'year' else '_vqbDay': 0}},
```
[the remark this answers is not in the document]
lukapeschke: I believe black will remove the parentheses

server/src/weaverbird/backends/mongo_translator/steps/date_extract.py
lukapeschke: That's true, but I believe the TS version is out-of-date (or maybe I missed something). In this file, it is only used for dicts, so I wanted the type to be as specific as possible. Maybe we can change it if we also use it for strings at some point ? Or I can change it now, I don't have a strong opinion on this one

server/src/weaverbird/backends/mongo_translator/steps/date_extract.py
lukapeschke: Because I thought that `NewColumn` was an object but it's only an alias for `str`. You're right, I'll update it :+1:

server/tests/backends/mongo_translator/test_mongo_translator_steps.py
lukapeschke: You're right on this. This implementation has problems anyway, so I'll just drop the commit! And since this is needed elsewhere, I created a separate ticket and proposed an implementation here: https://github.com/ToucanToco/weaverbird/pull/1316

lukapeschke (commented):
Thanks a lot for the review @Sanix-Darker :pray: And sorry for the huge PR, I didn't see before starting that these steps are so tighly coupled in the fixtures. Please see inline responses :slightly_smiling_face:

Sanix-Darker (approved):
LGTM !

PR #1318, opened 2022-05-03
feat(python-mongo-translator): evolution step [TCTC-2657]

Sanix-Darker:
# What
Implemented the evolution step in python for the mongo-translator, following the logic from `./src/lib/translators/mongo.ts`

⚠️ However am facing a difference behaviour between results from pandas <> mongo when it comes to group, that's why i added a new group fixture only reserved to mongo, not sure if am missing something somewhere or it should behave like that 🤔 .

### DIFF
--- a/server/tests/backends/fixtures/evolution/perc_groups_mongo.json
+++ b/server/tests/backends/fixtures/evolution/perc_groups_mongo.json
@@ -1,0 +1,113 @@
+{
+  "exclude": [
+    "mysql",
+    "postgres",
+    "snowflake",
+    "pandas"
+  ],
+  "step": {
+    "pipeline": [
+      {
+        "name": "evolution",
+        "date_col": "DATE",
+        "value_col": "VALUE",
+        "evolution_type": "vsLastYear",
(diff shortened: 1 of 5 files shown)


server/tests/backends/fixtures/evolution/perc_groups_mongo.json
```diff
@@ -0,0 +1,113 @@
...
+        "DATE": "2020-01-01 00:00:01",
+        "VALUE": 4,
+        "COUNTRY": "France",
+        "POUET": null
+      },
+      {
+        "DATE": "2020-01-01 00:00:02",
+        "VALUE": 0.5,
+        "COUNTRY": "Italy",
+        "POUET": null
```
lukapeschke, line 109: This does not look right, have you been able to reproduce this behaviour with the typescript mongo translator ? Given the input, this should be the same as in `server/tests/backends/fixtures/evolution/perc_groups.json` . Can't spot a typo in the translator here though

lukapeschke (approved):

PR #1322, opened 2022-05-03
feat(mongo_translator): implement duration step [TCTC-2658]

fspot:
Just copy logic from mongo js translator.


lukapeschke (approved):
:+1:

PR #1327, opened 2022-05-04
fix(backend/pandas_executor): Do not normalize standard deviation and variance in statistics step TCTC-2701

### LINKED ISSUE #1323: [pandas executor] In `statistics` step, variance does not return expected result
This is due to `pandas`'s `var` method being normalized by 1 by default: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.var.html

The expected result could be obtained by setting `ddof` to 0, or calling `numpy.var` rather than `pandas`'s `var` method.

lukapeschke:
By default, pandas normalizes 'var' and 'std' by 1, which is unexpected behaviour and
inconsistent with mongo. This fixes this behaviour.

closes #1323

### DIFF
--- a/server/src/weaverbird/backends/pandas_executor/steps/statistics.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/statistics.py
@@ -1,17 +1,19 @@
+from typing import Callable, Union
+
 import numpy as np
-from pandas import DataFrame
+from pandas import DataFrame, Series
 
 from weaverbird.backends.pandas_executor.types import DomainRetriever, PipelineExecutor
 from weaverbird.pipeline.steps.statistics import DUMB_GROUPBY_COLUMN_NAME, StatisticsStep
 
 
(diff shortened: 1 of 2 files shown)


server/src/weaverbird/backends/pandas_executor/steps/statistics.py
```diff
@@ -1,17 +1,19 @@
...
-def statistic_to_pandas_method(s: str) -> str:
+def statistic_to_pandas_method(s: str) -> Union[str, Callable[[Series], float]]:
     if s == 'average':
         return 'mean'
     elif s == 'variance':
-        return 'var'
+        return lambda x: x.var(ddof=0)
     elif s == 'standard deviation':
-        return 'std'
+        return lambda x: x.std(ddof=0)
```
Sanix-Darker, line 16: 👍🏾

Sanix-Darker (approved):

PR #1328, opened 2022-05-04
feat(python-mongo-translator): comparetext step [TCTC-2695]

Sanix-Darker:
# What
Implemented the comparetext step in python for the mongo-translator, following the logic from `./src/lib/translators/mongo.ts`


lukapeschke (approved):
:+1:

PR #1330, opened 2022-05-04
test: added missing waterfall fixtures [TCTC-2695]

Sanix-Darker:
# What
Noticed, there were some missing fixtures for the waterfall step, that's why i added it here.


raphaelvignes (approved):
Thanks for adding that!

PR #1331, opened 2022-05-06
feat: mongo translator waterfall [TCTC-2695]

Sanix-Darker:
# What
Implemented the waterfall step in python for the mongo-translator, following the logic from `./src/lib/translators/mongo.ts`

⚠️ Discovered a small behaviour on typing str <-> int on some columns (for some rows containing years in integer) (GROUP_waterfall, LABEL_waterfall, TYPE_waterfall), same behaviour for the mongo.ts, maybe a deep investigation on that behaviour is needed for the future !

**Note:** Merge this Pull Request after this one : https://github.com/ToucanToco/weaverbird/pull/1330 where some fixture for the waterfall was added because there were not present before !


raphaelvignes (commented):
Congrats for this PR.
Some comments from my end. 
I did a local test with fixtures and the query seemed well generated. 
Did you managed to test it with on-render? thx !

raphaelvignes (approved):
Thx, it works \w on-render.

PR #1332, opened 2022-05-09
feat(python-mongo-translator): new dataset to test the waterfall step on the playground [TCTC-2695]

Sanix-Darker:
# WHAT
Just added a good dataset to be able to well test the waterfall Step on python mongo translator

Screenshot from 2022-05-09 17-02-14
Screenshot from 2022-05-09 17-04-57

PS: this new dataset is comming from a fixture `waterfall/with_groups.json` **AND**, i converted the YEAR column to text first for the waterfall to work !

### DIFF
--- a/playground/datastore/revenue-product-dataset.json
+++ b/playground/datastore/revenue-product-dataset.json
@@ -1,0 +1,141 @@
+{
+    "schema": {
+      "fields": [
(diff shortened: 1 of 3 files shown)


fspot (approved):
:ok_hand:

PR #1334, opened 2022-05-10
fix(mongo_translator/evolution): Replace manual date substraction with mongo's operator TCTC-2704

lukapeschke:
There is a bug in the evolution step: when using `$dateFromParts`, we only keep a hourly granularity, meaning
that datetimes with a sub-hourly precision did never match our predicted offset.

This will not be backported to the TypeScript translator, as it must support mongo<5, and will eventually be
replaced by the Python implementation.


jGundermann (approved):
LGTM

lukapeschke:
@davinov since you know mongo well, it'd be great if you could have a quick look at this :pray:

davinov (approved):
Great :+1: 
This is way better!

Also nice to be able to verify the results with the JSON fixture :heart_eyes_cat:

PR #1342, opened 2022-05-27
fix(mongo): remove $expr wrapping for ifthenelse condtions  TCTC-2118

davinov:
$match and $cond doesn't function extacly the same way: $cond takes an
expression while $match only a filter. Therefore, in $match, we need to
wrap some stuff with $expr, but not with $cond.

Added a fixture to test this behavior for mongo and pandas backends


lukapeschke (approved):
:+1:

PR #1346, opened 2022-06-03
fix: logs steps objects [TCTC-2844]

Sanix-Darker:
## WHAT

An error was occurring on laputa for this log on any step object, with orjson failing to parse the json extras !
Screenshot from 2022-06-03 13-17-52


jGundermann (approved):
LGTM

PR #1352, opened 2022-06-13
feat(playground/pandas): Add a geographic pipeline TCTC-2969

lukapeschke:
TCTC-2969


lukapeschke:
ESLint was a bit eager sorry :/ interseting stuff in `app.js` happens on l.410-553

playground/datastore/departements-regions-france.csv
```diff
@@ -0,0 +1,102 @@
+code_departement,nom_departement,code_region,nom_region
```
davinov, line 1: This may be typically the data that would be embedded in a geojson file.
maybe we can find a dataset that adds some other scalar info? Such as population? That could make nicer examples I think

server/playground.py
```diff
@@ -88,12 +89,18 @@ async def parse_request_json(request: Request):
...
 DOMAINS = {
     **{splitext(basename(csv_file))[0]: pd.read_csv(csv_file) for csv_file in csv_files},
     **{
         splitext(basename(json_file))[0]: pd.read_json(json_file, orient='table')
         for json_file in json_files
     },
+    **{
+        splitext(basename(geojson_file))[0]: gpd.read_file(geojson_file)
+        for geojson_file in geojson_files
+    },
```
davinov, line 103: :+1:

davinov (approved):

playground/datastore/departements-regions-france.csv
lukapeschke: You're right, but I couldn't find anything on data.gouv.fr (I wanted to make sure w'ere actually allowed to use the data). Anyway, I found a dataset on available COVID-19 vaccine doses. Is it okay with you if I add it as a third dataset and do a double join ? Otherwise I'll merge it with this one

davinov: No worries, it's just for demo purposes. So if you found better, good! But if it's not practical, you can also discard this comment

PR #1353, opened 2022-06-13
feat(pandas): delete step should not fail on non-existing cols

davinov:
To prevent getting errors in pipelines executed with the pandas backend when unused columns have been deleted.

This should already be the case for mongo and SQL backends.

### DIFF
--- a/server/src/weaverbird/backends/pandas_executor/steps/delete.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/delete.py
@@ -10,4 +10,4 @@ def execute_delete(
     domain_retriever: DomainRetriever = None,
     execute_pipeline: PipelineExecutor = None,
 ) -> DataFrame:
-    return df.drop([c for c in step.columns if c in df.column.names], axis=1)
+    return df.drop(step.columns, axis=1, errors='ignore')
--- a/server/src/weaverbird/backends/pandas_executor/steps/delete.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/delete.py
@@ -10,4 +10,4 @@ def execute_delete(
     domain_retriever: DomainRetriever = None,
     execute_pipeline: PipelineExecutor = None,
 ) -> DataFrame:
-    return df.drop(step.columns, axis=1)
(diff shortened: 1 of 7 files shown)


server/src/weaverbird/backends/pandas_executor/steps/delete.py
```diff
@@ -10,4 +10,4 @@ def execute_delete(
     domain_retriever: DomainRetriever = None,
     execute_pipeline: PipelineExecutor = None,
 ) -> DataFrame:
-    return df.drop(step.columns, axis=1)
+    return df.drop([c for c in step.columns if c in df.column.names], axis=1)
```
PrettyWood, line 13: ou juste un 
```suggestion
    return df.drop(step.columns, axis=1, errors='ignore')
```

davinov: :shrug: yes pk pas

server/tests/backends/fixtures/delete/non_existing_column.json
```diff
@@ -0,0 +1,55 @@
+{
+  "description": "it should not fail when deleting a column that does not exist in the dataset",
+  "exclude": [
+    "mysql"
+  ],
```
lukapeschke, line 5: This test fails on snowflake and postgres, maybe you'll want to exclude them too ? Or adapt them to mkae the behaviour consistent

davinov: I'm gonna try to adapt them: if we can prevent a fail everywhere, let's do it!

lukapeschke (commented):
LGTM! Thanks for fixing it everywhere! Since this will be added to laputa soon, maybe you can bump the version to `0.14.0` ? It'll avoid an extra PR to do the release. Also, could you please add a release note indicating the behaviour change for `delete` to `server/CHANGELOG.md` ?

PrettyWood:
Personally I prefer single commits for new versions. Way easier to find it back

lukapeschke (approved):
LGTM!

PR #1354, opened 2022-06-13
feat(server/backends/pandas): Implement geographic dissolve step TCTC-2630

lukapeschke:


server/tests/backends/fixtures/dissolve/pandas_advanced.json
```diff
@@ -0,0 +1,265 @@
...
+                        "agg_function": "count distinct",
+                        "new_columns": ["region_distinct_count"]
+                    }
+                ],
+                "include_nulls": true
+            }
+        ]
+    },
+    "input": {
+        "schema": "geojson",
```
davinov, line 32: The table schema standard does provide a way to declare a geometry column:
https://specs.frictionlessdata.io/table-schema/#geojson
Should we use this instead?

PR #1355, opened 2022-06-14
chore(server): v0.13.1

davinov:
With #1353

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog (weaverbird python package)
 
-## Unreleased
+## [0.14.0] - 2022-06-14
 
-- Fix: stop failing when deleting a non-existing columns
+- Changed: deleting a non-existing columns doesn't trigger an error (pandas, SQL)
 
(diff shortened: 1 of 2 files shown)


lukapeschke:
Since the behaviour changed, I'd rather have this as `0.14.0`. apart from that, LGTM

davinov:
> Since the behaviour changed, I'd rather have this as `0.14.0`. apart from that, LGTM

Hehe I thought of that too, and then I remember that we actually did this to prevent an error, so it felt more like a "fix" to me. (proof is, we want to merge this to fix a release candidate)
However I have no strong opinion on this, so if you still think it's better to bump the minor, I will gladly change

lukapeschke:
Hmm I see what you mean. I see it like this: if it was an internal project, bumping a patch version would be okay, since we're the only users, and it's a patch from our pov. However, since this is a public library, I'd rather be cautious, since other people could rely on the previous behaviour (unlikely but let's be good OSS citizens). So I'd prefer `0.14.0` here if you don't mind

lukapeschke (approved):

PR #1356, opened 2022-06-15
fix(server/pandas): Fix df_to_geodf

lukapeschke:
Since the contract of steps is to receive and emit DataFrames (and not GeoDataFrames),
fix the df_to_geodf method. Also update the pandas executor tests to provide a DataFrame rather
than a GeoDataFrame in case a geo step is tested.

### DIFF
--- a/server/src/weaverbird/backends/pandas_executor/geo_utils.py
+++ b/server/src/weaverbird/backends/pandas_executor/geo_utils.py
@@ -8,8 +8,12 @@ class UnsupportedGeoOperation(Exception):
 
 
 def df_to_geodf(df: pd.DataFrame) -> gpd.GeoDataFrame:
-    if not (isinstance(df, gpd.GeoDataFrame) and hasattr(df, 'geometry')):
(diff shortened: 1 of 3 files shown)


server/src/weaverbird/backends/pandas_executor/geo_utils.py
```diff
@@ -8,8 +8,12 @@ def __init__(self, msg: str) -> None:
 
 
 def df_to_geodf(df: pd.DataFrame) -> gpd.GeoDataFrame:
```
Sanix-Darker, line 10:
```suggestion
def df_to_geodf(df: pd.DataFrame | gpd.GeoDataFrame) -> gpd.GeoDataFrame:
```
?

lukapeschke: `gpd.GeoDataFrame` inherits `pd.DataFrame` (which can be annoying, because some method are overriden and change the behaviour slightly) so this is already induced

PR #1362, opened 2022-06-16
feat(step): add absolute value step

pifauc:
Added an absolute value step.

- [x] [WIP] Missing documentation as of today
- [x] [WIP] Missing tests


PrettyWood (commented):
Looks great!

PrettyWood (changes requested):
You should indeed add some doc but also tests

PrettyWood:
Thank you @pifauc. #1376 has been merged

PR #1366, opened 2022-06-17
chore: add ff for basemap as usual datasource [TCTC-3016]

alice-sevin:
Retrieve the ff defined in tucana in order to hide feature for basemap as usual datasource currently under development

Also need to review: https://github.com/ToucanToco/tucana/pull/12280


bloodstorms (approved):
🙆‍♀️

PR #1374, opened 2022-06-21
trigger 22.04 CI on redshift

lukapeschke:


lukapeschke:
Doesn't fix it :(

PR #1384, opened 2022-06-24
feat(sql): add absolute value

PrettyWood:


PrettyWood:
You're absolutely right

Sanix-Darker (approved):

PR #1388, opened 2022-06-27
chore: improve dev xp

PrettyWood:
- fix Makefile goals
- make `pytest` command work directly
- make `make test` work even without env variables
Screenshot 2022-06-27 at 14 30 40

### DIFF
--- a/server/tests/backends/sql_translator/test_sql_athena_translator_steps.py
+++ b/server/tests/backends/sql_translator/test_sql_athena_translator_steps.py
@@ -11,11 +11,14 @@ from weaverbird.backends.pypika_translator.dialects import SQLDialect
 from weaverbird.backends.pypika_translator.translate import translate_pipeline
 from weaverbird.pipeline import Pipeline
 
-_REGION = environ['ATHENA_REGION']
-_DB = environ['ATHENA_DATABASE']
-_ACCESS_KEY_ID = environ['ATHENA_ACCESS_KEY_ID']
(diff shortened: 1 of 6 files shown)


server/tests/backends/sql_translator/test_sql_athena_translator_steps.py
```diff
@@ -11,11 +11,14 @@
...
-_SECRET_ACCESS_KEY = environ['ATHENA_SECRET_ACCESS_KEY']
-_OUTPUT = environ['ATHENA_OUTPUT']
+try:
+    _REGION = environ['ATHENA_REGION']
+    _DB = environ['ATHENA_DATABASE']
+    _ACCESS_KEY_ID = environ['ATHENA_ACCESS_KEY_ID']
+    _SECRET_ACCESS_KEY = environ['ATHENA_SECRET_ACCESS_KEY']
+    _OUTPUT = environ['ATHENA_OUTPUT']
+except KeyError as e:
+    pytestmark = pytest.mark.skip(reason=f'Missing environ variable: {e}')
```
Sanix-Darker, line 21: 👍🏾

Sanix-Darker (approved):
Thanks for this PR !

lukapeschke (commented):
2 remarks:

1. Are we sure we want to skip tests if the env vars are not set ? We might miss a misconfigured CI that way
2. I'm not sure we should force `poetry run` in the Makefile (or maybe create a different target for it ?) because of the reasons explained [here](https://github.com/ToucanToco/toucan-connectors/pull/606#discussion_r893188455)

PR #1394, opened 2022-06-29
fix(backend/fixtures/athena): Update fixtures according to recent dataset update

lukapeschke:


raphaelvignes (approved):
thx

PR #1404, opened 2022-07-07
Harmonize postgres integration tests

raphaelvignes:
Perform the same integration test logic as other pypika translators.


raphaelvignes:
BTW, we can also run these tests on our managed postgres.

lukapeschke (approved):

PR #1407, opened 2022-07-08
feat(backend/pypika): Implement JOIN step TCTC-3156

lukapeschke:


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -667,6 +669,65 @@ def ifthenelse(
...
+    def _get_join_type(join_type: Literal['left', 'inner', 'left outer']) -> JoinType:
+        match join_type:
+            case 'left':
+                return JoinType.left
+            case 'left outer':
+                return JoinType.left_outer
+            case 'inner':
+                return JoinType.inner
+
+    def join(
```
raphaelvignes, line 682: good stuff! 
I see one thing that I guess we wont be able to solve: after a custom step, we'll lose the list of columns.
Thus I guess join after a custom step won't work.

lukapeschke: Hmm I think it will work, it's just that our output columns will be weird (something like `["a", "b", "*"]`)... I'll add a fixture!

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -667,6 +669,75 @@ def ifthenelse(
...
+            .select(*left_cols, *right_cols)
+            .join(right_table, self._get_join_type(step.type))
+            .on(
+                Criterion.all(
+                    Field(f[0], table=left_table) == Field(f[1], table=right_table) for f in step.on
+                )
+            )
+            # Order of results is not consistent depending on the SQL Engine (inconsistencies
+            # observed with Athena and BigQuery).
+            .orderby(*(c[0] for c in step.on))
```
raphaelvignes, line 733: Should we document this in the weaverbird's doc?

raphaelvignes (approved):
Fantastic job! Thanks :+1:

server/src/weaverbird/backends/pypika_translator/translators/base.py
lukapeschke: done in https://github.com/ToucanToco/weaverbird/pull/1407/commits/096b61ab1996a3831010b8c3b7015f7dcf5bff8d

server/src/weaverbird/backends/pypika_translator/translators/base.py
lukapeschke: Done in https://github.com/ToucanToco/weaverbird/pull/1407/commits/ee77a6b88ec559ab0df1a928728d6012bd6eb4af . I think all the docs deserve a refresh though :sweat_smile:

raphaelvignes:
@lukapeschke your fixture seems to fails the CI :S

lukapeschke:
@raphaelvignes had to create a separate fixture for bigquery because the dataset name is required :disappointed:

PR #1408, opened 2022-07-08
feat(backend/tests): Add support for fixtures in YAML format TCTC-3159

lukapeschke:
TCTC-3159


server/tests/utils.py
```diff
@@ -115,9 +124,7 @@ def _datetime_parser(dct):
                         print(es)
             return dct
 
-        spec = json.loads(spec_file.read(), object_hook=_datetime_parser)
-    else:
-        spec = json.loads(spec_file.read())
-    spec_file.close()
+        # Sub-optimal but still fast enough for tests
+        spec = json.loads(json.dumps(spec), object_hook=_datetime_parser)
```
Sanix-Darker, line 128: indeed... this could slow our tests (not huge need on change that but, i wish we could handle the `_datetime_parser` somewhere else ?)

Sanix-Darker (approved):

PR #1411, opened 2022-07-11
test(server/pypika): Add concatenate fixture for pypika

lukapeschke:


raphaelvignes:
Good to know, I guess we'd have to override the method for Redshift and find a way to rename the column (or else, special case in the frontend).

raphaelvignes (approved):
Nice!

PR #1414, opened 2022-07-11
feat(backend/pypika): Implement APPEND step TCTC-3157

lukapeschke:
TCTC-3157

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -307,19 +307,35 @@ class SQLTranslator(ABC):
             column_lists.append(pipeline_ctx.columns)
             builder = pipeline_ctx.builder
 
-        # NOTE: Not sure if we should run this check or just let the UNION ALL happen...
-        # if not all(cols == column_lists[0] for cols in column_lists[1:]):
-        #     raise ValueError(
-        #         "Columns of all appended datasets must be the same than the current dataset."
-        #         f''' Expected "{', '.join(column_lists[0])}"'''
-        #     )
+        # By default, UNION ALL will append columns by index in SQL, and only append as much columns
+        # as available in the first SELECT statement. Also, the columns are appended in the order of
+        # the select statement, which is not what we want either (a merge by name is probably what's
+        # expected by the user)
+        columns_to_add = []
+        for column_list in column_lists:
+            for col in column_list:
+                if col not in columns and col not in columns_to_add:
+                    columns_to_add.append(col)
 
-        query = self.QUERY_CLS.from_(prev_step_name).select(*columns)
+        all_columns = columns + columns_to_add
+
+        query = self.QUERY_CLS.from_(prev_step_name).select(
+            *columns, *(LiteralValue('NULL').as_(col) for col in columns_to_add)
+        )
         for table, column_list in zip(tables, column_lists):
-            query = query.union_all(self.QUERY_CLS.from_(table).select(*column_list))
+            query = query.union_all(
+                self.QUERY_CLS.from_(table).select(
+                    *(
+                        # Selecting either the column from the dataset if it is available, or NULL
+                        # AS "col" otherwise. We iterate over all_columns rather than column_list in
+                        # order to have a merge by name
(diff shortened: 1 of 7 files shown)


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -271,6 +273,52 @@ def aggregate(
...
+            tables.append(pipeline_ctx.table_name)
+            column_lists.append(pipeline_ctx.columns)
+            builder = pipeline_ctx.builder
+
+        # NOTE: Not sure if we should run this check or just let the UNION ALL happen...
+        # if not all(cols == column_lists[0] for cols in column_lists[1:]):
+        #     raise ValueError(
+        #         "Columns of all appended datasets must be the same than the current dataset."
+        #         f''' Expected "{', '.join(column_lists[0])}"'''
+        #     )
```
lukapeschke, line 313: Behaviour here needs to be determined. I see three options:
* Leave things as is -> Most SQL-ish
* Uncomment this section -> Safer, more annoying for users
* Check all columns that'll be required in the final dataset by having a look at everyhting we resolved from the different pipelines, and cheat by doing literal selects:
  `SELECT col1_dataset1, NULL AS "col1_dataset2"` -> Here, we'd create a column containing NULLs, so we could append other columns from other datasets to it. 
:warning: I'm not sure this works, need to test it first

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -274,6 +275,68 @@ def aggregate(
...
+        )
+        for table, column_list in zip(tables, column_lists):
+            query = query.union_all(
+                self.QUERY_CLS.from_(table).select(
+                    *(
+                        # Selecting either the column from the dataset if it is available, or NULL
+                        # AS "col" otherwise. We iterate over all_columns rather than column_list in
+                        # order to have a merge by name
+                        col if col in column_list else LiteralValue('NULL').as_(col)
+                        for col in all_columns
```
raphaelvignes, line 333: Great for user's experience. Here you hide a bit of the complexity to user (in SQL she'd have to order correctly her columns), I agree with the decision.

raphaelvignes (approved):
great work. Good to see it was not to painful to implement :)

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -274,6 +275,68 @@ def aggregate(
...
+                    )
+                )
+            )
+
+        return StepContext(query.orderby(*all_columns), all_columns, builder)
```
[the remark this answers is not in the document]
lukapeschke: I'm afraid it might be a bit expensive, but maybe we could add this to `StepContext`'s `materialize` method ? That'd order the final dataset

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -274,6 +275,68 @@ def aggregate(
...
+
+        # By default, UNION ALL will append columns by index in SQL, and only append as much columns
+        # as available in the first SELECT statement. Also, the columns are appended in the order of
+        # the select statement, which is not what we want either (a merge by name is probably what's
+        # expected by the user)
+        columns_to_add = []
+        for column_list in column_lists:
+            for col in column_list:
+                if col not in columns and col not in columns_to_add:
+                    columns_to_add.append(col)
```
Sanix-Darker, line 318: What about using set here ? instead of looping over columns... 🤔 
```suggestion
         columns_to_add = list((set(columns_to_add).union(set(column_lists))).difference(columns) )
```
What do you think about something like this ? 🤔 


PS : not tested (it may be need to be updated)

server/src/weaverbird/backends/pypika_translator/translators/base.py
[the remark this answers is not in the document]
lukapeschke: On a second thought, it's not a good idea, as for simple queries, users probably expect their dataset's row order to be conserved. And there's a `sort` step, which would get overridden if we had a final sort on all columns

PR #1417, opened 2022-07-13
feat(frontend): Implemented form for Dissolve step TCTC-2962

lukapeschke:


jGundermann (commented):
LGTM

src/components/stepforms/DissolveStepForm.vue
```diff
@@ -0,0 +1,115 @@
...
+      :automatic-new-field="false"
+      data-path=".aggregations"
+      :errors="errors"
+      :available-variables="availableVariables"
+      :variable-delimiters="variableDelimiters"
+    />
+    <CheckboxWidget
+      class="keepNullsCheckbox"
+      label="Include null values in results"
+      v-model="editedStep.include_nulls"
```
davinov, line 36: It's been a while since we made a new step so I forgot how we usually handle case in there. Do we have something that does snake to camel somewhere? Or do we keep snake in the front-end too in other steps?

src/components/stepforms/DissolveStepForm.vue
```diff
@@ -0,0 +1,115 @@
...
+    <MultiselectWidget
+      class="groupbyColumnsInput"
+      v-model="editedStep.groups"
+      name="Group rows by..."
+      :options="columnNames"
+      placeholder="Add columns"
+      data-path=".groups"
+      :errors="errors"
+      :available-variables="availableVariables"
+      :variable-delimiters="variableDelimiters"
```
davinov, line 18: Either use dashes everywhere or camelCase like L7, but maybe not both in the same template?
(We should have an eslint rule for this :p)

tests/unit/action-toolbar.spec.ts
```diff
@@ -16,7 +16,11 @@ describe('ActionToolbar', () => {
   it('should instantiate action toolbar buttons with right classes', () => {
     const store = setupMockStore(
       // TO_REMOVE: add a ff for geo category until development is done
-      buildStateWithOnePipeline([], { featureFlags: { BASEMAP_AS_USUAL_DATASOURCE: 'enable' } }),
+      buildStateWithOnePipeline([], {
+        featureFlags: { BASEMAP_AS_USUAL_DATASOURCE: 'enable' },
```
davinov, line 20: Not related to this PR as it seems to come from before but: should the feature flag be disabled in that describe block, and enabled only in the `describe('with feature flags', () => {` block?
Otherwise seems like we never test the case without the feature flag

src/components/stepforms/DissolveStepForm.vue
```diff
@@ -0,0 +1,115 @@
...
+     * to the automatically generated newcolumn name
+     */
+    const newcolumnOccurences: { [prop: string]: number } = {};
+    for (const agg of this.editedStep.aggregations) {
+      agg.newcolumns = [...agg.columns];
+      for (const c of agg.newcolumns) {
+        newcolumnOccurences[c] = (newcolumnOccurences[c] || 0) + 1;
+      }
+    }
+    this.$$super.submit();
```
davinov, line 114: Would be nice to test this part (or factor it with the aggregation step)

davinov (approved):
:+1: Good! Thanks for this PR, here are a few comments but nothing blocking the merge

src/components/stepforms/DissolveStepForm.vue
lukapeschke: Haha sorry I copied this from https://github.com/ToucanToco/weaverbird/blob/5fd8b5a862dfb4f4a4f1b79a8f645a4db9b44e61/src/components/stepforms/AggregateStepForm.vue#L31 

I can't override the prop names of `ListWidget` without changing everything and I don't think I can use kebab-case for variable names in javascript ? So I'm not really sure about what to do here

src/components/stepforms/DissolveStepForm.vue
lukapeschke: I think it's up to the pydantic models to provide a camel case alias, I'll update `DissolveStep`

src/components/stepforms/DissolveStepForm.vue
lukapeschke: I'll factor it. Is there a good place for utils such as this one ?

PR #1418, opened 2022-07-13
feat: ifthenelse pypika [TCTC-3294]

Sanix-Darker:
## WHAT
- IfthenElse on Dates Step
- IfthenElse on Regex Step

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/operators.py
+++ b/server/src/weaverbird/backends/pypika_translator/operators.py
@@ -6,9 +6,6 @@ class RegexOp(Enum):
     SIMILAR_TO = auto()
     CONTAINS = auto()
     REGEXP_LIKE = auto()
-    REGEXP_SUBSTR = auto()
-    REGEXP_EXTRACT = auto()
-    REGEXP_MATCH = auto()
     REGEXP_CONTAINS = auto()
 
 
--- a/server/src/weaverbird/backends/pypika_translator/operators.py
+++ b/server/src/weaverbird/backends/pypika_translator/operators.py
@@ -2,9 +2,9 @@ from enum import Enum, auto
 
 
 class RegexOp(str, Enum):
-    REGEXP = auto()
-    SIMILAR_TO = auto()
-    CONTAINS = auto()
+    REGEXP = "REGEXP"
+    SIMILAR_TO = "SIMILAR TO"
+    CONTAINS = "CONTAINS"
     REGEXP_LIKE = "REGEXP_LIKE"
     REGEXP_CONTAINS = "REGEXP_CONTAINS"
     NOT_REGEXP_LIKE = "NOT REGEXP_LIKE"
--- a/server/src/weaverbird/backends/pypika_translator/operators.py
+++ b/server/src/weaverbird/backends/pypika_translator/operators.py
@@ -5,6 +5,11 @@ class RegexOp(Enum):
     REGEXP = auto()
     SIMILAR_TO = auto()
     CONTAINS = auto()
+    REGEXP_LIKE = auto()
+    REGEXP_SUBSTR = auto()
+    REGEXP_EXTRACT = auto()
+    REGEXP_MATCH = auto()
+    REGEXP_CONTAINS = auto()
 
 
 class FromDateOp(Enum):
--- a/server/src/weaverbird/backends/pypika_translator/operators.py
+++ b/server/src/weaverbird/backends/pypika_translator/operators.py
@@ -1,12 +1,14 @@
 from enum import Enum, auto
 
 
-class RegexOp(Enum):
+class RegexOp(str, Enum):
     REGEXP = auto()
     SIMILAR_TO = auto()
     CONTAINS = auto()
-    REGEXP_LIKE = auto()
-    REGEXP_CONTAINS = auto()
+    REGEXP_LIKE = "REGEXP_LIKE"
+    REGEXP_CONTAINS = "REGEXP_CONTAINS"
+    NOT_REGEXP_LIKE = "NOT REGEXP_LIKE"
+    NOT_REGEXP_CONTAINS = "NOT REGEXP_CONTAINS"
 
 
 class FromDateOp(Enum):
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -1279,9 +1279,7 @@ def _compliant_regex(pattern: str, dialect: SQLDialect) -> str:
     those %
     """
 
-    if dialect in [SQLDialect.POSTGRES, SQLDialect.REDSHIFT]:
-        return f"%%{pattern}%%"
-    elif dialect in [SQLDialect.ATHENA, SQLDialect.GOOGLEBIGQUERY]:
+    if dialect in [SQLDialect.ATHENA, SQLDialect.GOOGLEBIGQUERY]:
         return f"{pattern}"
 
     return f"%{pattern}%"
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -667,7 +667,6 @@ class SQLTranslator(ABC):
                                 column_field.wrap_constant(compliant_regex),
                             )
                         case RegexOp.REGEXP_LIKE:
-
                             compliant_regex = (
                                 compliant_regex.replace('%', '')
                                 if self.DIALECT in [SQLDialect.ATHENA]
@@ -675,6 +674,28 @@ class SQLTranslator(ABC):
                             )
                             return RegexpFunction(
                                 RegexpFunction.REGEXP_LIKE,
+                                column_field,
+                                column_field.wrap_constant(compliant_regex),
+                            )
+                        case RegexOp.REGEXP_CONTAINS:
+                            compliant_regex = (
+                                compliant_regex.replace('%', '')
+                                if self.DIALECT in [SQLDialect.GOOGLEBIGQUERY]
+                                else compliant_regex
+                            )
+                            return RegexpFunction(
+                                RegexpFunction.NOT_REGEXP_CONTAINS,
+                                column_field,
+                                column_field.wrap_constant(compliant_regex),
+                            )
+                        case RegexOp.REGEXP_LIKE:
+                            compliant_regex = (
+                                compliant_regex.replace('%', '')
+                                if self.DIALECT in [SQLDialect.ATHENA]
+                                else compliant_regex
+                            )
+                            return RegexpFunction(
+                                RegexpFunction.NOT_REGEXP_LIKE,
                                 column_field,
                                 column_field.wrap_constant(compliant_regex),
                             )
@@ -1254,11 +1275,10 @@ class ParseDate(functions.Function):  # type: ignore[misc]
 
 
 class RegexpFunction(functions.Function):
-    REGEXP_EXTRACT = "REGEXP_EXTRACT"
-    REGEXP_SUBSTR = "REGEXP_SUBSTR"
     REGEXP_LIKE = "REGEXP_LIKE"
-    REGEXP_MATCH = "REGEXP_MATCH"
     REGEXP_CONTAINS = "REGEXP_CONTAINS"
+    NOT_REGEXP_LIKE = "NOT REGEXP_LIKE"
+    NOT_REGEXP_CONTAINS = "NOT REGEXP_CONTAINS"
 
     def __init__(
         self, keyword: str, term: str | Field, regexp_expression: str, alias: str | None = None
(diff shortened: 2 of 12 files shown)


server/src/weaverbird/backends/pypika_translator/operators.py
```diff
@@ -5,6 +5,11 @@ class RegexOp(Enum):
     REGEXP = auto()
     SIMILAR_TO = auto()
     CONTAINS = auto()
+    REGEXP_LIKE = auto()
+    REGEXP_SUBSTR = auto()
+    REGEXP_EXTRACT = auto()
+    REGEXP_MATCH = auto()
+    REGEXP_CONTAINS = auto()
```
jGundermann, line 12: These cases are not tested yet ?
Shouldn't we add the corresponding unit test ?

jGundermann (changes requested):

server/src/weaverbird/backends/pypika_translator/operators.py
Sanix-Darker: Only `REGEXP_LIKE` is tested in this Pull Request, is it better for to remove the others ? 🤔 
Because i added them after checking some regex SQL syntax available !

jGundermann: I mean making something similar to this :
https://github.com/ToucanToco/weaverbird/blob/master/server/tests/backends/sql_translator_unit_tests/test_filter_translators.py#L124

Sanix-Darker: Alright, i see, you mean something like this : https://github.com/ToucanToco/weaverbird/pull/1418/commits/e8c542b9f513e4f415570dff7b89ad129cbc78a5 ?

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1126,6 +1176,18 @@ def __init__(self, term: str | Field, date_format: str, alias: str | None = None
...
+class RegexpFunction(functions.Function):
+    REGEXP_LIKE = "REGEXP_LIKE"
+    REGEXP_CONTAINS = "REGEXP_CONTAINS"
+    NOT_REGEXP_LIKE = "NOT REGEXP_LIKE"
+    NOT_REGEXP_CONTAINS = "NOT REGEXP_CONTAINS"
+
+    def __init__(
+        self, keyword: str, term: str | Field, regexp_expression: str, alias: str | None = None
+    ) -> None:
+        super().__init__(keyword, term, regexp_expression, alias=alias)
```
lukapeschke, line 1188: Shouldn't `keyword` be a RegexOp ? Otherwise, I don't think this constructor is required, since it only calls `super().__init__`

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -535,23 +533,52 @@ def _get_single_condition_criterion(
...
                     match self.REGEXP_OP:
                         case RegexOp.REGEXP:
                             return column_field.regexp(condition.value)
                         case RegexOp.SIMILAR_TO:
+
+                            compliant_regex = (
+                                compliant_regex.replace('%', '%%')
+                                if self.DIALECT in [SQLDialect.POSTGRES, SQLDialect.REDSHIFT]
+                                else compliant_regex
+                            )
```
lukapeschke, line 548: Why is this needed ?

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1126,6 +1176,18 @@ def __init__(self, term: str | Field, date_format: str, alias: str | None = None
         super().__init__("PARSE_DATE", term, date_format, alias=alias)
 
 
+class RegexpFunction(functions.Function):
+    REGEXP_LIKE = "REGEXP_LIKE"
+    REGEXP_CONTAINS = "REGEXP_CONTAINS"
+    NOT_REGEXP_LIKE = "NOT REGEXP_LIKE"
+    NOT_REGEXP_CONTAINS = "NOT REGEXP_CONTAINS"
```
lukapeschke, line 1183: I think this should live in the `RegexOp` enum, and it should be changed to `RegexOp(str, Enum)`

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -562,16 +589,43 @@ def _get_single_condition_criterion(
                         case RegexOp.REGEXP:
                             return column_field.regexp(condition.value).negate()
                         case RegexOp.SIMILAR_TO:
+                            compliant_regex = (
+                                compliant_regex.replace('%', '%%')
+                                if self.DIALECT in [SQLDialect.POSTGRES, SQLDialect.REDSHIFT]
+                                else compliant_regex
+                            )
```
lukapeschke, line 596: If this is intended for escaping, it does not work, an `ESCAPE` statement must be used:

```
playground_db=# select "Transaction_date", "City" from sales WHERE "sales"."State" SIMILAR TO '%Aviv';
  Transaction_date   |   City
---------------------+----------
 2009-01-04 13:17:00 | Tel Aviv
(1 row)

playground_db=# select "Transaction_date", "City" from sales WHERE "sales"."State" SIMILAR TO '%%Aviv';
  Transaction_date   |   City
---------------------+----------
 2009-01-04 13:17:00 | Tel Aviv
(1 row)

playground_db=# select "Transaction_date", "City" from sales WHERE "sales"."State" SIMILAR TO '%Aviv' ESCAPE '%';
 Transaction_date | City
------------------+------
(0 rows)
```

jGundermann (approved):
Thx for the changes!

LGTM :cat:

lukapeschke (changes requested):

server/src/weaverbird/backends/pypika_translator/translators/base.py
Sanix-Darker: Make sense, updated here : https://github.com/ToucanToco/weaverbird/pull/1418/commits/748984ee43328c4f6d25ac0abaf82c1f2bf513a7

server/src/weaverbird/backends/pypika_translator/translators/base.py
Sanix-Darker: True, the constructor is not needed !
For the name i rather prefer this class to stay as `RegexFunction` to not collide with the `RegexOp` class from `operators` what do you think ?
I updated that here : https://github.com/ToucanToco/weaverbird/pull/1418/commits/a3c4751870350c89f8da3617ccad6f4ea3dfc30d

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -535,23 +533,52 @@ def _get_single_condition_criterion(
...
 
             case MatchCondition():
+                compliant_regex = _compliant_regex(condition.value)
 
                 if condition.operator == "matches":
                     match self.REGEXP_OP:
                         case RegexOp.REGEXP:
                             return column_field.regexp(condition.value)
                         case RegexOp.SIMILAR_TO:
+
```
lukapeschke, line 543: I feel like there should be an abstract method returning the match operator: Here, we're guessing what the child class needs and we have cases dependent on child classes (like on l.544, 561 or 572). I believe that it would lead to a better separation of concerns and easier to read code (and probably less LoC if we eliminate some cases).

We can keep this as is for this PR and do the refactoring later though

server/tests/backends/fixtures/filter/dates_from_until_pypika.json
```diff
@@ -20,7 +20,7 @@
             {
               "column": "brewing_date",
               "operator": "until",
-              "value": "2022-01-07 00:00:00"
+              "value": "2022-01-07 01:00:00"
```
lukapeschke, line 23: Why did the input have to be changed ? We should keep the same contract

server/tests/backends/fixtures/ifthenelse/dates_conditions_pypika.json
```diff
@@ -0,0 +1,121 @@
...
+            "column": "brewing_date",
+            "operator": "from",
+            "value": "2022-01-05 00:00:00"
+        },
+        "then": "'tintin'",
+        "else": {
+          "if": {
+              "column": "brewing_date",
+              "operator": "until",
+              "value": "2022-01-07 00:00:00"
```
lukapeschke, line 24:
```suggestion
              "value": "2022-01-03 00:00:00"
```

We'll never get `tata` with a date later than the date in the first predicate

server/tests/backends/fixtures/filter/dates_from_until_pypika.json
Sanix-Darker: https://github.com/ToucanToco/weaverbird/pull/1418/commits/8d6d27c067adf502a2ea3f6bd619470e99c8f58f

server/src/weaverbird/backends/pypika_translator/operators.py
```diff
@@ -1,10 +1,14 @@
 from enum import Enum, auto
 
 
-class RegexOp(Enum):
+class RegexOp(str, Enum):
     REGEXP = auto()
     SIMILAR_TO = auto()
     CONTAINS = auto()
```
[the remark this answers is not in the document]
Sanix-Darker: Cleaned here : https://github.com/ToucanToco/weaverbird/pull/1418/commits/6ecbbd6737cfa6fbe9131eeb6b2a4853fb76b6a1

server/src/weaverbird/backends/pypika_translator/translators/base.py
Sanix-Darker: removed here : https://github.com/ToucanToco/weaverbird/pull/1418/commits/d28d51bfb69073501f9bb829c725240ed122c65d

server/src/weaverbird/backends/pypika_translator/translators/base.py
lukapeschke: Thanks! I still don't understand why double `%` are required for postgres and redshift for `SIMILAR TO` though :thinking:

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -683,10 +718,33 @@ def _get_single_condition_criterion(
...
-                if condition.operator == "until":
-                    return column_field <= condition.value
-                elif condition.operator == "from":
-                    return column_field >= condition.value
+
+                if isinstance(condition.value, (RelativeDate, datetime, str)):
+                    if isinstance(condition.value, RelativeDate):
+                        value_str_time = (
+                            evaluate_relative_date(condition.value)
+                            .astimezone()
```
lukapeschke, line 726: This converts the TZ to the system time, and forces the TZ on naive objects. I thought we agreed on UTC ?

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -683,10 +718,33 @@ def _get_single_condition_criterion(
...
+                            .strftime("%Y-%m-%d %H:%M:%S")
+                        )
+                    elif isinstance(condition.value, datetime):
+                        value_str_time = condition.value.astimezone().strftime("%Y-%m-%d %H:%M:%S")
+                    else:
+                        from dateutil import parser as dateutil_parser
+
+                        value_str_time = (
+                            dateutil_parser.parse(condition.value)
+                            .astimezone()
```
lukapeschke, line 736: This converts the TZ to the system time, and forces the TZ on naive objects. I thought we agreed on UTC ?

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -683,10 +718,33 @@ def _get_single_condition_criterion(
...
+
+                if isinstance(condition.value, (RelativeDate, datetime, str)):
+                    if isinstance(condition.value, RelativeDate):
+                        value_str_time = (
+                            evaluate_relative_date(condition.value)
+                            .astimezone()
+                            .strftime("%Y-%m-%d %H:%M:%S")
+                        )
+                    elif isinstance(condition.value, datetime):
+                        value_str_time = condition.value.astimezone().strftime("%Y-%m-%d %H:%M:%S")
```
lukapeschke, line 730: This converts the TZ to the system time, and forces the TZ on naive objects. I thought we agreed on UTC ?

server/src/weaverbird/backends/pypika_translator/translators/base.py
Sanix-Darker: I may not getting well these TZ manipulation but, adding as argument `timezone.utc` to astimezone, force me to change fixtures tests, otherwise tests will not work anymore and we should not change fixture tests !

That's why i remove that parameter !

These are tests failing !
```
tests/backends/sql_translator_integration_tests/test_sql_postgres_pypika_translator_steps.py::test_sql_translator_pipeline[/filter_dates_from_until_pypika]
FAILED tests/backends/sql_translator_integration_tests/test_sql_postgres_pypika_translator_steps.py::test_sql_translator_pipeline[/filter_relative_dates_pypika]
FAILED tests/backends/sql_translator_integration_tests/test_sql_postgres_pypika_translator_steps.py::test_sql_translator_pipeline[/ifthenelse_dates_conditions_pypika]
```

Can you be more precise ?

lukapeschke (approved):
:medal_sports:

PR #1422, opened 2022-07-19
test(pypika top with groups): add fixture for pypika's top step with groups TC-3301

raphaelvignes:
The step is already implemented, only a fixture was missing.


lukapeschke (approved):
:+1:

PR #1425, opened 2022-07-20
fix: formula step pypika [TCTC-3293]

Sanix-Darker:
## WHAT

- added a sanitizer for columns on the formula step

### BEFORE
before

### AFTER
Peek 2022-07-20 08-33

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -129,7 +129,7 @@ class SQLTranslator(ABC):
     REGEXP_OP: RegexOp
     TO_DATE_OP: ToDateOp
     # depending on the translator, this may change to ` or '
-    QUOTE_CHAR = '"'
+    QUOTE_CHAR: str
 
     def __init__(
         self: Self,
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -822,7 +822,8 @@ class SQLTranslator(ABC):
 
             return formula
 
-        query = query.select(LiteralValue(_sanitize_formula(step.formula)).as_(step.new_column))
+        step.formula = _sanitize_formula(step.formula)
+        query = query.select(LiteralValue(step.formula).as_(step.new_column))
 
         return StepContext(query, columns + [step.new_column])
 
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -806,12 +806,13 @@ class SQLTranslator(ABC):
 
         def _sanitize_formula(formula):
             # we replace [ to "
-            formula = re.sub(r'[\[\]]', self.QUOTE_CHAR, formula)
+            formula = re.sub(r'[\[\]]|["]|[`][\']', self.QUOTE_CHAR, formula)
 
             # we add quotes on columns for postgresql
             formula = re.sub(
                 r'([a-zA-Z_a-zA-Z]+)', r'{}\1{}'.format(self.QUOTE_CHAR, self.QUOTE_CHAR), formula
-            ).replace('""', self.QUOTE_CHAR)
+            )
+            formula = formula.replace(self.QUOTE_CHAR + self.QUOTE_CHAR, self.QUOTE_CHAR)
 
             if '/' in formula:
                 formula = re.sub(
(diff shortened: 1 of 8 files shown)


server/tests/backends/fixtures/formula/special_brackets_pypika.json
```diff
@@ -0,0 +1,94 @@
+{
+  "exclude": [
+    "mongo",
+    "mysql",
+    "pandas",
+    "postgres",
+    "snowflake",
+    "bigquery_pypika"
```
lukapeschke, line 8: Why is bigquery excluded here ? Which column has a space in its name ?

Sanix-Darker: No columns has spaces for the dataset used to test this fixture !
I assumed the user trying to encapsulate it with brackets and/or something else, and genuinely, we should never have a dataset with spaces inside columns (https://stackoverflow.com/questions/28926175/can-you-use-field-alias-with-space-in-google-big-query) that's why i added this exclusion.

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -128,6 +128,8 @@ class SQLTranslator(ABC):
     FROM_DATE_OP: FromDateOp
     REGEXP_OP: RegexOp
     TO_DATE_OP: ToDateOp
+    # depending on the translator, this may change to ` or '
+    QUOTE_CHAR = '"'
```
lukapeschke, line 132:
```suggestion
    QUOTE_CHAR: str
```

Let's be explicit :smile: Having the `QUOTE_CHAR` defined in all connectors, like the rest of the attributes, will allow us to see errors more easily

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -801,8 +802,32 @@ def formula(
...
+                formula = re.sub(
+                    r'(?<=%)\s*(\w+)|(?<=%)\s*(\"?.*\"?)', r' NULLIF(\1\2, 0)', formula
+                )
+
+            return formula
+
+        step.formula = _sanitize_formula(step.formula)
+        query: "QueryBuilder" = self.QUERY_CLS.from_(prev_step_name).select(
+            *columns, LiteralValue(step.formula).as_(step.new_column)
+        )
```
lukapeschke, line 830:
```suggestion
        formula = _sanitize_formula(step.formula)
        query: "QueryBuilder" = self.QUERY_CLS.from_(prev_step_name).select(
            *columns, LiteralValue(formula).as_(step.new_column)
        )
```

Don't mutate the pipeline

server/src/weaverbird/backends/pypika_translator/translators/googlebigquery.py
```diff
@@ -77,6 +79,20 @@ def split(
...
+        prev_step_name: str,
+        columns: list[str],
+        step: "FormulaStep",
+    ) -> StepContext:
+        query: "QueryBuilder" = self.QUERY_CLS.from_(prev_step_name).select(
+            *columns, LiteralValue(handle_zero_division(step.formula)).as_(step.new_column)
+        )
+
+        return StepContext(query, columns + [step.new_column])
+
```
lukapeschke, line 95: Is this still required ?

server/tests/backends/fixtures/formula/special_brackets_pypika.json
lukapeschke: Users might still want to surround quotes in brackets (maybe because they copy and adapt a query from somewhere), even if spaces are not supported. I don't feel comfortable with having an inconsistent API, supporting brackets in some places but not everywhere

lukapeschke (changes requested):

server/src/weaverbird/backends/pypika_translator/translators/base.py
Sanix-Darker: https://github.com/ToucanToco/weaverbird/pull/1425/commits/cf01b697aa9209da620bdf02edca5f1697a75cd3

server/src/weaverbird/backends/pypika_translator/translators/base.py
Sanix-Darker: https://github.com/ToucanToco/weaverbird/pull/1425/commits/ab26fe1251ba3344c4df7244c6a6655c064d7065

server/src/weaverbird/backends/pypika_translator/translators/googlebigquery.py
Sanix-Darker: https://github.com/ToucanToco/weaverbird/pull/1425/commits/36e3f7b11634bdfdfdb8540552291e9762c697ad

PR #1428, opened 2022-07-22
feat: formula builder for our translators [TCTC-3390]

Sanix-Darker:
## WHAT

- standardize formula building across our back-ends/translators
- refactor an updates by integrating those changes
- add some new tests


server/src/weaverbird/backends/mongo_translator/steps/formula.py
```diff
@@ -1,128 +1,6 @@
...
-        columns_aliases[f'__vqb_col_{i}__'] = match
-        sanitized_formula = sanitized_formula.replace(f'[{match}]', f'__vqb_col_{i}__')
-    return columns_aliases, sanitized_formula
-
-
-class InvalidFormula(Exception):
-    """
-    Raised when a formula is not supported
-    """
+    return MongoFormulaBuilder.translate_formula(step)
```
jGundermann, line 6: Maybe we should rename this to something more generic.

Or create a new specifique class for SQL

jGundermann (changes requested):
Overall seems fine for me but we should change naming to avoid confusion

server/src/weaverbird/backends/mongo_translator/steps/formula.py
Sanix-Darker: mhhh... not sure to understand well, can you provide an example of your renaming proposition please ?

lukapeschke (commented):
I'm not sure about the approach here... We're still using regexes, which is what we wanted to avoid.

I thought we'd define our own AST, implement a parser capable of transforming python's ASTs into one ouf our own (and raise errors in case of unsupported operations), and have an implementation of `our AST format -> something understood by the backend`.

Here, I feel like we're introducing extra complexity without solving the "We are in complete control of what we're processing" issue

server/src/weaverbird/backends/mongo_translator/steps/formula.py
jGundermann:
```suggestion
    return FormulaBuilder.translate_formula(step)
```

server/src/weaverbird/backends/utils/formula_builder.py
```diff
@@ -0,0 +1,566 @@
...
+            'Pow': cls.POW_OP,
+        }
+
+        try:
+            return operator_str[type(binop.op).__name__]
+        except IndexError as excp:
+            raise InvalidFormula(f'Operator {binop.op.__class__} is not supported') from excp
+
+    @classmethod
+    def evaluate(cls, formula: Any):
```
davinov, line 225: Do we need this here? Each backend will use its own engine to evaluate the formula, so I guess this formula_builder could stick to just proving the syntax tree.

server/src/weaverbird/backends/utils/formula_builder.py
```diff
@@ -0,0 +1,566 @@
...
+        return f'{self.prefix}{self.column_name}{self.suffix}'
+
+    def __add__(self, other) -> str:
+        return self.__str__() + other
+
+    def __radd__(self, other) -> str:
+        return self.__str__() + other
+
+
+class FormulaBuilder:
```
davinov, line 72: I'm not sure that the class helps much here, as there is no instance data. Maybe just exporting functions directly would be enough?

server/src/weaverbird/backends/utils/formula_builder.py
```diff
@@ -0,0 +1,566 @@
...
+                built_expr = expr
+
+            return built_expr
+        elif type(formula) in (int, float, bool):
+            return formula
+        else:
+            raise InvalidFormula
+
+
+class MongoFormulaBuilder(FormulaBuilder):
```
davinov, line 362: Should we move this to the mongo backend?
(Same for the others below)

server/src/weaverbird/backends/utils/formula_builder.py
```diff
@@ -0,0 +1,566 @@
...
+        return formula
+
+    @classmethod
+    def sanitize_formula(cls, formula: str) -> Any:
+        return {}, formula
+
+    @classmethod
+    def build_formula_tree(
+        cls, formula: str | int | float | bool, quote_char: str | None = None
+    ) -> Any:
```
davinov, line 333: Would that be clearer if that would return tree (i.e. the root the  (`Node`) and not `Any`?

davinov (commented):
Thanks for this good work!

The content is good, but we may be able to re-organize it in a simpler way if we take a step back and try to map an overview of how things gt together in each translator:
Untitled-2022-07-22-1521

For this, maybe simple functions would be enough.
If you prefer to uses classes, then we could have an `AbstractFormulaStep`, but that would be a bit different of the organization of the rest of the backends (which just uses functions).

What do you think?

Sanix-Darker:
Closing this PR in favor of : https://github.com/ToucanToco/weaverbird/pull/1445
A better implementation is ongoing there with more abstraction than this one !

PR #1437, opened 2022-07-27
feat(domain-step): enable to label domain step with query name when query is referenced behind an uid [TCTC-3100]

alice-sevin:
Retrieve the name of the query referenced behind uid in domain step label

Before:
before

After:
after

### DIFF
--- a/src/components/Step.vue
+++ b/src/components/Step.vue
@@ -130,14 +130,10 @@ export default class Step extends Vue {
     | undefined;
 
   get stepName(): string {
-    if (this.step.name === 'domain') {
-      // enable to retrieve the related name of a query referenced behind an uid
-      return humanReadableLabel(this.step, (domain: string | ReferenceToExternalQuery) =>
-        retrieveDomainName(domain, this.queries),
-      );
-    } else {
-      return humanReadableLabel(this.step);
-    }
+    // enable to retrieve the related name of a query referenced behind an uid
+    return humanReadableLabel(this.step, (domain: string | ReferenceToExternalQuery) =>
+      retrieveDomainName(domain, this.queries),
+    );
   }
 
   get canConfigurePreviewSourceSubset(): boolean {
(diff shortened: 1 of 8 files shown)


src/components/Step.vue
```diff
@@ -124,7 +130,14 @@ export default class Step extends Vue {
...
 
   get stepName(): string {
-    return humanReadableLabel(this.step);
+    if (this.step.name === 'domain') {
+      // enable to retrieve the related name of a query referenced behind an uid
+      return humanReadableLabel(this.step, (domain: string | ReferenceToExternalQuery) =>
+        retrieveDomainName(domain, this.queries),
+      );
+    } else {
+      return humanReadableLabel(this.step);
```
davinov, line 139: Maybe that function could be passed to all `humanReadableLabel`?
Even if it's not used, it shouldn't be a problem

davinov: And it makes one less if :D

tests/unit/step.spec.ts
```diff
@@ -130,6 +138,25 @@ describe('Step.vue', () => {
...
+        isLast: true,
+        step: { name: 'domain', domain: { uid: '1', type: 'ref' } },
+        indexInPipeline: 0,
+        variableDelimiters: { start: '{{ ', end: ' }}' },
+      },
+    });
+    expect(retrieveDomainNameStub).toHaveBeenCalledWith({ uid: '1', type: 'ref' }, [
+      { uid: '1', name: 'Query 1' },
+    ]);
+  });
```
davinov, line 158: :+1:

davinov (approved):

src/components/Step.vue
alice-sevin: je fais la modif :pray:

alice-sevin: https://github.com/ToucanToco/weaverbird/pull/1437/commits/2e058ec4390ca176a5376c4d0ad76f72ff9d33b7

PR #1439, opened 2022-07-28
feat(pypika Snowflake): date extract

raphaelvignes:
This step was not implemented. 
Only Snowflake pypika for now as it was requested within a bug card


jGundermann (approved):
LGTM

PR #1441, opened 2022-07-29
feat: Make backend discoverable TCTC-3456

lukapeschke:
TCTC-3456

### DIFF
--- a/server/src/weaverbird/backends/mongo_translator/registry.py
+++ b/server/src/weaverbird/backends/mongo_translator/registry.py
@@ -1,0 +1,16 @@
+from typing import Callable
+
+from weaverbird.utils.step_registry import REGISTRY, BackendStepRegistry
+
+
+def mongo_registry() -> BackendStepRegistry:
+    return REGISTRY.registry("mongo")
+
+
+def register(*args, **kwargs) -> Callable:
+    called_with_kwargs = bool(kwargs)
+    step_name = kwargs.pop('step_name', None) or args[0].__name__.replace('translate_', '')
+    if called_with_kwargs:
+        return mongo_registry().register(*args, step_name=step_name, **kwargs)
+    else:
+        return mongo_registry().register(step_name=step_name, **kwargs)(*args)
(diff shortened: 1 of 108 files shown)


davinov (commented):
Thanks for working on this, it's a very interesting for the architecture of the weaverbird's python module!

Your approach is nice and pragamatic, additive-only. The added features such as the minimum version are useful touches :+1: 
Yet I feel we kinda already have the same sort of information in the files where we import all steps in each backend that the one you put in the registry. Maybe we can think of something that would use it? 
This would be to avoid an unfortunate state when we forget either the decorator or the import in the backend main method.
But this would probably mean reconsidering also how we write backends, maybe with an abstract base for all of them? Happy to discuss alternatives if you feel like it. Otherwise, it could also be a nice first version (if so, we need to update the documentation about how to add a step)

server/src/weaverbird/backends/mongo_translator/registry.py
```diff
@@ -0,0 +1,16 @@
...
+from weaverbird.utils.step_registry import REGISTRY, BackendStepRegistry
+
+
+def mongo_registry() -> BackendStepRegistry:
+    return REGISTRY.registry("mongo")
+
+
+def register(*args, **kwargs) -> Callable:
+    called_with_kwargs = bool(kwargs)
+    step_name = kwargs.pop('step_name', None) or args[0].__name__.replace('translate_', '')
```
PrettyWood, line 12: I think it's better to add an enum and use everytime an explicit `@register(step_name=...)` instead of the magic `translate_...`.
We spend way more time reading and refactoring than writing new code

davinov:
@lukapeschke I think we should close this very old PR for 2 reasons:
- as we're moving towards hybrid pipelines, exposing which backends supports what has less interest
- as we're moving this logic to our data execution service, it'll be there that it would make sense

lukapeschke:
Yeah I agree, the most important thing here was "is feature X supported by mysql Y". we can think about that in hades

PR #1442, opened 2022-08-01
fix(pypika): snowflake date extract

raphaelvignes:
Fixed the formulas having a "%" for Snowflake's dateextract, it seemd to mess up with the interpolation.
'%' were replaced by 'MOD(arg1, arg2)


fspot (approved):

lukapeschke (approved):
Nice!

PR #1443, opened 2022-08-01
chore: return to_date result as timestamps

raphaelvignes:
* Make `to_date` return timestamps.

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/operators.py
+++ b/server/src/weaverbird/backends/pypika_translator/operators.py
@@ -20,3 +20,4 @@ class ToDateOp(Enum):
     TO_DATE = auto()
     STR_TO_DATE = auto()
     PARSE_DATE = auto()
+    TO_TIMESTAMP = auto()
--- a/server/src/weaverbird/backends/pypika_translator/operators.py
+++ b/server/src/weaverbird/backends/pypika_translator/operators.py
@@ -17,7 +17,6 @@ class FromDateOp(Enum):
 
 
 class ToDateOp(Enum):
-    TO_DATE = auto()
     STR_TO_DATE = auto()
     PARSE_DATE = auto()
     TO_TIMESTAMP = auto()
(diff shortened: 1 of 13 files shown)


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1262,11 +1263,13 @@ def todate(
                     convert_fn = StrToDate
                 case ToDateOp.PARSE_DATE:
                     convert_fn = ParseDate
+                case ToDateOp.TO_TIMESTAMP:
+                    convert_fn = functions.Timestamp
                 case _:
                     raise NotImplementedError(f"[{self.DIALECT}] todate has no set operator")
             date_selection = convert_fn(col_field, step.format)
         else:
```
lukapeschke, line 1271: We're still returning something that's not a timestamp if the format is specified. Could you please add `self._cast_to_timestamp` of the result in both cases ?

server/src/weaverbird/backends/pypika_translator/translators/postgresql.py
```diff
@@ -30,6 +30,7 @@ class PostgreSQLTranslator(SQLTranslator):
         integer="INTEGER",
         text="TEXT",
         datetime="TIMESTAMP",
+        timestamp="TIMESTAMP",
```
lukapeschke, line 33: Could you please set the `TO_DATE_OP` to `TO_TIMESTAMP` ?

lukapeschke (changes requested):
Thanks for the proposal! I've got a few remarks and suggestions, see inline comments

server/src/weaverbird/backends/pypika_translator/operators.py
```diff
@@ -20,3 +20,4 @@ class ToDateOp(Enum):
     TO_DATE = auto()
```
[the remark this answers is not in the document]
raphaelvignes: :+1:

server/src/weaverbird/backends/pypika_translator/translators/postgresql.py
raphaelvignes: good catch.

jGundermann (approved):
LGTM

Issue #1294, opened 2022-04-14
Rewrite formula implementation for python/pandas backend


davinov:
The actual implementation is based on pandas' eval, which is quite permissive (close to any valid python) but doesn't work with some modern dtypes, such as https://pandas.pydata.org/docs/user_guide/integer_na.html

My suggestion would be to use the same parser for formulas for all backends, so we get alsways the the same ast, and the same set of nodes and operations to support. Then each backend could implement the transformation of this ast to its engine.
For pandas, that would mean translating formula `a + b` to `df[a] + df[b]` instead of using `df.eval(a + b)`.

Advantages:
- same syntax for all backend formulas
- clear set of formula operator supported across backend
- eval is quite limited in pandas
- eval is not safe

PR #1445, opened 2022-08-01
Implement custom AST for formula step TCTC-3390

lukapeschke:


server/src/weaverbird/pipeline/formula_ast/types.py
```diff
@@ -0,0 +1,72 @@
...
+
+class ColumnName(BaseModel):
+    name: str
+    alias: str
+
+
+Expression = Union['Operation', ColumnName, Constant]
+
+
+# Dataclasses do not supported recursive types for now
```
davinov, line 27: Maybe it's a bit overkill to bring pydantic for this? (we don't need validation)

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
```diff
@@ -1,40 +1,21 @@
...
 
 def eval_formula(df: DataFrame, formula: str) -> DataFrame:
+    pandas_expr = formula_expression_to_pandas(FormulaParser(formula).parse())
     try:
-        result = df.eval(formula)
+        result = df.eval(pandas_expr)
     except Exception:
         # for all cases not handled by NumExpr
-        result = df.eval(formula, engine='python')
+        result = df.eval(pandas_expr, engine='python')
```
davinov, line 18: One of the interest of parsing the formula ourselves is to overcome the limitations of pandas' eval and use standard vectored operations instead.
So it would be best if we did the eval_formula was doing operations directly when reading a node:
- walking along a name node => df[that name]
- walking along a constant node => that constant value
- walking along an operation node => getattr(right_result,'__<this op>__')(left_result) (maybe there's a more idiomatic way of writing this, or just doing a match with the 6 cases would be enough)

server/src/weaverbird/pipeline/formula_ast/eval.py
```diff
@@ -0,0 +1,155 @@
...
+
+class UnsupportedExpression(ASTError):
+    """Unsupported expression"""
+
+
+class UnsupportedConstant(ASTError):
+    """Unsupported constant"""
+
+
+class FormulaParser:
```
davinov, line 41: :+1:

server/src/weaverbird/pipeline/formula_ast/eval.py
```diff
@@ -0,0 +1,155 @@
...
+
+        while token := next_token():
+            if token.type in (tokenize.ENCODING, tokenize.NEWLINE, tokenize.ENDMARKER):
+                continue
+            elif token.type == tokenize.STRING:
+                yield self.sanitize_string(token.string)
+            elif token.type == tokenize.OP and token.string == '[':
+                yield parse_col_name(token)
+            else:
+                yield token.string
```
davinov, line 88: That seems a clever way for parsing `[some stuff]` as a unique `Name` called `some stuff`, yet it's a bit difficult to understand.
The prior implementation seemed more simple to me (I'm probably surely on this). It was about replacing `[...]` with a regex and the replacing back in `Name` elements (by walking the tree).

Do you have some arguments for one or the other approach?

lukapeschke: I feel more confident in this approach, because we will deny everything that is not deemed "an identifier" by python, so it'll avoid bad suprises later on when parsing and evaluating (for example there were some tests with `a column ^ name`, which could be interpreted as a valid formula with `^` being a power operator by some backends). 
And I'm really not a big fan of regexes :sweat_smile: 

Although now that I'm thinking about it, this approach does not allow for column names containing several consecutive whitespaces or starting with whitespaces, as they'll be stripped. when rebuilding the column identifier. So the best approach would probably be to just take everything between the start of the first and the end of the last `NAME` tokens. 

I'd rather keep this approach (and I'll definitely add comments everywhere), but if the majority prefers regexes, I'll change the implem :smile:

server/src/weaverbird/pipeline/formula_ast/types.py
lukapeschke: I agree that Pydantic is overkill (especially since I had to use `smart_union`) but I haven't been able to make recursive types work (at least not without some hacks) with dataclasses :confused: 

OTOH, most of our formulas are very simple, it's not like we'll build ASTs with several hundred pydantic model instances, so I don't think the performance overhead here is important compared to execution time. And we'd need some profiling to make sure this is really slower than regexes :smile: .

Maybe it could be implemented with typed NamedTuples though, I'll have a look into that !

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
lukapeschke: I assumed pandas would be faster than us, that's why I went with it although we could maybe build a function that'd be applied only once.

But since we now control the AST, I think `numexpr` covers all our usecases: https://numexpr.readthedocs.io/projects/NumExpr3/en/latest/user_guide.html#supported-operators . Maybe we could build a numexpr-compatible formula from the AST and remove the `try/except` ? It'll probably be much faster than us (no python function call stack/dereferencing).

Again, I'm also fine with the manual eval approach, whatever the majority prefers

lukapeschke (commented):
@davinov  Thanks for the review! I hope I haven't forgotten any comment

server/src/weaverbird/pipeline/formula_ast/eval.py
davinov: The point about invalid characters and spaces are definitely valid :+1:  
Eventually, it shows that the decision is a product-led one: which column names should be supported, and which shouldn't.
Let's ask Tarek about this, and maybe the care team / product analytics, and decide afterwards! (In the meantime, keep the current approach)

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
davinov: I remember a case where we had a problem with something like `df.eval('a + b')`, and where `df['a'] + df['b']` worked.
I thunk this was because the nullable dtypes were not supported in eval. I'll try to find this bug again in the archives

lukapeschke: Ah shit forgot about connectors and their exotic dtypes, nice catch :+1:  Okay let's go for manual eval then. I'll see what I can do to manipulate `Series` as much as possible rather than `DataFrame`s

davinov: I also found this issue about it: https://github.com/ToucanToco/weaverbird/issues/1294

lukapeschke: Pandas manual eval implemented in https://github.com/ToucanToco/weaverbird/commit/b32a026fb6598f5040d1807092d536c0b6fa17dd

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
```diff
@@ -1,27 +1,37 @@
...
+    Operator.MUL: operator.mul,
+    Operator.DIV: operator.truediv,
+    Operator.MOD: operator.mod,
+}
+
 
+def _eval_operation(df: DataFrame, op: Operation) -> Series:
+    return _OP_MAP[op.operator](
+        _eval_expression(df, op.left), _eval_expression(df, op.right)
+    ).replace([np.inf, -np.inf], np.nan)
```
davinov, line 22: Super neat is this implem :ok_hand: 
:+1:

davinov (approved):

server/src/weaverbird/pipeline/formula_ast/eval.py
```diff
@@ -0,0 +1,165 @@
...
+    @staticmethod
+    def sanitize_string(s: str) -> str:
+        # Removing outer quotes
+        unquoted = ast.literal_eval(s)
+        # Escaping single quotes in the string
+        with_escaped_single_quotes = unquoted.replace('\'', r'\'')
+        # Wrapping the entire thing in single quotes
+        return f"'{with_escaped_single_quotes}'"
+
+    def _iterate_tokens(self, tokens: Iterator[tokenize.TokenInfo]) -> Generator[str, None, None]:
```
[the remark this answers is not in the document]
lukapeschke: Oops sorry, since its WIP I procrastinated the comments :sweat_smile: But you're right, I'll definitely add some!

server/src/weaverbird/pipeline/formula_ast/eval.py
```diff
@@ -61,9 +64,12 @@ def next_token():
 
         def parse_col_name(prev_token: tokenize.TokenInfo) -> str:
             chunks = []
+            # Skipping the current token, which is [
             while tok := next_token():
+                # We have an identifier, append it
                 if tok.type == tokenize.NAME:
                     chunks.append(tok.string)
+                # We reached the end of the column name. Store an alias for it and yield the alias
```
Sanix-Darker, line 72: 👍🏼

server/src/weaverbird/pipeline/formula_ast/eval.py
```diff
@@ -78,16 +84,29 @@ def parse_col_name(prev_token: tokenize.TokenInfo) -> str:
...
+                tokenize.ENDMARKER,
+            ):
+                # Those are whitespace tokens we don't care about, so not yielding them
                 continue
             elif token.type == tokenize.STRING:
+                # In case we have a string literal, we sanitize it
                 yield self.sanitize_string(token.string)
             elif token.type == tokenize.OP and token.string == '[':
+                # Square brackets are delimiters for column names, so here we read everyhting until
+                # the closing bracket
```
Sanix-Darker, line 100: 👍🏼

Sanix-Darker (approved):
Thanks for this masterpiece !
LGTM !

PR #1447, opened 2022-08-05
Snowflake timezone

jGundermann:
Snowflake has some troubles with the new way to handle timezone casting. 

Added a new Cast option for Snowflake


lukapeschke (approved):
:+1:

PR #1452, opened 2022-08-16
feat(pypika): unpivot [TCTC-3668]

raphaelvignes:
Pypika Translator Unpivot

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/athena.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/athena.py
@@ -15,6 +15,7 @@ class AthenaTranslator(SQLTranslator):
     QUERY_CLS = Query
     SUPPORT_ROW_NUMBER = True
     SUPPORT_SPLIT_PART = True
+    SUPPORT_UNPIVOT = False
     DATA_TYPE_MAPPING = DataTypeMapping(
         boolean="BOOLEAN",
         date="DATE",
@@ -25,7 +26,6 @@ class AthenaTranslator(SQLTranslator):
         timestamp="TIMESTAMP",
     )
     REGEXP_OP = RegexOp.REGEXP_LIKE
-    QUOTE_CHAR = '"'
     TO_DATE_OP = ToDateOp.TIMESTAMP
 
     @classmethod
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -22,6 +22,7 @@ from pypika import (
 from pypika.enums import Comparator, JoinType
 from pypika.queries import QueryBuilder, Selectable
 from pypika.terms import AnalyticFunction, BasicCriterion, LiteralValue, Term
+from pypika.utils import format_quotes
 
 from weaverbird.backends.pandas_executor.steps.utils.dates import evaluate_relative_date
 from weaverbird.backends.pypika_translator.dialects import SQLDialect
@@ -78,6 +79,7 @@ if TYPE_CHECKING:
         TopStep,
         TrimStep,
         UniqueGroupsStep,
+        UnpivotStep,
         UppercaseStep,
     )
     from weaverbird.pipeline.steps.aggregate import AggregateFn
@@ -97,7 +99,7 @@ class DataTypeMapping:
 
 @dataclass
 class StepContext:
-    selectable: Selectable
+    selectable: Selectable | LiteralValue
     columns: list[str]
     builder: QueryBuilder | None = None
 
@@ -128,12 +130,11 @@ class SQLTranslator(ABC):
     # supported extra functions
     SUPPORT_ROW_NUMBER: bool
     SUPPORT_SPLIT_PART: bool
+    SUPPORT_UNPIVOT: bool = False
     # which operators should be used
     FROM_DATE_OP: FromDateOp
     REGEXP_OP: RegexOp
     TO_DATE_OP: ToDateOp
-    # depending on the translator, this may change to ` or '
-    QUOTE_CHAR: str
     EVOLUTION_DATE_UNIT: dict['EVOLUTION_TYPE', DATE_INFO] = {
         'vsLastYear': 'year',
         'vsLastMonth': 'month',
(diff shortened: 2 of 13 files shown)


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1357,8 +1357,13 @@ def uniquegroups(
         )
 
     @classmethod
-    def _build_unpivot_col(cls, *, step: 'UnpivotStep') -> Term:
-        raise NotImplementedError(f"[{cls.DIALECT}] _build_unpivot_col is not implemented")
+    def _build_unpivot_col(cls, *, step: 'UnpivotStep') -> LiteralValue:
+        unpivot_token = LiteralValue(
+            f''' UNPIVOT({cls.QUOTE_CHAR}{step.value_column_name}{cls.QUOTE_CHAR} '''
```
PrettyWood, line 1362: Couldn't you use `pypika.Column` for that?

raphaelvignes: Didn't find a way to avoid this %µ¨!$ issue described below. 
When using any kind of selectable in the step, `BigQuery` & `Redshift` translators generates a `FROM` clause as such 
`FROM "__step_x_yyyyy UNPIVOT(......)"` which is not valid, whereas with literal I get rid of the quotes.

lukapeschke (commented):
I pushed a proposal to rely a bit more on PyPika and a bit less on custom formatting. 

Apart from that, I just have one remark: Shouldn't the `NotImplementedError` be raised in the `unpivot` method rather than in `_build_unpivot_col` ? I think the message is not as explicit for end users as something like `UNPIVOT is not implemented for Athena`.

Apart from that, code LGTM, nice job!

raphaelvignes:
Right, should be raise at Unpivot Level :+1:

server/src/weaverbird/backends/pypika_translator/translators/postgresql.py
```diff
@@ -69,5 +63,18 @@ def duration(
...
+
+        if issubclass(cls, RedshiftTranslator):
+            return super()._build_unpivot_col(step=step)
+        else:
+            value_col = format_quotes(step.value_column_name, cls.QUOTE_CHAR)
+            unpivot_col = format_quotes(step.unpivot_column_name, cls.QUOTE_CHAR)
+            in_cols = ', '.join(format_quotes(col, cls.QUOTE_CHAR) for col in step.unpivot)
+            in_single_quote_cols = ', '.join(format_quotes(col, "'") for col in step.unpivot)
+            return f' t1 CROSS JOIN unnest(array[{in_single_quote_cols}], array[{in_cols}]) t2 ({unpivot_col}, {value_col})'
+
```
lukapeschke, line 78: Since PostgreSQL and Athena seem to have the same implementation, Why not introduce a new `SUPPORTS_UNPIVOT` attribute in the base translator and use one implem or another depending on the value of the attribute ? And simply set the attribute to `true` in the Redshift translator ? I'm not fond of circular imports :grimacing:

server/src/weaverbird/backends/pypika_translator/translators/athena.py
```diff
@@ -37,8 +38,12 @@ def _add_date(
...
 
     @classmethod
-    def _build_unpivot_col(cls, *, step: 'UnpivotStep') -> Term:
-        raise NotImplementedError(f"[{cls.DIALECT}] _build_unpivot_col is not implemented")
+    def _build_unpivot_col(cls, *, step: 'UnpivotStep') -> str:
+        value_col = format_quotes(step.value_column_name, cls.QUOTE_CHAR)
+        unpivot_col = format_quotes(step.unpivot_column_name, cls.QUOTE_CHAR)
+        in_cols = ', '.join(format_quotes(col, cls.QUOTE_CHAR) for col in step.unpivot)
+        in_single_quote_cols = ', '.join(format_quotes(col, "'") for col in step.unpivot)
+        return f' t1 CROSS JOIN unnest(array[{in_single_quote_cols}], array[{in_cols}]) t2 ({unpivot_col}, {value_col})'
```
lukapeschke, line 46:
```suggestion
        return f' t1 CROSS JOIN UNNEST(ARRAY[{in_single_quote_cols}], ARRAY[{in_cols}]) t2 ({unpivot_col}, {value_col})'
```

server/src/weaverbird/backends/pypika_translator/translators/mysql.py
```diff
@@ -53,8 +53,15 @@ def _add_date(
...
-        raise NotImplementedError(f"[{cls.DIALECT}] _build_unpivot_col is not implemented")
+    def _build_unpivot_col(cls, *, step: 'UnpivotStep') -> str:
+        value_col = format_quotes(step.value_column_name, cls.QUOTE_CHAR)
+        unpivot_col = format_quotes(step.unpivot_column_name, cls.QUOTE_CHAR)
+        in_cols = [format_quotes(col, cls.QUOTE_CHAR) for col in step.unpivot]
+        in_single_quote_cols = [format_quotes(col, "'") for col in step.unpivot]
+        rows = ' UNION ALL SELECT '.join(
+            [f'{col}, {val}' for col, val in zip(in_cols, in_single_quote_cols)]
+        )
+        return f' t1 CROSS JOIN lateral(SELECT {rows}) as t2({value_col}, {unpivot_col})'
```
lukapeschke, line 64:
```suggestion
        return f' t1 CROSS JOIN LATERAL(SELECT {rows}) AS t2({value_col}, {unpivot_col})'
```

lukapeschke (approved):
:+1:

server/src/weaverbird/backends/pypika_translator/translators/athena.py
```diff
@@ -35,5 +38,13 @@ def _add_date(
         add_date_func = CustomFunction('DATE_ADD', ['interval', 'increment', 'datecol'])
         return add_date_func(add_date_unit, add_date_value, date_column)
 
+    @classmethod
+    def _build_unpivot_col(cls, *, step: 'UnpivotStep') -> str:
+        value_col = format_quotes(step.value_column_name, cls.QUOTE_CHAR)
+        unpivot_col = format_quotes(step.unpivot_column_name, cls.QUOTE_CHAR)
+        in_cols = ', '.join(format_quotes(col, cls.QUOTE_CHAR) for col in step.unpivot)
+        in_single_quote_cols = ', '.join(format_quotes(col, "'") for col in step.unpivot)
+        return f' t1 CROSS JOIN unnest(array[{in_single_quote_cols}], array[{in_cols}]) t2 ({unpivot_col}, {value_col})'
```
Sanix-Darker, line 47: 👌🏾

Sanix-Darker (approved):

PR #1456, opened 2022-08-30
fix: step names should be predictive for cache to work

PrettyWood:
if can have a patch version of weaverbird that'd be great


fspot (approved):
:v:

PR #1457, opened 2022-08-30
Formula step : Adding supported operators

tarekmkl:
In the interface (user) doc, we added the listing of supported operators.


lukapeschke (approved):
:+1:

PR #1463, opened 2022-08-31
feat(sql): add source_rows_subset option on translate sql

fspot:
## What

This option will be useful to preview the result of complex queries on a subset of the source dataset (e.g when a table has 10M rows and you want to preview your transformations on 10k rows).

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -148,12 +148,14 @@ class SQLTranslator(ABC):
         tables_columns: Mapping[str, Sequence[str]] | None = None,
         db_schema: str | None = None,
         known_instances: dict[int, str] | None = None,
+        source_rows_subset: int | None = None,
     ) -> None:
         self._tables_columns: Mapping[str, Sequence[str]] = tables_columns or {}
         self._db_schema_name = db_schema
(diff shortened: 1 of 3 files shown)


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -633,6 +635,8 @@ def _domain(self: Self, *, step: "DomainStep") -> StepContext:
         query: "QueryBuilder" = self.QUERY_CLS.from_(
             Table(step.domain, schema=self._db_schema)
         ).select(*selected_cols)
+        if self._source_rows_subset:
+            query = query.limit(self._source_rows_subset)
```
lukapeschke, line 639: Shoudn't we rather have this as an optional parameter of `materialize` ? That way, we'd limit the global query rather than the first step ? SQL engines should be smart enough to optimize the whole thing rather than processing all rows, right ? I'm afraid we might get "the preview returns nothing but I know I have data for the given filters" issues again.

If we get many complaints that the whole thing is too slow, we can still apply the limit in the domain step.

Also, on a join or append step, we'd get LIMIT rows per `domain` step, whereas with a limit on the whole query we'd get LIMIT rows in total

lukapeschke (approved):

lukapeschke:
OK for me after discussion with @fspot

PR #1470, opened 2022-09-22
feat(backend-service): retrieve translator from execute pipeline meta [TCTC-3370]

alice-sevin:
As now translator is retrieved from execute pipeline backendResponse in meta, we can add it to store directly from executePipeline call, no need to retrieve it from Tucana

Tucana PR to clean logic: https://github.com/ToucanToco/tucana/pull/12759
after


davinov (approved):
Simple and neat!

PR #1478, opened 2022-09-23
feat(server/pypika): Implement cumsum step TCTC-3801

lukapeschke:
TCTC-3801


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -39,6 +39,7 @@
 )
 from weaverbird.pipeline.dates import RelativeDate
 from weaverbird.pipeline.pipeline import Pipeline
+from weaverbird.pipeline.steps.cumsum import CumSumStep
```
PrettyWood, line 42: Mieux de les importer dans le `if TYPE_CHECKING`

fspot (approved):
:medal_sports:

PR #1484, opened 2022-09-27
chore: bump toucan-connectors to 3.21.1 and geopandas to 0.11.1

Sanix-Darker:
## WHAT

- bump geopandas to 0.11.1
- bump toucan-connectors to 3.21.1

## WHY

- We need to bump geopandas tobe able to bump peakina on the backend !

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -17,7 +17,7 @@ pandas = "^1.2.5"
 pydantic = "^1.9.1"
 numexpr = "^2.8.1"
 typing_extensions = "^4.2"
-geopandas = ">=0.11.1,<1"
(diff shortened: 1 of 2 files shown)


server/pyproject.toml
```diff
@@ -17,7 +17,7 @@ pandas = "^1.2.5"
 pydantic = "^1.9.1"
 numexpr = "^2.8.1"
 typing_extensions = "^4.2"
-geopandas = "^0.10.2"
+geopandas = ">=0.11.1,<1"
```
lukapeschke, line 20:
```suggestion
geopandas = "^0.11.1"
```
We should be flexible in our libs but strict in our apps

PR #1507, opened 2022-10-03
fix(server/pypika/filter): Fix "notmatches" operator for backends using REGEXP

lukapeschke:
closes TCTC-3808

### DIFF
--- a/server/tests/backends/sql_translator_unit_tests/test_filter_translators.py
+++ b/server/tests/backends/sql_translator_unit_tests/test_filter_translators.py
@@ -177,7 +177,7 @@ def test_notmatches_regexp_filter(
     ctx = regexp_translator.filter(step=step, columns=selected_columns, **default_step_kwargs)
     expected_query = (
         Query.from_(previous_step)
-        .where(Field(column).regexp(regex).negate())
+        .where(Field(column).regexp(f"%{regex}%").negate())
         .select(*selected_columns)
     )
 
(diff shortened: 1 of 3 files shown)


server/tests/backends/sql_translator_unit_tests/test_filter_translators.py
```diff
@@ -177,7 +177,7 @@ def test_notmatches_regexp_filter(
     ctx = regexp_translator.filter(step=step, columns=selected_columns, **default_step_kwargs)
     expected_query = (
         Query.from_(previous_step)
-        .where(Field(column).regexp(regex).negate())
+        .where(Field(column).regexp(f"%{regex}%").negate())
```
Sanix-Darker, line 180: Something like what has been done here : https://github.com/ToucanToco/weaverbird/pull/1508/files#diff-8837be4fbba25365c35a1d19668a79c588fac28c89e62882642b0dc93b5557fdL159-R163 sound better IMO

Sanix-Darker (approved):

server/tests/backends/sql_translator_unit_tests/test_filter_translators.py
lukapeschke: I disagree. Here, I want to test that the function is called with the regex surrounded by `%` characters. The test form you suggest would not catch an error if the `get_compliant_regex` function was bogus

PR #1508, opened 2022-10-03
fix: dict not sequence postgresql [TCTC-3809]

Sanix-Darker:
## WHAT

- fix: postgresql matches/notmatches surounding character
- test: add test

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -900,7 +900,7 @@ class SQLTranslator(ABC):
                     return column_field.notin(condition.value)
 
             case MatchCondition():
-                compliant_regex = _compliant_regex(condition.value, self.DIALECT)
+                compliant_regex = get_compliant_regex(condition.value, self.DIALECT)
 
                 if condition.operator == "matches":
                     match self.REGEXP_OP:
@@ -1567,7 +1567,7 @@ class RegexpMatching(Comparator):
     regexp = " REGEXP "
 
 
-def _compliant_regex(pattern: str, dialect: SQLDialect) -> str:
+def get_compliant_regex(pattern: str, dialect: SQLDialect) -> str:
     """
     Like LIKE, the SIMILAR TO operator succeeds only if its pattern matches the entire string;
     this is unlike common regular expression behavior wherethe pattern
@@ -1582,5 +1582,7 @@ def _compliant_regex(pattern: str, dialect: SQLDialect) -> str:
         return f"{pattern}"
     elif dialect in (SQLDialect.SNOWFLAKE, SQLDialect.MYSQL):
         return f".*{pattern}.*"
+    elif dialect == SQLDialect.POSTGRES:
+        return f"%%{pattern}%%"
 
     return f"%{pattern}%"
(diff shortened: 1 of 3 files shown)


lukapeschke (changes requested):
Could you please explain how this solves the encountered  bug ? I have trouble understanding:

* The current syntax is valid, I don't understand why you change it. I tested the following on the playground database, and it returns the expected results: `WITH __step_0_postgresqltranslator__ AS (SELECT * FROM sales) ,__step_1_postgresqltranslator__ AS (SELECT * FROM "__step_0_postgresqltranslator__" WHERE "Country" SIMILAR TO '%United%') SELECT * FROM "__step_1_postgresqltranslator__";`. This is also checked by this fixture: https://github.com/ToucanToco/weaverbird/blob/master/server/tests/backends/fixtures/ifthenelse/regex_pypika.json

* `dist is not a sequence` error originates from pandas' `read_sql` function and is a TypeError. Looks like the provided parameters are not valid.

Sanix-Darker:
Thanks for the review @lukapeschke !

How are you testing the sql query ? with raw sql or step by step like the card description ?
Because if it's using the raw sql, you're not doing it right !
The problem is that, we have to give `%%` to use it as `%` because `%` in python (a.k.a pypkia engine) is use as string formatting so when we write single `%` its assume that we are going to replace some value with this.
So when you want to place single `%` in string with query allways place double `%`

I tested that update on la-cuisine instance and it works 🤔 !

This is not the first time we got this king of issue before, that's why in a previous task, i had to add sqlalchemy.text() method here https://github.com/ToucanToco/weaverbird/blob/master/server/tests/backends/sql_translator_integration_tests/test_sql_postgres_translator_steps.py#L87 to prevent this to happen !

lukapeschke:
> Thanks for the review @lukapeschke !
> 
> How are you testing the sql query ? with raw sql or step by step like the card description ? Because if it's using the raw sql, you're not doing it right ! The problem is that, we have to give `%%` to use it as `%` because `%` in python (a.k.a pypkia engine) is use as string formatting so when we write single `%` its assume that we are going to replace some value with this. So when you want to place single `%` in string with query allways place double `%`
> 
> I tested that update on la-cuisine instance and it works thinking !
> 
> This is not the first time we got this kind of issue, that's why in a previous task, i had to add sqlalchemy.text() method here https://github.com/ToucanToco/weaverbird/blob/master/server/tests/backends/sql_translator_integration_tests/test_sql_postgres_translator_steps.py#L87 to prevent this to happen !

Aaah I see, thank you for the clarification! In that case, I believe this is not the right place to fix the bug. The generated SQL query is valid, It's the way we execute it that is not. How about fixing this method in `toucan-connectors` instead ? https://github.com/ToucanToco/toucan-connectors/blob/f39c2bd8fbdeed938995d48ad392aa40fa943cb3/toucan_connectors/common.py#L404

Sanix-Darker:
Closing this in favor of : https://github.com/ToucanToco/toucan-connectors/pull/707

Sanix-Darker:
Am reopening this because : 
- There is too much impacts if we try to deal with this behaviour on toucan-connectors repository
- I had to install sqlAlchemy since it were not present there... and i don't think we should add that just to deal with this issue

What do you think ? @lukapeschke

lukapeschke:
> Am reopening this because :
> 
>     * There is too much impacts if we try to deal with this behaviour on toucan-connectors repository
> 
>     * I had to install sqlAlchemy since it were not present there... and i don't think we should add that just to deal with this issue
> 
> 
> What do you think ? @lukapeschke

After ready the documentation of the `pandas.read_sql` function, I believe that we should either introduce SQLAlchemy into `toucan-connectors` or use something else than `pandas.read_sql`. But that's beyond the scope of this bug.

I still believe that this is not the right place to fix the issue, as the generated queries are valid. The problem lies with `pandas.read_sql`. For a quick fix I'd rather have the `pandas_read_sql` function of the connectors module escape `%` symbols.

PR #1524, opened 2022-10-10
fix(server/pandas/filter): Fix datetime object comparisons TCTC-4268

lukapeschke:


server/tests/steps/test_filter.py
```diff
@@ -278,3 +288,91 @@ def test_benchmark_filter(benchmark):
...
+
+
+@pytest.fixture
+def date_df() -> DataFrame:
+    with open(path_join(dirname(__file__), "fixtures", "sales_df.json"), "r") as fd:
+        return read_json(fd, orient="table")
+
+
+def test_date_filter(date_df: DataFrame):
+    step = FilterStep(
```
Sanix-Darker, line 300:
```suggestion
    # trying with datetime dates
    step = FilterStep(
```

server/tests/steps/test_filter.py
```diff
@@ -278,3 +288,91 @@ def test_benchmark_filter(benchmark):
...
+                DateBoundCondition(
+                    column="Transaction_date",
+                    operator="until",
+                    value=datetime(2009, 1, 3),
+                ),
+            ]
+        )
+    )
+
+    expected = read_json(
```
Sanix-Darker, line 315: maybe this could have been a fixture, but it's not a block point !

server/src/weaverbird/backends/pandas_executor/steps/utils/condition.py
```diff
@@ -16,6 +19,18 @@
...
+def _date_bound_condition_to_tz_aware_timestamp(value: str | RelativeDate | datetime) -> Timestamp:
+    if isinstance(value, RelativeDate):
+        value = evaluate_relative_date(value)
+    if isinstance(value, datetime):
+        tz = value.tzinfo or "UTC"
+        # Cannot pass a tz-aware datetime object with tz arg
+        return Timestamp(value.replace(tzinfo=None), tz=tz)
+    else:  # str
+        ts = Timestamp(value)
+        return ts if ts.tzinfo else ts.replace(tz="UTC")
```
Sanix-Darker, line 31:
```suggestion
    # str
    ts = Timestamp(value)
    return ts if ts.tzinfo else ts.replace(tz="UTC")
```

Sanix-Darker (approved):
LGTM 🔥 !
Some smalls none block comments left !

server/src/weaverbird/backends/pandas_executor/steps/utils/condition.py
lukapeschke: Linters prefer this, but our style in the various repos is to have an `else` branch even when returning in previous branches

PR #1533, opened 2022-10-18
fix: snowflake minux/underscore in aggregated column names [TCTC-4341]

Sanix-Darker:
## WHAT

- feat: added the sanitize-column-name as a class method on pypika/base translator
- feat: implement the sanitizer for snowflake translator
- test: some tests for the sanitizer
- doc: added an unrelease note for the new fix


server/src/weaverbird/backends/pypika_translator/translators/snowflake.py
```diff
@@ -34,6 +35,23 @@ class SnowflakeTranslator(SQLTranslator):
     REGEXP_OP = RegexOp.REGEXP
     TO_DATE_OP = ToDateOp.TO_TIMESTAMP_NTZ
 
+    @classmethod
+    def sanitize_column_name(cls, col: str) -> str:
```
PrettyWood, line 39: Don't we want this to be for all backends? In which case it could be a pure util function and not a method

Sanix-Darker: mhhh... am not sure we have to propagate this...
i mean at this time, the only place we found this issue was on snowflake aggregations...so... idk..

Sanix-Darker: I may be wrong but at this day...  postgresql, bigquery, athena or redshift... seems not having this issue...

fspot (approved):
lgtm ! c'est seulement sur les aggrégations que ce bug pouvait survenir ?

Sanix-Darker:
> lgtm ! c'est seulement sur les aggrégations que ce bug pouvait survenir ?

Pour le moment oui 🤔 si je ne me trompes pas @fspot !

lukapeschke (commented):
Before merging, could you please check if this would be solved by quoting the schema and the columns ? If yes, I don't think editing columns names is the right approach, as users will get some weird renaming in their columns (+ renaming would be inconsistent as we only do it for aggs)

lukapeschke (changes requested):
.

Sanix-Darker:
> Before merging, could you please check if this would be solved by quoting the schema and the columns ? 

- This update is related to **new** columns not **existing** columns !
- we were already updating those column names before by suffixing them with the corresponding aggregations names  (sum, avg, ...)
  ```
  SQ0.TEST ->SQ0.TEST-sum
  SQ0.EXAMPLE ->SQ0.EXAMPLE-avg
  ```
 
  this doesn't affect existing columns of the dataset !

lukapeschke:
>     * This update is related to **new** columns not **existing** columns !
> 
>     * we were already updating those column names before by suffixing them with the corresponding aggregations names  (sum, avg, ...)
>       ```
>       SQ0.TEST -> SQ0.TEST-sum
>       SQ0.EXAMPLE -> SQ0.EXAMPLE-avg
>       ```
>       
>       
>           
>             
>           
>       
>             
>           
>       
>           
>         
>       this doesn't affect existing columns of the dataset !
>       This fix just replace those specials caracters not supported by snowflake to -> `_`

The bug is not exclusive to the columns we add:

```python
>>> steps = [{"name": "domain", "domain": "beers_tiny"}, {"name": "rename", "toRename": [("price_per_l", "price-per-l")]}]
>>> query = translate_pipeline(sql_dialect=SQLDialect.SNOWFLAKE, pipeline=Pipeline(steps=steps), tables_columns={"beers_tiny": _BEERS_TABLE_COLUMNS})
>>> query
'WITH __step_0_snowflaketranslator__ AS (SELECT price_per_l,alcohol_degree,name,cost,beer_kind,volume_ml,brewing_date,nullable_name FROM beers_tiny) ,__step_1_snowflaketranslator__ AS (SELECT price_per_l "price-per-l",alcohol_degree,name,cost,beer_kind,volume_ml,brewing_date,nullable_name FROM __step_0_snowflaketranslator__) SELECT price-per-l,alcohol_degree,name,cost,beer_kind,volume_ml,brewing_date,nullable_name FROM __step_1_snowflaketranslator__'
>>> pd.read_sql(query, engine) # ProgrammingError
```

If I add quotes around `"price-per-l"` in the query above, I get the expected result

Sanix-Darker:
Thanks for the reply @lukapeschke !
Am understanding now, what you meant !

Do you have a proposition of a better way to handle this bug ? (even if it's out of the scope from **aggregations columns** ).

What if, in snowflake translator, i replace the `sanitize_column_name` function to : 
```python
@classmethod
def sanitize_column_name(cls, col: str) -> str:
      return re.sub(r'\w+(?:-\w+)+', '"\g<0>"', col)

# >> ss.sanitize_column_name('price-test-o_example')
# << '"price-test-o_example"'
```

then ? 🤷🏾‍♂️

WDYT ?

lukapeschke:
> Thanks for the reply @lukapeschke ! Am understanding now, what you meant !
> 
> Do you have a proposition of a better way to handle this bug ? (even if it's out of the scope from **aggregations columns** ).
> 
> What if, in snowflake translator, i replace the `sanitize_column_name` function to :
> 
> ```python
> @classmethod
> def sanitize_column_name(cls, col: str) -> str:
>       return re.sub(r'\w+(?:-\w+)+', '"\g<0>"', col)
> 
> # >> ss.sanitize_column_name('price-test-o_example')
> # << '"price-test-o_example"'
> ```
> 
> ? 🤷🏾‍♂️
> 
> WDYT ?

I think regexes have cost us a lot of trouble until now, and that it is PyPika's responsibility to quotes identifiers. My proposition is:

1. Create our custom `SnowflakeQueryBuilder` which basically just inherits this one https://github.com/kayak/pypika/blob/46c90d1638e2090cc0e8e5adb87d9b9cbd66ec01/pypika/dialects.py#L37 . But sets the `QUOTE_CHAR` to `"`.
2. Check if everything is running fine
3. After having used it for a while, if it works as expected, contribute it back to PyPika.

@fspot @PrettyWood  thoughts ?

server/src/weaverbird/backends/pypika_translator/translators/snowflake.py
```diff
@@ -15,6 +15,22 @@
...
+class SnowflakeQueryBuilder(QueryBuilder):
+    QUOTE_CHAR = '"'
+    ALIAS_QUOTE_CHAR = '"'
+    QUERY_ALIAS_QUOTE_CHAR = ""
+    QUERY_CLS = SnowflakeQuery
+
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(dialect=Dialects.SNOWFLAKE, **kwargs)
+
+
```
lukapeschke, line 33:
```suggestion
from pypika.dialects import SnowflakeQueryBuilder

class QuotedSnowflakeQueryBuilder(SnowflakeQueryBuilder):
    QUOTE_CHAR = '"'
```

We want to be as close as possible to the PyPika implem

lukapeschke (changes requested):

server/src/weaverbird/backends/pypika_translator/translators/snowflake.py
Sanix-Darker: https://github.com/ToucanToco/weaverbird/pull/1533/commits/cf2ee8780ca551454e5d1212304757a9b1996bbf

PR #1538, opened 2022-10-19
feat(textstep): allow to create columns of other types [TCTC-4365]

fspot:
**Before** : textstep can add columns of type string, only
**After** : other types are allowed. Useful when using variables like : 

```
{
  "name": "text",
  "new_column": "current_date",
  "text": "{{ TODAY }}"
},
```

:information_source: since we only deal with `datetime` variables (and not `date`) in laputa, and since mongo doesnt have distinct types, I chose to only handle `datetime` type

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Pypika: Snowflake translator now has a custom queryBuilder class to force the `QUOTE_CHAR` to `"`
 - Feat: Added `replacetext` step for all backends
-- Feat: `text` step now allows to create other types of column (when used in conjunctions with variables)
+- Feat: `text` step now allows to create other types of column (when used in conjunction with variables)
 
 ### Breaking
 * Dropped support for the old snowflake  SQL translator
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Pypika: Snowflake translator now has a custom queryBuilder class to force the `QUOTE_CHAR` to `"`
 - Feat: Added `replacetext` step for all backends
+- Feat: `text` step now allows to create other types of column (when used in conjunctions with variables)
 
 ### Breaking
 * Dropped support for the old snowflake  SQL translator
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -1392,14 +1392,32 @@ class SQLTranslator(ABC):
     ) -> StepContext:
         from pypika.terms import ValueWrapper
 
+        # Since we're using WITH...AS syntax, we add an explicit cast here to provide type
+        # context to the engine. Without that, we might encounter "failed to find conversion
+        # function from "unknown" to text" errors
+        value = step.text
+
+        if isinstance(value, datetime):
+            # ValueWrapper(value) would produce an iso8601 string which
+            # is not properly handled by some backends
+            value_wrapped = ValueWrapper(value.strftime("%Y-%m-%d %H:%M:%S"))
+        else:
+            value_wrapped = ValueWrapper(value)
+
+        if isinstance(value, datetime):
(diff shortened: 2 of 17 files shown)


server/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Pypika: Snowflake translator now has a custom queryBuilder class to force the `QUOTE_CHAR` to `"`
 - Feat: Added `replacetext` step for all backends
+- Feat: `text` step now allows to create other types of column (when used in conjunctions with variables)
```
```suggestion
- Feat: `text` step now allows to create other types of column (when used in conjunction with variables)
```

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1392,14 +1392,32 @@ def text(
...
+        # context to the engine. Without that, we might encounter "failed to find conversion
+        # function from "unknown" to text" errors
+        value = step.text
+
+        if isinstance(value, datetime):
+            # ValueWrapper(value) would produce an iso8601 string which
+            # is not properly handled by some backends
+            value_wrapped = ValueWrapper(value.strftime("%Y-%m-%d %H:%M:%S"))
+        else:
+            value_wrapped = ValueWrapper(value)
```
lukapeschke, line 1405: I think we should ensure that the object is naive. Maybe in the validator ?Something along the lines of
```python
from zoneinfo import ZoneInfo

if value.tzinfo is not None:
    value = value.astimezone(ZoneInfo('UTC')).replace(tzinfo=None)
```

server/src/weaverbird/pipeline/steps/text.py
```diff
@@ -1,13 +1,14 @@
...
 from weaverbird.pipeline.steps.utils.base import BaseStep
 from weaverbird.pipeline.steps.utils.render_variables import StepWithVariablesMixin
 from weaverbird.pipeline.types import ColumnName
 
 
-class TextStep(BaseStep):
+class TextStep(BaseStep, smart_union=True):
     name: Literal["text"] = "text"
-    text: str
+    text: datetime | int | float | bool | str
```
lukapeschke, line 11: Some SQL backends might return `date` rather than `datetime`. I'm fine with handling only `datetime` here, but maybe we could add a conversion to the validator just to be 100% sure ? 

```python
if isinstance(value, date):
    value = datetime(date.year, date.month, date.day)
```

server/tests/backends/sql_translator_integration_tests/test_sql_athena_translator_steps.py
```diff
@@ -42,11 +43,15 @@ def boto_session() -> Session:
...
-def test_athena_translator_pipeline(boto_session: Session, case_id: str, case_spec_file: str):
+def test_athena_translator_pipeline(
+    boto_session: Session, case_id: str, case_spec_file: str, available_variables: dict
+):
     pipeline_spec = get_spec_from_json_fixture(case_id, case_spec_file)
 
     steps = [{"name": "domain", "domain": "beers_tiny"}] + pipeline_spec["step"]["pipeline"]
-    pipeline = Pipeline(steps=steps)
+    pipeline = PipelineWithVariables(steps=steps).render(
+        available_variables, nosql_apply_parameters_to_query
```
lukapeschke, line 53: I thought `nosql_apply_parameters_to_query` should never be used with SQL engines ?

lukapeschke (commented):
Code LGTM overall, thanks! Just a few inline remarks/comments.

Also, could you please update the SQL translator unit tests ?

server/src/weaverbird/pipeline/steps/text.py
fspot: the value stored in `text` comes from the interpolated jinja, it's not something "returned by the sql backend" :thinking:

lukapeschke: Woops yes true, I probably have `dtype` trauma :sweat_smile:

lukapeschke (approved):
perfect, thanks! :ok_hand:

PR #1540, opened 2022-10-20
chore(front): added dependabot for weaverbird frontend

Sanix-Darker:
## WHAT

Just added dependabot for weaverbird front

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,16 +7,11 @@ updates:
       interval: "daily"
   # yarn
   - package-ecosystem: "npm"
-    directory: "/src"
+    directory: "/"
     schedule:
       interval: "daily"
     allow:
-      - dependency-type: "all"
-    assignees:
-      - "davinov"
-      - "alice-sevin"
-      - "bloodstorms"
+      - dependency-type: "production"
     commit-message:
       prefix: "yarn prod"
-      prefix-development: "yarn dev"
       include: "scope"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,3 +5,18 @@ updates:
     directory: "/server"
     schedule:
       interval: "daily"
+  # yarn
+  - package-ecosystem: "npm"
+    directory: "/src"
```
davinov, line 10: I think this should be where the `package.json` and `yarn.lock` are stored, so it should be `directory: "/"`

.github/dependabot.yml
```diff
@@ -5,3 +5,18 @@ updates:
...
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-type: "all"
```
davinov, line 14: Fine by me, but honestly only the production really matters here

.github/dependabot.yml
```diff
@@ -5,3 +5,18 @@ updates:
...
+    assignees:
+      - "davinov"
+      - "alice-sevin"
+      - "bloodstorms"
```
davinov, line 18: This will be a pain to maintain if the squad changes, maybe we just keep an eye on it without mentioning anyone in particular?

davinov: Also maybe reviewers is better than assignees?

Sanix-Darker: alright, let's remove that !

PR #1541, opened 2022-10-21
refactor(front/PyPika): Make Snowflake translator inherit PypikaTranslator

lukapeschke:
Since the legacy Snowflake translator does not exist anymore, make the class inherit `PypikaTranslator`


PrettyWood (approved):
Yes top!

PR #1549, opened 2022-10-24
chore: prevent CustomSqlStepWithVariables

fspot:
This PR doesnt change existing behaviour since CustomSqlStepWithVariables was not part of the union PipelineStepWithVariables ; but we want to make sure this is done on purpose.


lukapeschke (approved):
LGTM :+1:

PR #1556, opened 2022-10-26
feat(front): Implement front-end for replacetext step TCTC-4368

lukapeschke:
TCTC-4368

### DIFF
--- a/server/src/weaverbird/pipeline/steps/replacetext.py
+++ b/server/src/weaverbird/pipeline/steps/replacetext.py
@@ -2,12 +2,15 @@ from typing import Literal
 
 from pydantic import Field
 
-from weaverbird.pipeline.steps.utils.base import BaseStep
+from weaverbird.pipeline.steps.utils.base import BaseStep, to_camelcase
 from weaverbird.pipeline.steps.utils.render_variables import StepWithVariablesMixin
 from weaverbird.pipeline.types import ColumnName, TemplatedVariable
 
 
 class ReplaceTextStep(BaseStep):
+    class Config:
+        alias_generator = to_camelcase
+
     name: Literal["replacetext"] = "replacetext"
     search_column: ColumnName
     old_str: str
--- a/src/components/stepforms/ReplaceTextStepForm.vue
+++ b/src/components/stepforms/ReplaceTextStepForm.vue
@@ -1,0 +1,82 @@
+<template>
+  <div>
+    <StepFormHeader
+      :title="title"
+      :stepName="editedStep.name"
+      :version="version"
+      :backendError="backendError"
+    />
+    <ColumnPicker
+      class="column"
+      v-model="editedStep.searchColumn"
+      name="Extract a substring from..."
+      placeholder="Enter a column"
+      data-path=".column"
+      :errors="errors"
+    />
+    <InputTextWidget
+      v-model="editedStep.oldStr"
+      name="Old string"
+      placeholder="Text to replace"
+      :errors="errors"
+    />
+    <InputTextWidget
+      v-model="editedStep.newStr"
+      name="New string"
+      placeholder="Replacement text"
+      :errors="errors"
+    />
+    <StepFormButtonbar />
+  </div>
+</template>
+
+<script lang="ts">
+import Component from 'vue-class-component';
+import { Prop } from 'vue-property-decorator';
+
+import ColumnPicker from '@/components/stepforms/ColumnPicker.vue';
+import InputTextWidget from '@/components/stepforms/widgets/InputText.vue';
+import { ColumnTypeMapping } from '@/lib/dataset';
+import { PipelineStepName, ReplaceTextStep } from '@/lib/steps';
+import { VariableDelimiters, VariablesBucket } from '@/lib/variables';
+import { VQBModule } from '@/store';
+
+import BaseStepForm from './StepForm.vue';
+
+@Component({
+  name: 'replacetext-step-form',
+  components: {
+    ColumnPicker,
+    InputTextWidget,
+  },
+})
+export default class ReplaceTextStepForm extends BaseStepForm<ReplaceTextStep> {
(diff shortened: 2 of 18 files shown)


server/src/weaverbird/pipeline/steps/replacetext.py
```diff
@@ -2,12 +2,15 @@
...
 
-from weaverbird.pipeline.steps.utils.base import BaseStep
+from weaverbird.pipeline.steps.utils.base import BaseStep, to_camelcase
 from weaverbird.pipeline.steps.utils.render_variables import StepWithVariablesMixin
 from weaverbird.pipeline.types import ColumnName, TemplatedVariable
 
 
 class ReplaceTextStep(BaseStep):
+    class Config:
+        alias_generator = to_camelcase
```
PrettyWood, line 12: isn't it something we should have by default on all base steps ?

PrettyWood (commented):
Otherwise looks good!

src/components/stepforms/schemas/replaceText.ts
```diff
@@ -0,0 +1,31 @@
...
+      },
+    },
+    oldStr: {
+      type: 'string',
+      description: 'String to replace',
+      minLength: 1,
+    },
+    newStr: {
+      type: 'string',
+      description: 'Value to use as a placeholder',
```
Sanix-Darker, line 26:
```suggestion
      description: 'Value to use as a placeholder',
      minLength: 1,
```

src/components/stepforms/schemas/replaceText.ts
```diff
@@ -0,0 +1,31 @@
...
+    },
+  },
+  required: ['name', 'searchColumn', 'oldStr'],
```
Sanix-Darker, line 29: Should newStr not a required field ?

src/components/stepforms/ReplaceTextStepForm.vue
```diff
@@ -0,0 +1,82 @@
...
+      v-model="editedStep.searchColumn"
+      name="Extract a substring from..."
+      placeholder="Enter a column"
+      data-path=".column"
+      :errors="errors"
+    />
+    <InputTextWidget
+      v-model="editedStep.oldStr"
+      name="Old string"
+      placeholder="Text to replace"
```
Sanix-Darker, line 20:
```suggestion
      placeholder="String to replace"
```

src/components/stepforms/schemas/replaceText.ts
lukapeschke: This would prevent us to replace string with an empty string

src/components/stepforms/schemas/replaceText.ts
lukapeschke: No, why ?

src/components/stepforms/ReplaceTextStepForm.vue
lukapeschke: I figured "text" would be easier to understand by non-technical users

lukapeschke: Also, it's the name of the step :slightly_smiling_face:

src/components/stepforms/schemas/replaceText.ts
```diff
@@ -0,0 +1,31 @@
...
+      title: 'Search',
+      description: 'Column in which to replace text',
+      attrs: {
+        placeholder: 'Enter a column',
+      },
+    },
+    oldStr: {
+      type: 'string',
+      description: 'String to replace',
+      minLength: 1,
```
PrettyWood, line 22: Why min length? If I want to replace empty strings with something else

lukapeschke: Then you'd use the normal `replace` step, which replaces values

server/src/weaverbird/pipeline/steps/replacetext.py
lukapeschke: I agree, but It'll probably be a breaking change. I'm preparing another PR for that, to be merged after the LTS branches have been created

lukapeschke: #1559

src/components/stepforms/schemas/replaceText.ts
PrettyWood: I don't know the codebase as well as you but from a user perspective it makes no sense to have two different steps if I want to replace '' or 'a'

lukapeschke: At first, we were wondering if replacing text and values should both be done in the same step or not. After discussion with product, we've decided to to it in two separate steps.

The difference between the two steps is:

* I want to replace all values in column `col` containing the literal value `"a"` with `42` -> that's a `replace`, the type can change. Only cells with `"a"` in them will be modified -> Cells with "baba" are left untouched.
* In my string columns, I want to replace the string "guys" with "people" -> That's a `replacetext`. All cells containing "guys" will be modified: `"guys"` will become `"people"`, `"hello guys"` will become `"hello people"`. 

In your example the expeected behaviour of a `replacetext("", "a")` would be `"foo"` -> `"afaoaoa"`, which is probably not what the end user wants

alice-sevin (approved):

src/components/stepforms/schemas/replaceText.ts
PrettyWood: Ok got it thanks a lot

bloodstorms: Too bad it wasn't called "replace substring"
I didn't realize what this would do until I read your last example 😅

lukapeschke: Yes, naming is a bit frustrating for devs, but with product we thought that it was better to have "text" everywhere rather than let end users figure out by themselves wtf a "substring" is supposed to be :smile:

PR #1559, opened 2022-10-27
feat: Allow all steps to be populated with camelCase fields

lukapeschke:


lukapeschke:
depends on #1556

PrettyWood (commented):
Why not just add alias_generator in the Config of BaseStep and that's it

lukapeschke:
> Why not just add alias_generator in the Config of BaseStep and that's it

Because there's a bit more that needs to be solved, such as `IfThenElseStep`, which contains an `IfThenElse` model that had its own config. There's a lot of noise from #1556 here though, it should be more readable once #1556 is merged :sweat_smile:

server/src/weaverbird/pipeline/steps/evolution.py
```diff
@@ -12,13 +12,13 @@
...
+    value_col: str
+    evolution_type: EVOLUTION_TYPE
+    evolution_format: EVOLUTION_FORMAT
+    index_columns: list[str] = Field(default_factory=list)
+    new_column: str | None
 
 
 class EvolutionStepWithVariable(EvolutionStep, StepWithVariablesMixin):
-    index_columns: TemplatedVariable | list[TemplatedVariable] = Field([], alias="indexColumns")
+    index_columns: TemplatedVariable | list[TemplatedVariable]
```
PrettyWood, line 24:
```suggestion
    index_columns: TemplatedVariable | list[TemplatedVariable] = Field(default_factory=list)
```

PrettyWood (changes requested):
LGTM otherwise thanks!

PrettyWood (approved):

PR #1566, opened 2022-10-28
fix(playground): _SNOWFLAKE_CONNECTION needs to be defined

lukapeschke:

### DIFF
--- a/server/playground.py
+++ b/server/playground.py
@@ -348,7 +348,7 @@ async def handle_mongo_translated_backend_request():
 
 
 ### Snowflake back-end routes
-
+_SNOWFLAKE_CONNECTION = None
 if os.getenv("SNOWFLAKE_ACCOUNT"):
     try:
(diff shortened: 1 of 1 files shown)


server/playground.py
```diff
@@ -359,7 +359,7 @@ async def handle_mongo_translated_backend_request():
             client_session_keep_alive=True,
         )
     except Exception:
-        _SNOWFLAKE_CONNECTION = None
+        pass
```
davinov, line 362: Maybe we can just log that the connection failed?

lukapeschke: Sure! Is a simple `print` okay for you ?

davinov: just to make sure it's not silent in the console

lukapeschke: Added a print and checked, it's visible in the console :slightly_smiling_face:

davinov (approved):

PR #1574, opened 2022-11-03
feat: from single quote to escaped quote [TCTC-3948]

Sanix-Darker:
## WHAT

- feat: added a custom ValueWrapper in pypika base translator
- test: a small test that illustrate the escape

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog (weaverbird python package)
 
+## Unreleased
+
+- PyPika: SQLTranslator has now an attribute for a custom `ValueWrapper`
+  defined as `VALUE_WRAPPER_CLS`.
+- PyPika: Google Big Query translator now use a custom `VALUE_WRAPPER_CLS` for
+  escaping single quote with a slash `\` rather than duplicate it.
+
 ## [0.26.1] - 2022-10-27
 
 - in order to prevent SQL injection, it is no longer possible to create a CustomSQL step with variables.
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - PyPika: Google Big Query translator now escapes single quotes in string literals with
-a backslash (\) rather than two double quotes ('').
+a backslash (`\'`) rather than two double quotes (`''`).
 
 ## [0.26.1] - 2022-10-27
 
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -2,10 +2,8 @@
 
 ## Unreleased
 
-- PyPika: SQLTranslator has now an attribute for a custom `ValueWrapper`
-  defined as `VALUE_WRAPPER_CLS`.
-- PyPika: Google Big Query translator now use a custom `VALUE_WRAPPER_CLS` for
-  escaping single quote with a slash `\` rather than duplicate it.
+- PyPika: Google Big Query translator now escapes single quotes in string literals with
+a backslash (\) rather than two double quotes ('').
 
 ## [0.26.1] - 2022-10-27
 
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -1,11 +1,9 @@
-import uuid
 from abc import ABC
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
-from datetime import date, datetime, timezone
-from enum import Enum
+from datetime import datetime, timezone
 from functools import cache
-from typing import TYPE_CHECKING, Any, Literal, Optional, Type, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Literal, TypeVar, cast
 
 from dateutil import parser as dateutil_parser
 from pypika import (
@@ -25,17 +23,12 @@ from pypika.enums import Comparator, Dialects, JoinType
 from pypika.functions import Extract, ToDate
 from pypika.queries import QueryBuilder, Selectable
 from pypika.terms import (
-    JSON,
     AnalyticFunction,
-    Array,
     BasicCriterion,
     Function,
     Interval,
     LiteralValue,
-    Node,
-    NullValue,
     Term,
-    Tuple,
     ValueWrapper,
 )
 from pypika.utils import format_quotes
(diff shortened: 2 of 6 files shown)


Sanix-Darker:
> Je comprends pas pourquoi on fait tout dans base.py alors que c'est juste pour GBQ qu'on veut un truc custom

parceque ce n'est pas si mal d'avoir un echapement de quotes uniforme aux pypika backends 🤔.
Mais c'est vrai que le contextualiser sur GBQ c'est aussi une bonne approche. vos avis @lukapeschke @fspot ?

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -139,6 +154,77 @@ def __init__(self, date_part, interval, term, alias=None):
...
+            return val
+        if val is None:
+            return NullValue()
+        if isinstance(val, (str, int, bool)):
+            wrapper_cls = wrapper_cls or ValueWrapperCustom
+            return wrapper_cls(val)
+
+        return JSON(val)
+
+
```
lukapeschke, line 227:
```suggestion
class GoogleBigQueryValueWrapper(ValueWrapper):
    @classmethod
    def get_formatted_value(cls, value: Any, **kwargs):
        if isinstance(value, str):
            value = value.replace("'", r"\'")
            return f"'{value}'"
        return super().get_formatted_value(value, **kwargs)
```

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1390,19 +1476,17 @@ def text(
...
         # context to the engine. Without that, we might encounter "failed to find conversion
         # function from "unknown" to text" errors
         value = step.text
 
         if isinstance(value, datetime):
-            # ValueWrapper(value) would produce an iso8601 string which
+            # ValueWrapperCustom(value) would produce an iso8601 string which
             # is not properly handled by some backends
-            value_wrapped = ValueWrapper(value.strftime("%Y-%m-%d %H:%M:%S"))
+            value_wrapped = ValueWrapperCustom(value.strftime("%Y-%m-%d %H:%M:%S"))
```
lukapeschke, line 1487: This behaviour is only for GBQ. Changing the behaviour here breaks all other translators (see the added fixture)

server/tests/backends/sql_translator_unit_tests/test_base_translator.py
```diff
@@ -604,6 +607,24 @@ def test_text(base_translator: BaseTranslator, default_step_kwargs: dict[str, An
...
+    ctx = base_translator.text(step=step, columns=selected_columns, **default_step_kwargs)
+
+    expected_query = Query.from_(previous_step).select(
+        *selected_columns, functions.Cast(ValueWrapper(text), "TEXT").as_(new_column_name)
+    )
+
+    # ' are replaced with \' and not '' anymore
+    assert (
+        expected_query.get_sql()
+        == 'SELECT "name","pseudonyme",CAST(\'Hello World \\\' there \\\' i\\\' am\' AS TEXT) "name" FROM "previous_with"'
```
lukapeschke, line 623:
```suggestion
        == r'''SELECT "name","pseudonyme",CAST('Hello World \' there \' i\' am' AS TEXT) "name" FROM "previous_with"'''
```

However this should only be the case with GBQ

lukapeschke (changes requested):
Thanks for the PR! However I feel like you implementation is very complex, see the inline suggestion. Also, this behaviour should be for GBQ only, other backends use regular SQL escaping (i.e. double single quotes)

server/tests/backends/sql_translator_unit_tests/test_base_translator.py
Sanix-Darker: removed !
will relay on the integration test, thanks 🙏🏽

lukapeschke (changes requested):
Thanks for the update! One last remark  and we should be good!

lukapeschke:
Ah and could you please add an entry to the changelog ?

server/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog (weaverbird python package)
 
+## Unreleased
+
+- PyPika: SQLTranslator has now an attribute for a custom `ValueWrapper`
+  defined as `VALUE_WRAPPER_CLS`.
```
lukapeschke, line 6:
```suggestion
```
Not relevant for end users

server/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
...
+- PyPika: Google Big Query translator now use a custom `VALUE_WRAPPER_CLS` for
+  escaping single quote with a slash `\` rather than duplicate it.
```
lukapeschke, line 8: End users do not care about the implem... How about something like "Google Big Query translator now escapes single quotes in string literals with a backslash (`\`) rather than two double quotes (`''`)" ?

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -142,6 +150,7 @@ def __init__(self, date_part, interval, term, alias=None):
 class SQLTranslator(ABC):
     DIALECT: SQLDialect
     QUERY_CLS: Query
+    VALUE_WRAPPER_CLS: ValueWrapper
```
lukapeschke, line 153:
```suggestion
    VALUE_WRAPPER_CLS = ValueWrapper
```

I believe it won't be defined otherwise

lukapeschke (changes requested):

server/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog (weaverbird python package)
 
+## Unreleased
+
+- PyPika: Google Big Query translator now escapes single quotes in string literals with
+a backslash (\) rather than two double quotes ('').
```
lukapeschke, line 6:
```suggestion
a backslash (`\`) rather than two double quotes (`''`).
```

server/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog (weaverbird python package)
 
+## Unreleased
+
+- PyPika: Google Big Query translator now escapes single quotes in string literals with
+a backslash (`\`) rather than two double quotes (`''`).
```
PrettyWood, line 6:
```suggestion
a backslash (`\'`) rather than two double quotes (`''`).
```

PrettyWood (approved):

PR #1577, opened 2022-11-04
chore(server): 0.27.0

Sanix-Darker:
## WHAT

bump from 0.26.1 to 0.26.2

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


lukapeschke (changes requested):
This should not be done on master, but on the `backend/release/0.26.2` branch. The patch needs to be backported to the branch first.

lukapeschke:
OR we consider that we don't want to backport this, in which case a 0.27.0 should be bumped.

lukapeschke (approved):

PR #1578, opened 2022-11-04
chore(front): upgrade node to LTS (v18)

davinov:
The CI/CD is blocked by the version of node it uses (too old for dependencies that we just upgraded).
https://github.com/ToucanToco/weaverbird/actions/runs/3384849837/jobs/5622306593

### DIFF
--- a/.github/workflows/ci-ui.yml
+++ b/.github/workflows/ci-ui.yml
@@ -33,7 +33,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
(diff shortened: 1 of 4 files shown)


PrettyWood (approved):

Sanix-Darker (commented):
In the readme line 41, we need to change also the version IMO !

davinov:
Thx @Sanix-Darker: e58b960 so it's more future proof :p

PR #1580, opened 2022-11-04
ci(front): check that build types complete

davinov:
The new update of ajv seems to block the build of types. However, this was not detected by the CI! :scream: 
Let's fix this


PrettyWood (approved):

davinov:
@PrettyWood not so fast :p it will fail !
Probably becaue typescipt verision needs to be updated as weel. I'll do that in this PR

davinov:
Closing in favor of #1581

Issue #912, opened 2021-05-28
Reorganize backends versus UI


davinov:
Since the introduction of the `server/` module, weaverbird is becoming more than just a front-end project to translate pipelines to Mongo queries.

We now have two different ways of implementing **backends**:
- translators: like the mongo one, which convert pipelines into corresponding queries
- executors: like the one made with pandas, which produces transformation functions 

and two different ways to implement them:
- using TypeScript: to be executed in a browser or with node
- using Python: to be executed server-side.

They are similar in how they must handle steps, except the executors don't need an external DB to be run against.

The front-end part of weaverbird consist only of **UI** components to visualize and edit pipelines. It could be completely separated from the translators or executors in a different folder/module. The two only needs to share a list of supported steps for the translators and executors. This could be done using a simple JSON, (optionally generated from the translators/executors files), instead of typescript files without any implementation like https://github.com/ToucanToco/weaverbird/blob/master/src/lib/translators/pandas.ts 
A third part, shared between the UI module and the translators/executors, would be the definition of the pipeline steps.

The proposed architecture would look like:
```
lib/ # definitions of the pipeline steps
ui/ # front-end components
backends/ # or engines? where to implement translators and executors
```

This leads to a first action plan:
- [x] document the difference between translators and executors
- [ ] move the UI part in the ui/ folder
- [x] move the steps types to the lib/ folder
- [x] move the mongo translator to a backends/mongo subfolder
- [x] move the pandas executor to a backends/pandas subfolder
- [ ] rewrite the supported/unsupported steps logic in the UI part to remove the dependency to any translators
- [ ] document this architecture breakdown for contributors

Corollary: I'd like some suggestions for a common word to regroup translators and executors. I've thought of "backends" or "engines" but I'm not entirely convinced they are the right word.

Notes:
- the lib part should contain steps definitions in every language we'd like to allow backends implementation. In our case, it would be a collection of 
- another issue/RFC will follow regarding the internal organization of the backends/ folder

austil:
100% on board with this "structural" refactoring !

> I'd like some suggestions for a common word to regroup translators and executors. I've thought of "backends" or "engines" but I'm not entirely convinced they are the right word.

Some food for though using the programming language taxonomy : 

- translators "compiles" pipelines to another query language
- while executors "interpret" or "run" pipelines on the fly

and those are two different type of ... implementations ?May be "executors" should be

davinov:
Done with #1581!

PR #1581, opened 2022-11-04
refactor(ui): move to /ui folder, switch to vite for build

davinov:
Part of #912 was to move everything UI-related into its own folder, so the root stays cleans.
I took this opportunity to also switch the build system to vite (it became necessary because typescript needed to be updated too, and the rollup plugins were not following).

Sorry for this faaaaaaat PR, tat contained a lot of sed/replacements in order to clean all the build errors.

TODO:
- finish cleaning error of types build
- ~~repair storybook~~
- ~~switch tests to vitest~~
- deduplicate playground code
- ~~adapt github actions and sonar~~


davinov:
OK, I think it's good enough for now. I prefer merging this ASAP to unblock the ability to do releases.
Need to keep in mind that:
- A lot of typechecking errors are still there, there were not previously visible because the CI doesn't check for them. the migration to vue 2.7 also improved the typechecking.
- A few tests were skipped (popover, filter editor). Their components works OK, I just hadn't had the time to figure their problem (which doesn't seem trivial).
- Only some stories were migrated, there is still many to do!

.gitignore
```diff
@@ -1,25 +1,8 @@
-node_modules
-/dist
-.storybook/dist
-playground/*.map
-
+# Editors
 .idea
```
jGundermann, line 2: This should maybe stay here

```suggestion
.idea
.vscode
```

jGundermann (commented):
Thx for the great PR! Huge work congrats. 
Got some type error during build. and found a minor change to make.

ui/src/components/stepforms/widgets/Autocomplete.vue
```diff
@@ -60,7 +60,7 @@ import VariableInput from './VariableInput.vue';
     FAIcon,
   },
 })
-export default class AutocompleteWidget extends Mixins(FormWidget) {
+export default class AutocompleteWidget extends FormWidget {
```
davinov, line 63: Ah oui t'as carrément enlever l'usage de Mixins! Bonne idée ^^

bloodstorms (approved):
Tout OK pour moi, testé en local super fast 🚀, tout fonctionne comme avant, pas d'erreur en console 👌 
Let's meeeeeeerge

PR #1589, opened 2022-11-08
feat(server/playground): Add support for "geometry" dtype for the pandas executor

lukapeschke:


davinov (approved):
:+1:

PR #1590, opened 2022-11-08
feat(data-viewer): use map icon when column type is geometry [TCTC-4376]

bloodstorms:
Use map icon when column type is geometry

Before
Screenshot 2022-11-08 at 12 07 44

After
Screenshot 2022-11-08 at 12 08 26


davinov (approved):
:+1:

PR #1607, opened 2022-11-16
Adapt playground to recent UI build changes

lukapeschke:
Work items:
* fix(playground/frontend): Fix links for dist files
* fix(playground/backend): Enforce the right mimetype for .cjs files

### DIFF
--- a/playground/dist/index.html
+++ b/playground/dist/index.html
@@ -255,7 +255,7 @@
     </div>
     <script src="https://cdn.jsdelivr.net/npm/vue/dist/vue.js"></script>
     <script src="https://unpkg.com/vuex@3.1.1/dist/vuex.js"></script>
-    <script type="module" src="/weaverbird.umd.js"></script>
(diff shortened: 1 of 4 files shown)


playground/dist/index.html
```diff
@@ -255,7 +255,7 @@
     </div>
     <script src="https://cdn.jsdelivr.net/npm/vue/dist/vue.js"></script>
     <script src="https://unpkg.com/vuex@3.1.1/dist/vuex.js"></script>
-    <script type="module" src="/weaverbird.umd.js"></script>
+    <script type="module" src="/weaverbird.umd.cjs"></script>
```
davinov, line 258: Awwww nice catch! I totally forgot about that ><

davinov (approved):

davinov (commented):
We could also generate a dediacted format for browser ([iife](https://vitejs.dev/config/build-options.html#build-lib) ), but if the UMD works OK, it's enough

PR #1608, opened 2022-11-16
fix(frontend): Try to also access camel case attributes in the step labeller TCTC-4619

lukapeschke:
Depends on #1607

### DIFF
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog (weaverbird npm package)
 
+### Changed
+
+- The playground has been adapted to changes introduced in v0.95.0
+- Since steps can now be represented with all their fields in camelCase, try to access both snake_case and camelCase fields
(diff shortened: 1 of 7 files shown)


ui/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog (weaverbird npm package)
 
+### Changed
+
+- The playground has been adapted to changes introduced in v0.95.0
+- Since steps can now be represented with all their fields in camelCase, try to access both snake_case and camelCase fields
```
davinov, line 6: Maybe the v3 is good time to make a breaking change and fix a convention for all steps? (not for this PR of course :p )

lukapeschke: Yup definitely agree! But it'll be a big PR to move everything to camelCase so we'll probably need some planning for that

davinov (approved):

PR #1617, opened 2022-11-23
refactor: Adapt playground response format to new pagination

lukapeschke:


server/pyproject.toml
```diff
@@ -50,9 +47,7 @@ sqlalchemy = "^1.4.44"
 pymysql = "^1.0.2"
 psycopg2 = "^2.9.5"
 pymongo = ">=4.2.0"
-redshift-connector = "^2.0.909"
 sqlalchemy-redshift = ">=0.8.11,<1"
-google-cloud-bigquery = {version = "^3.3.0", extras = ["pandas", "bqstorage"]}
```
Sanix-Darker, line 55: oohhh, good catch, we were not using those libs in the codebase ?

server/pyproject.toml
```diff
@@ -25,21 +25,18 @@ quart = {version = ">=0.17,<0.19", optional = true}
...
 pytest-xdist = ">=2.5,<4.0"
 
 [tool.poetry.dev-dependencies]
 pytest-cov = "^4.0.0"
 pytest-mock = "^3.10.0"
 # required so we can use nosql_apply_parameters_to_query during tests
-toucan-connectors = { version = "^3.17.4", extras = ["awsathena"] }
+toucan-connectors = { version = "^4.0.0", extras = ["awsathena", "google_big_query", "mongo", "Redshift", "snowflake"] }
 pytest-benchmark = "^4.0.0"
 snowflake-sqlalchemy = "^1.4.4"
```
Sanix-Darker, line 38: We probably want to get rid of this one too ?

server/pyproject.toml
lukapeschke: Well we're using them in  the tests, but they're pulled in by toucan-connectors anyway so :shrug:

server/pyproject.toml
lukapeschke: Nope, it's used in the tests

PR #1625, opened 2022-11-28
fix(playground): Add pagination_info for pandas

lukapeschke:
Seems like this disappeared during a previous rebase..


bloodstorms (approved):
👌

PR #1626, opened 2022-11-29
deps: dependabot from daily to weekly checks

Sanix-Darker:
## WHAT

Slow down dependabot checks from day to day to week !


PrettyWood (approved):
perfect thanks champ !

PR #1632, opened 2022-12-06
fix(server): date as null default on switch to prevent PlanExecutor error [TCTC-4622]

Sanix-Darker:


lukapeschke:
Could you add a new fixture ? Fixed must not remove existing tests when they make sense

lukapeschke (approved):

PR #1633, opened 2022-12-06
fix(mongo-front): added a null default value on switch for mongo date transformations [TCTC-4622]

Sanix-Darker:


Sanix-Darker:
We can ignore the sonarCube error !
So i think this PR is ready to be reviewed !

PR #1653, opened 2022-12-12
fix(ui): authorize having no aggregation in rollup step form TCTC-1897

davinov:
Until now, the aggregation array in the rollup step form couldn't be rest to no elements.
This is now fixed by:
- starting the rollup step with no aggregation: image
- allowing to delete the first element of the list (trash icon): 
image

This will be backported on v0.92.x (longer support version)


alice-sevin (approved):
LGTM :pray:

PR #1655, opened 2022-12-12
fix(new-date-input): popover position [TCTC-4324]

bloodstorms:
The Popover was overflowing the window, so we couldn't click on the bottom buttons

Before
https://user-images.githubusercontent.com/18734475/207180976-41706956-b73e-41d9-a576-898c13accadd.mp4

After
https://user-images.githubusercontent.com/18734475/207180990-b5657c32-b5f4-409d-b423-848d13fe1045.mp4

ℹ️ I add a prop `shouldCalculateHeight` prop set to false by default to prevent to break other components that are using Popover.

### DIFF
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## [0.98.2] - 2022-12-13
+
+### Fixed
+
+- Avoid date picker popover to overflow the window
+- Allow to remove any aggregation in rollup step form
+
 ## [0.98.1] - 2022-12-07
 
 ### Changed
@@ -1409,7 +1416,7 @@ TabbedRangeCalendars: keep selected tab unchanged when updated enabled calendars
 
 - Initial version, showtime!
 
-
+[0.98.2]: https://github.com/ToucanToco/weaverbird/compare/v0.98.2...v0.98.1
 [0.98.1]: https://github.com/ToucanToco/weaverbird/compare/v0.98.1...v0.98.0
 [0.98.0]: https://github.com/ToucanToco/weaverbird/compare/v0.98.0...v0.97.0
 [0.97.0]: https://github.com/ToucanToco/weaverbird/compare/v0.97.0...v0.96.0
--- a/ui/src/components/Popover.vue
+++ b/ui/src/components/Popover.vue
@@ -1,7 +1,10 @@
 <template>
   <div
     class="weaverbird-popover"
-    :class="{ 'weaverbird-popover--always-opened': alwaysOpened }"
+    :class="{
+      'weaverbird-popover--always-opened': alwaysOpened,
+      'weaverbird-popover--calculated-height': shouldCalculateHeight,
+    }"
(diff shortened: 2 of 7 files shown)


ui/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+## [0.98.2] - 2022-12-13
+
+### Fixed
+
+- Avoid date picker popover to overflow the window
+- Allow to remove any aggregation in rollup step form
+
```
bloodstorms, line 11: J'avais oublié de pull avant 🤦‍♀️ 
Je corrige les nums de version

ui/src/components/Popover.vue
```diff
@@ -214,4 +227,7 @@ export default class Popover extends Vue {
   visibility: visible;
   z-index: 2;
 }
+.weaverbird-popover--calculated-height {
```
alice-sevin, line 230: Not sure to understand the commentary... The subchild is not applied without declaring the css class here? It's strange... It's should works without it no :thinking: ?

alice-sevin: Or it is just a commentary to notify devs that they should not remove the class herself? On the component ?

bloodstorms: I worry that someone see the unused class in the file and delete it 😅 
but maybe I'm a bit paranoid 🤣 
I can delete it

ui/src/components/domutil.ts
```diff
@@ -258,3 +258,30 @@ export function computeTop(fromBottom = false, ctx: PartialPositionContext) {
...
+   * We assume that if the top of the element is above the middle of the page,
+   * we will have enough space to let it grow.
+   */
+  if (top < bodyHeight / 2) {
+    // leave the default height
+    height = undefined;
+  } else {
+    // avoid the element to overflow the body
+    height = bodyHeight - top;
+  }
```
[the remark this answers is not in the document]
bloodstorms: completely!

ui/src/components/Popover.vue
bloodstorms: At the time I wrote it, it sounded clear to me, but in review you're right, it's weird, I'm going to delete this

ui/src/components/domutil.ts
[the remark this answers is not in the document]
bloodstorms: So, I tried with max-height, but it can't work the same way.

Indeed, as super well explained [here](https://stackoverflow.com/a/14263416/19354626), “when you don't specify an explicit height on the parent, then there's no base height for the child's height to be calculated from, so max-height computes to none, allowing the child to be as tall as possible”

 I made a video to try to explain better the issue.

https://user-images.githubusercontent.com/18734475/207436399-e32cc8ce-d438-4d75-888f-f5fc86e5084a.mp4

I think that the only solution left is to pass the max-height value to the child, but I didn’t succeed to apply the style through slot + component, and it's not really clean 😟

Or we can keep the height property 😛

PR #1666, opened 2022-12-26
fix(backend/pandas): Convert target column to datetime before extracting datetime information TCTC-4825

lukapeschke:
closes TCTC-4823

### DIFF
--- a/server/src/weaverbird/backends/pandas_executor/steps/date_extract.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/date_extract.py
@@ -30,70 +30,75 @@ def execute_date_extract(
         new_columns = step.new_columns
 
     for dt_info, new_col in zip(date_info, new_columns):
-        serie_dt = df[step.column].dt
+        series_dt = to_datetime(df[step.column], utc=True)
+        series_dt_accessor = series_dt.dt
         if dt_info == "week":
             # cast in float and not in int to manage NaN properly
-            result = serie_dt.strftime("%U").astype(float)
+            result = series_dt_accessor.strftime("%U").astype(float)
         elif dt_info == "dayOfWeek":
             # result should be between 1 (sunday) and 7 (saturday)
-            result = (serie_dt.dayofweek + 2) % 7
+            result = (series_dt_accessor.dayofweek + 2) % 7
             result = result.replace({0: 7})
         elif dt_info == "isoYear":
-            result = serie_dt.isocalendar().year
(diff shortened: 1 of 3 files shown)


server/src/weaverbird/backends/pandas_executor/steps/date_extract.py
```diff
@@ -102,45 +107,45 @@ def execute_date_extract(
...
         elif dt_info == "firstDayOfPreviousQuarter":
-            first_month_of_quarter = 3 * ((serie_dt.month - 1) // 3) + 1
+            first_month_of_quarter = 3 * ((series_dt_accessor.month - 1) // 3) + 1
             first_month_of_prev_q = first_month_of_quarter - 3
             first_month_of_prev_q = first_month_of_prev_q.replace({-2: 10})
             result = to_datetime(
                 DataFrame(
                     {
-                        "year": serie_dt.year - (first_month_of_prev_q == 10),
+                        "year": series_dt_accessor.year - (first_month_of_prev_q == 10),
```
PrettyWood, line 116: I don't like the usage of `True == 1 / False == 0` here. It's not golf programming.
Not related to this issue but we should probably rewrite this piece of code to be explicit on the why we have 10, and when we want to do what

server/tests/steps/test_date_extract.py
```diff
@@ -260,10 +248,7 @@ def test_date_extract_no_uint32(sample_df: DataFrame):
 
 
 def test_benchmark_dateextract(benchmark):
-    dates = [
-        datetime.datetime.today() + datetime.timedelta(days=nb_day)
-        for nb_day in list(range(1, 2001))
-    ]
+    dates = [datetime.today() + timedelta(days=nb_day) for nb_day in list(range(1, 2001))]
```
PrettyWood, line 251:
```suggestion
    dates = [datetime.today() + timedelta(days=nb_day) for nb_day in range(1, 2001)]
```

PrettyWood (approved):
LGTM! I approve because my comments are not related to the fix but one of them can be included in this patch

server/src/weaverbird/backends/pandas_executor/steps/date_extract.py
lukapeschke: Agreed, I'll submit a master-only patch for that

lukapeschke: https://github.com/ToucanToco/weaverbird/pull/1669

PR #1668, opened 2022-12-26
fix(backend/pandas): Convert target columns to timestamps before operating on them in evolution step TCTC-4826

lukapeschke:


PrettyWood (approved):
LGTM

PR #1675, opened 2022-12-30
fix(backend/mongo/aggregate): Rename `_id` to `__id` in case it's part of an aggregation's new columns

lukapeschke:
_id is reserved by mongo. Thus, we cannot overwrite it


jGundermann (approved):
LGTM

PR #1678, opened 2022-12-30
fix(backend/mongo/aggregate): Rename _id to __id in case it's part of an aggregation's new columns

lukapeschke:


jGundermann (approved):
LGTM

PR #1689, opened 2023-01-09
fix(backend/pandas): Add support for non-numeric columns to "top" step TCTC-4604

lukapeschke:


Sanix-Darker (commented):
is it possible to have a test that covers this specific change?

lukapeschke:
Woops yes forgot to commit the tests :sweat_smile:

PR #1693, opened 2023-01-10
fix(ui): prevent column name to be undefined when filtering TCTC-4962

davinov:
In prod, I did reproduce a bug where the filter created when we unselect some values from the column menu is broken. The column is `undefined`.
However, I could not reproduce this locally.

I'm suspecting it to come from the columnName prop value when `<ActionMenu` is instantiated.
Let's test this with the deployed environment:


alice-sevin (approved):
let's try it

PR #1694, opened 2023-01-10
fix(ui): load unique values never tries to be applied on undefined column TCTC-4960

davinov:
Similar than #1693 , except I've opted for a more reliable solution :)


alice-sevin (approved):
LGTM :pray:  y a juste le ci qui ne passe pas :cry:

PR #1699, opened 2023-01-11
fix(server/pypika): Allow queries consisting of a single CustomSQL step to be materialized when no columns are specified

lukapeschke:


server/tests/backends/sql_translator_unit_tests/test_base_translator.py
```diff
@@ -767,3 +767,16 @@ def test_no_extra_quotes_in_base_translator_with_entire_pipeline(base_translator
...
+
+def test_materialize_customsql_query_with_no_columns(base_translator: BaseTranslator):
+    pipeline = [steps.CustomSqlStep(query="SELECT titi, tata FROM toto")]
+
+    # For CustomSQL, we must have exactly one table
+    base_translator._tables_columns = {"toto": []}
+
+    translated = base_translator.get_query_str(steps=pipeline)
+    assert translated == (
+        "WITH __step_0_basetranslator__ AS (SELECT titi, tata FROM toto) "
```
Sanix-Darker, line 780:
```suggestion
        'WITH __step_0_basetranslator__ AS (SELECT titi, tata FROM toto) '
```

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -128,7 +128,15 @@ class QueryBuilderContext:
     table_name: str
 
     def materialize(self) -> QueryBuilder:
-        return self.builder.from_(self.table_name).select(*self.columns)
+        # In case we have no columns, select everything by default.
```
Sanix-Darker, line 131: i feel like it could be interessant to add an if check here on the step name too... what do you think ? (no need ? too much ?)

Sanix-Darker (approved):

server/src/weaverbird/backends/pypika_translator/translators/base.py
lukapeschke: We can't, as the step name is not available here anymore

PR #1703, opened 2023-01-17
feat(disco): multiStatement queries [TCTC-5057]

Sanix-Darker:
# MULTIPLE QUERIES STATEMENTS

## WHY

Today it's not possible for our AppBuilders to make custom SQL step with multiple statements, 
Ex:
```sql
SELECT set_config("my_user_id", 12345);
SET TIME ZONE 'Europe/Paris';
/* actual query */
SELECT ... FROM "my_db"."my_table"
```

This discovery aims to be able to find some solutions to fix that problem.

On our backend, we have a feature flag called `vqb_translator`. 
This flag switch the use of our pypika backend (to run a query natively on the specified engine, postgreql/mysql) when building a query or our old behavior with our sql connector directly.

We want to have this work on both cases, flag activates or not.

## WITHOUT PYPIKA, THE OLD WAY with connectors (flag deactivated)

When the vqb_translator is disabled, it's already possible to mix multiple sql queries since the only thing we're doing here is to pass that
raw query to the `pandas.read_sql` method with all connections parameters needed for the query to be executed, I tested this on my instance.

## WITH PYPIKA (flag activated)

On Pypika, customSQL is disabled in the middle of the pipeline because it's impossible for our backend to have the context of available tables or columns.

As a solution for this scope, we can split and Parse the content of the SQL query string and transform it into dynamic objects in python we can manipulate!

- A simple solution could be to use [sqlparse](https://pypi.org/project/sqlparse/) first which is a library we can use to split the raw string in multiple queries and build customstep from those chunks if they are well described (tables and columns provided)
```python
>>> import sqlparse
>>> from weaverbird.pipeline.steps.custom import CustomStep


>>> # Split a string containing three SQL statements:
>>> raw = """
SELECT set_config("my_user_id", 12345);
SET TIME ZONE 'Europe/Paris';

SELECT column1 FROM "my_db"."my_table"
"""
>>> statements = sqlparse.split(raw)
>>> statements
[
    "SELECT set_config('my_user_id', 12345);",
    "SET TIME ZONE 'Europe/Paris';",
    "SELECT column1 FROM 'my_db'.'my_table'"
]
>>> [CustomStep(name=f"cu_{i}", query=f"{q}") for i, q in enumerate(statements)]
# assuming the customsql pypika not raising an error anymore and using the implementation 
# done here https://github.com/ToucanToco/weaverbird/blob/master/server/src/weaverbird/backends/pypika_translator/translators/base.py#L226
```

- But let's say we want to obtain as a result of a given sql query string, Pypika 'like' objects 
This means we need to write our own Parser/Lexer to obtain such a result.
To achieve this, we could use [SLY (Sly Lex-Yacc)](https://github.com/dabeaz/sly) which is a Python implementation of the lex and yacc tools commonly used to write parsers and compilers. Parsing is based on the same LALR(1) algorithm used by many yacc tools.
However, this may sound heavy to do especially if we want to cover everything...
In their documentation, there is a section dedicated to explaining how to write a Parser/Lexer: https://sly.readthedocs.io/en/latest/sly.html#writing-a-parser 
An example of POC that we can use to tokenize a query string is available here for the Parser: https://github.com/ToucanToco/weaverbird/pull/1703/commits/1b493338e68c30cb0b7c9b5c19f2ede8001f2936

Before going further with this proposal, it could be great to know if it's really something we want to spend the time on!


Sanix-Darker:
> I don't think having custom queries in the middle of a pipeline is in the scope of this disco. Also, is it really a legitimate use case ? Are there cases where you need to use custom SQL in the middle of a pipeline and it cannot be done as the first step ?

The idea I was trying to come up with was to convert multi-statement queries into a separated customSQL step (if all conditions are good for it, like the type of the query, if there is a table and columns well described...) with `sqlparse.split` !

Sanix-Darker:
> Is it really working with all different backends ? Have you tested connectors where queries get wrapped to insert extra OFFSET and LIMIT parameters ?

In most cases, custom raw multistatement queries work fine, **however**, it's true, there are a lot of edge cases, I tested on my instance the sample provided by the Jira card only and didn't check further for problematic cases!
Screenshot from 2023-01-18 11-59-02

Sanix-Darker:
Another simple and cheap solution can be to split the multiquery statement and build custom queries at the top step, see https://github.com/ToucanToco/weaverbird/pull/1703/commits/f6cebae7f76b19500eaa14fb0a3a3093dffffe40 and https://github.com/ToucanToco/weaverbird/pull/1703/commits/a1b4f15984eb1c35e0c599eae8d59cf589a24f2a

lukapeschke:
As explained previously, I believe we should never have several Custom SQL steps in our pipeline. If you want to go with a solutions creating multiple CTEs, it should be done with a single Custom SQL step object.

Also, before going any further, could you please add a mutli-statement test fixture, so we can validate that the multi-CTE approach is actually working ?

server/tests/backends/fixtures/customsql/multistatements.yml
```diff
@@ -0,0 +1,44 @@
...
+    fields:
+    - name: name
+      type: string
+    pandas_version: 1.4.0
+step:
+  name: _customsql
+  pipeline:
+  - name: _customsql
+    query: |
+      /* Are we going to support SET or not ? */
```
lukapeschke, line 42: We probably will have to, as it is a requirement for RLS

lukapeschke (commented):
Thanks for the fixture. Could you please update it to include statements that actually have an impact on later statements ? For example one `set_config`, one `SET` (or `SET LOCAL`), and finally a `SELECT` statement that uses the previously set options

fspot:
Sinon on pourrait modifier dans toucan-connectors les modèles des data sources de type sql (ex: `PostgresDataSource`), en leur ajoutant un field du style `initial_statements: list[str] | None = None`. En terme d'UI on pourrait définir une valeur pour ce field lorsqu'on crée une query, au même moment où on sélectionne la table / les colonnes etc.

Ça demanderait de modifier un peu le code des connecteurs sql pour prendre en compte ces statements, par exemple pour le bout de code dont on parle + haut ça donnerait : 

```python
raw_query = f"""SELECT * FROM ({data_source.query.replace(';','')}) AS q LIMIT 0;"""
if stmts := data_source.initial_statements:
    raw_query = f"{';'.join(stmts)}; {raw_query}"
cursor.execute(raw_query)
```

(du coup ça n'aurait aucun rapport avec weaverbird au final)

PrettyWood:
@fspot l'idee me plait perso !

Sanix-Darker:
already handled

PR #1715, opened 2023-02-02
chore(ui): v1.0.0

alice-sevin:


bloodstorms (approved):
🎉

alice-sevin:
Closed because we do v0.101 instead, we can have breaking changes on V0

PR #1759, opened 2023-04-27
fix(server/tests): Skip flaky MySQL tests

lukapeschke:


server/tests/backends/sql_translator_integration_tests/test_sql_mysql_translator_steps.py
```diff
@@ -76,6 +76,7 @@ def mysql_container():
 @pytest.mark.parametrize(
     "case_id, case_spec_file_path", retrieve_case("sql_translator", "mysql_pypika")
 )
+@pytest.mark.skip("MySQL result order is not consistent with CTEs")
```
Sanix-Darker, line 79: Can we add a FIXME comment for this to be fixed in the future? WDYT ?

Sanix-Darker (approved):

server/tests/backends/sql_translator_integration_tests/test_sql_mysql_translator_steps.py
lukapeschke: We have an associated confluence entry + I don't know if a fixme is more visible than a skip marker :man_shrugging:

PR #1777, opened 2023-05-22
fix: we need to cast to integer a column resulting from a date extract when needed [TCTC-6089]

Sanix-Darker:
## WHAT

We need to cast all columns coming from a "date extract" step to have only an integer value when needed, not float.

## WHY

resolves : TCTC-6089

## BEFORE
before2

## AFTER
after2

[TCTC-6089]: https://toucantoco.atlassian.net/browse/TCTC-6089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/src/weaverbird/pipeline/steps/date_extract.py
```diff
@@ -45,6 +49,20 @@
     ],
 ]
 
+DATE_UNIT_WITH_INT_CAST_NOT_NEEDED: list[str] = [
```
lukapeschke, line 52: Seems like a duplicate of `FIRST_DAY_DATE_PARTS`, is there a reason to have both ? Also, why is this all lowercase ?

server/src/weaverbird/pipeline/steps/date_extract.py
```diff
@@ -23,20 +23,24 @@
     "milliseconds",
 ]
 
+FIRST_DAY_DATE_PARTS = Literal[
```
lukapeschke, line 26: This name might be misleading as it also contains `previousDay`. Also, the name should be and indication about why it is needed. From what I understand, these should be timestamp types in the end, right ? how about distinguishing between `TIMESTAMP_DATE_PARTS` and `INTEGER_DATE_PARTS` ?

lukapeschke (changes requested):
Thanks for the fix ! Could you please add a test fixture for native SQL reproducing the bug you're encountering ?

server/src/weaverbird/pipeline/steps/date_extract.py
Sanix-Darker: One is a list, the other one is a type declaration

server/src/weaverbird/pipeline/steps/date_extract.py
Sanix-Darker: Good idea 👌🏽

Sanix-Darker:
> Could you please add a test fixture for native SQL reproducing the bug you're encountering ?

I think this is what this fixture already do : https://github.com/ToucanToco/weaverbird/blob/3a3d9efe1d700fc10c81956343dad165b3506276/server/tests/backends/fixtures/dateextract/pypika_all_cases.json#L1

Is not this Unit Test enough to guarantee this Pull Request changes : https://github.com/ToucanToco/weaverbird/blob/3a3d9efe1d700fc10c81956343dad165b3506276/server/tests/backends/sql_translator_unit_tests/test_base_translator.py#L784 @lukapeschke  ?

lukapeschke:
> I think this is what this fixture already do :

Well it was working before your fix, so no, it does not validate this bug

> Is not this Unit Test enough to guarantee this Pull Request changes :

It tests how the query is built, not that it works

Sanix-Darker:
Well, after more tests, i noticed that for the scope of this card, this bug concern only pandas and not pypika, that's why i revert changes made for the pypika_translator

lukapeschke:
> Well, after more tests, i noticed that for the scope of this card, this bug concern only pandas and not pypika, that's why i revert changes made for the pypika_translator

Are you sure ? The ticket mentions a NativeSQL translation error

Sanix-Darker:
> Are you sure ? The ticket mentions a NativeSQL translation error

ScreenShots (from the ticket card) are telling a different story I guess? there are dataframe errors 

image-20230522-151455

lukapeschke:
Right, but from another screenshot it seems like an invalid query can be generated by the pypika translator... Anyway, if an UAT with this branch fixes the problem, just add an UT validating the pandas issue and we'll be good :+1:

Sanix-Darker:
> Right, but from another screenshot it seems like an invalid query can be generated by the pypika translator... Anyway, if a UAT with this branch fixes the problem, just add a UT validating the pandas issue and we'll be good +1

What do you think about this UT: https://github.com/ToucanToco/weaverbird/pull/1777/commits/d4fe9333b20a4f21aff443b35788398ba2876931 @lukapeschke ?

lukapeschke:
> > Right, but from another screenshot it seems like an invalid query can be generated by the pypika translator... Anyway, if a UAT with this branch fixes the problem, just add a UT validating the pandas issue and we'll be good +1
> 
> What do you think about this UT: [d4fe933](https://github.com/ToucanToco/weaverbird/commit/d4fe9333b20a4f21aff443b35788398ba2876931) @lukapeschke ?

Looks good, but I don't think the pandas functions should be patched: we want to test the actual behaviour since it was failing before your fix

lukapeschke:
LGTM! Do you have a guru for UATs ?

lukapeschke (approved):

PR #1795, opened 2023-06-08
feat(server/pypika): Merge first two steps when possible 

lukapeschke:


Sanix-Darker (approved):
✨ Thanks for this PR 👌🏽

PR #1797, opened 2023-06-12
feat: add remove void from mongo step [TCTC-6102]

Sanix-Darker:
## WHAT

We need to remove void steps from mongo queries from weaverbird like we do with filter steps
on normal pipelines


lukapeschke:
How can a mongo pipeline with `__VOID__` steps happen ? I thought we'd always translate the pipeline ourselves, so why can't we remove the related steps ?

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -256,6 +256,63 @@ def remove_void_conditions_from_filter_steps(
     return final_steps
 
 
+def _is_empty(data: dict[str, Any] | list[Any]) -> bool:
+    if isinstance(data, dict):
+        return len(data) == 0
+    elif isinstance(data, list):
+        return len(data) == 0
+    return False
```
lukapeschke, line 264:
```suggestion
    return len(data) == 0
```

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -256,6 +256,63 @@ def remove_void_conditions_from_filter_steps(
...
+            k: _remove_empty_elements_that_has_empty_value(v)
+            for k, v in data.items()
+            if not _is_empty(v)
+        }
+    elif isinstance(data, list):
+        return [
+            _remove_empty_elements_that_has_empty_value(item)
+            for item in data
+            if not _is_empty(item)
+        ]
```
lukapeschke, line 284: We might still have empty elements here: if `_remove_empty_elements_that_has_emtpy_value` returns an emtpy dict. Is that the desired behaviour ?

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -256,6 +256,63 @@ def remove_void_conditions_from_filter_steps(
...
 
+def _is_empty(data: dict[str, Any] | list[Any]) -> bool:
+    if isinstance(data, dict):
+        return len(data) == 0
+    elif isinstance(data, list):
+        return len(data) == 0
+    return False
+
+
+def _remove_empty_elements_that_has_empty_value(
```
lukapeschke, line 267: I don't understand the function name... It's supposed to remove emtpy elements right ?

```suggestion
def _remove_empty_elements(
```

lukapeschke (commented):
Ok I see, thanks for the clarification

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -256,6 +256,74 @@ def remove_void_conditions_from_filter_steps(
...
+                data_transformed := {
+                    k: _remove_empty_elements(v)
+                    for k, v in data.items()
+                    if v is not None and not len(v) == 0
+                }
+            )
+            == 0
+        ):
+            return None
+        return data_transformed
```
lukapeschke, line 278: I think this code can be improved:
* The outer `if` can be replaced by a truthiness check
* The return value of `_remove_empty_elements(v)` should be checked

```suggestion
        data_transformed = {
              k: cleaned
              for k, v in data.items()
              if (cleaned := _remove_empty_elements(v)) is not None
        }
        return data_transformed or None
```

A similar form can be used for the `list` case below

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -256,6 +256,74 @@ def remove_void_conditions_from_filter_steps(
     return final_steps
 
 
+def _remove_empty_elements(
+    data: dict[str, Any] | list[Any] | None
+) -> dict[str, Any] | list[Any] | None:
```
lukapeschke, line 261: I believe this signature is wrong: You're calling the function on the values of a `dict[str, Any]`. Therefore, it should accept `Any`

```suggestion
def _remove_empty_elements(data: Any) -> Any:
```

server/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Now we can clean mongo steps containing `__VOID__` as values with `pipeline.remove_void_conditions_from_mongo_steps()`.
```
lukapeschke, line 5:
```suggestion
- `__VOID__` values can now be removed from mongo queries with `weaverbird.pipeline.remove_void_conditions_from_mongo_steps()`.
```

Please use simple statements for release notes

lukapeschke (approved):

Sanix-Darker:
I made new updates to the PR @lukapeschke , please let me know wdyt, if it's all good, I will merge it and bump a new version.

server/tests/test_pipeline.py
```diff
@@ -285,3 +285,43 @@ def test_skip_void_parameter_from_variables_for_mongo_steps():
...
+                }
+            },
+            {"$unwind": "$_vqbPipelinesUnion"},
+            {"$replaceRoot": {"newRoot": "$_vqbPipelinesUnion"}},
+            {"$project": {"_id": 0}},
+        ]
+    ) == [
+        # NOTE: the $match step is rip off because of the cleaning on empty elements,
+        # it's the responsability to the client to add that {$match: {}} as the
+        # first steps
```
lukapeschke, line 314: This sounds dangerous. Why would it be removed ? There's some sanitization logic in toucan-connectors already, maybe you can take inspiration from there ? https://github.com/ToucanToco/toucan-connectors/blob/master/toucan_connectors/mongo/mongo_connector.py#L76

Sanix-Darker: > Why would it be removed ?

because by definition of the function, if a given key/value is given and the value is an empty dict, there is no point keeping it ! i already fix that here on laputa : https://github.com/ToucanToco/laputa/pull/5956/files#diff-20f538e207e705508e72573b79c68ceeb8454d11d85169c6838e8353d9ba2ba5R674-R675
+ 
am not sure that that sanitizer fix this related issue of the missing `$match` key only... or am wrong ? 🤔

Sanix-Darker: https://github.com/ToucanToco/weaverbird/pull/1797/commits/76ba2b3ed072815645eddc2f6f31fcc2050c3a65 i added the sanitizer method

Sanix-Darker:
> Est-ce qu'il n'aurait pas été + simple d'harmoniser la façon dont on gère l'interpolation / la traduction pour éviter d'avoir à réimplémenter ça ? Ou est-ce que ça aurait posé d'autres soucis ? thinking

D'accord avec toi, mais, "Plus simple", je ne suis pas certain (je peux me tromper)... car cette harmonisation inclut des refactos plus grandes que le scope de cette carte, car il faudra les faire dans 3 repos au moins, mais si je resumes bien, on parles bien de : 
- supprimer l'interpolation des variables qui se fait de tous les cotes(laputa (vqb), connectors(mongo), ...) pour en faire qu'a un seul endroit et pour les 2 types de pipelines ?
- repondre a la question de savoir : qui geres l'interpolation des trusted/untrusted et comment ?

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -319,16 +314,63 @@ def remove_void_conditions_from_mongo_steps(
...
+        return isinstance(d, dict) and list(d.keys()) == ["$match"]
+
+    def _sanitize_match(query: dict) -> dict:
+        if _is_match_empty(query):
+            return {}
+        if "$and" in query:
+            and_condition = query["$and"]
+            if isinstance(and_condition, list):
+                query["$and"] = [elem for elem in and_condition if not _is_empty_match_column(elem)]
+        return query
```
lukapeschke, line 355: Why do these have to be closures ?

lukapeschke (commented):
If this is blocking for you, we can try like this.. Just move the closures out of the function body and we can merge this

Sanix-Darker:
Done here : https://github.com/ToucanToco/weaverbird/pull/1797/commits/52c49b62

PR #1804, opened 2023-06-19
feat(append/join): use uid refs as source of join/append step and disable unjoinable datasets [TCTC-6262][TCTC-6261]

alice-sevin:
1. Enable to use uid as source join/append step (before we only passed dataset name, it will create problems with potential dataset renaming).
2. Add a new unjoinableDataset state property in store that can be used to disable options in join/append step.

This PR should have no impact on existing content, it should not include regressions of any type.


davinov:
As this contains also the PR #1805, I squashed all the other commits into one, so we can merge this PR __without__ squashing

davinov (approved):

PR #1805, opened 2023-06-21
feat(server): new model PipelineWithRefs, that can resolve its references

davinov:
Adds support for references and their resolution in weaverbird's backends.
Also removes the old references resolving that happened front-end side.


davinov:
On a encore des soucis sur le playground à cause du dereferencePipelines qui est encore fait côté front, à cleaner.

PR #1818, opened 2023-07-06
fix: keep None/null on $ne and $eq direct value to fix isnull/isnotnull filter step condition for mongo [TCTC-6339]

Sanix-Darker:
## WHAT

While cleaning `__VOID__` from a mongo pipeline, we introduced a regression [HERE](https://github.com/ToucanToco/weaverbird/pull/1797) because we were removing also normal/casual `$ne/$eq` direct isnull/isnotnull values from filter condition,.

## HOW
This PR is to isolate those specific cases and adapt the test for the fix.

### BEFORE
before

### AFTER

after

### DIFF
--- a/server/src/weaverbird/pipeline/pipeline.py
+++ b/server/src/weaverbird/pipeline/pipeline.py
@@ -194,6 +194,8 @@ PipelineStepWithVariables = Annotated[
 
 VOID_REPR = "__VOID__"
 EXCLUDE_CLEANING_FOR = (
+    "$ne",  # for isnotnull (None -> null for mongo)
+    "$eq",  # for isnull (None -> null for mongo)
     "pipeline",
     "localField",
     "foreignField",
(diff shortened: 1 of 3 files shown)


server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -371,7 +373,8 @@ def _sanitize_query_matches(query: dict | list[dict]) -> Any:
 def remove_void_conditions_from_mongo_steps(
     mongo_steps: dict | list[dict],
 ) -> dict | list[dict]:
-    return _sanitize_query_matches(_clean_mongo_steps(mongo_steps) or [])
+    mongo_query = _sanitize_query_matches(_clean_mongo_steps(mongo_steps) or [])
+    return mongo_query
```
jGundermann, line 377: Oups ?

jGundermann (approved):

fspot (approved):

PR #1820, opened 2023-07-07
fix: mongo rank step (due to remove_void regression) [TCTC-6370]

Sanix-Darker:
## WHAT

Yet another regression introduce by the remove_void from mongo steps.

## HOW

Fixed by what caracterize the rank step (some specifics columns that needed to be None).

### DIFF
--- a/server/src/weaverbird/pipeline/pipeline.py
+++ b/server/src/weaverbird/pipeline/pipeline.py
@@ -194,7 +194,9 @@ PipelineStepWithVariables = Annotated[
 
 VOID_REPR = "__VOID__"
 EXCLUDE_CLEANING_FOR = (
-    "$ne",  # for isnotnull (None -> null for mongo)
+    "prevValue",  # for the rank step, this need to be keep
+    "prevRank",  # for the rank step, this need to be keep
+    "$ne",  # for isnotnull (None -> null for mongo),
     "$eq",  # for isnull (None -> null for mongo)
     "pipeline",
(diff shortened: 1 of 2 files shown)


fspot (approved):

server/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog (weaverbird python package)
 
+### Changed
+
+- Fix[Mongo]: Fix mongo rank step by ignoreing some specific keys that defines it and should be consider as 'None'.
```
PrettyWood, line 5:
```suggestion
- Fix[Mongo]: Ignore in the mongo rank step some specific keys that define it and should be considered as 'None'.
```

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -194,7 +194,9 @@ def dict(self, *, exclude_none: bool = True, **kwargs) -> dict:
 
 VOID_REPR = "__VOID__"
 EXCLUDE_CLEANING_FOR = (
-    "$ne",  # for isnotnull (None -> null for mongo)
+    "prevValue",  # for the rank step, this need to be keep
+    "prevRank",  # for the rank step, this need to be keep
```
PrettyWood, line 198:
```suggestion
    "prevValue",  # for the rank step, this needs to be kept
    "prevRank",  # for the rank step, this needs to be kept
```

PR #1832, opened 2023-07-17
fix(join-step): retrieve columns for query behind ref

alice-sevin:
Enable to retrieve columns names for a query behind ref (before we passed domain name while we should pass reference)


davinov (approved):
Bien vu :+1:

PR #1838, opened 2023-07-24
fix: keep empty list to make mongo steps still working as before [TCTC-6370]

Sanix-Darker:
## WHAT

While removing __VOID__ from mongo pipeline, we were also removing key/value with empty list or empty dict !
But after some investigations, we noticed that some steps like the `rank` needs some of it keys to be empty such as `'a':[]` or `pipeline:[]`
In this example :
```
[
            {"$match": {"domain": "cars"}},
            {"$sort": {"color": 1}},
            {"$group": {"_id": None, "_vqbArray": {"$push": "$$ROOT"}}},
            {
                "$project": {
                    "_vqbSortedArray": {
                        "$let": {
                            "vars": {
                                "reducedArrayInObj": {
                                    "$reduce": {
                                        "input": "$_vqbArray",
                                        "initialValue": {
                                            "a": [],
                                            "order": 0,
                                            "prevValue": None,
                                            "prevRank": None,
                                        },
                                        "in": {
                                            "$let": {
                                                "vars": {
                                                    "order": {
                                                        "$cond": [
                                                            {
                                                                "$ne": [
                                                                    "$$this.color",
                                                                    "$$value.prevValue",
                                                                ]
                                                            },
                                                            {"$add": ["$$value.order", 1]},
                                                            "$$value.order",
                                                        ]
                                                    },
                                                    "rank": {
                                                        "$cond": [
                                                            {
                                                                "$ne": [
                                                                    "$$this.color",
                                                                    "$$value.prevValue",
                                                                ]
                                                            },
                                                            {"$add": ["$$value.order", 1]},
                                                            "$$value.prevRank",
                                                        ]
                                                    },
                                                },
                                                "in": {
                                                    "a": {
                                                        "$concatArrays": [
                                                            "$$value.a",
                                                            [
                                                                {
                                                                    "$mergeObjects": [
                                                                        "$$this",
                                                                        {"color-rank": "$$rank"},
                                                                    ]
                                                                }
                                                            ],
                                                        ]
                                                    },
                                                    "order": "$$order",
                                                    "prevValue": "$$this.color",
                                                    "prevRank": "$$rank",
                                                },
                                            }
                                        },
                                    }
                                }
                            },
                            "in": "$$reducedArrayInObj.a",
                        }
                    }
                }
            },
            {"$unwind": "$_vqbSortedArray"},
            {"$replaceRoot": {"newRoot": "$_vqbSortedArray"}},
        ]
```

For it to work.

### DIFF
--- a/server/src/weaverbird/pipeline/pipeline.py
+++ b/server/src/weaverbird/pipeline/pipeline.py
@@ -198,7 +198,6 @@ EXCLUDE_CLEANING_FOR = (
     "prevRank",  # for the rank step, this needs to be kept
     "$ne",  # for isnotnull (None -> null for mongo),
     "$eq",  # for isnull (None -> null for mongo)
-    "pipeline",
     "localField",
     "foreignField",
     "_id",
@@ -280,14 +279,22 @@ def _remove_empty_elements(data: Any) -> Any:
                 if (cleaned := _remove_empty_elements(v)) is not None:
                     data_transformed[k] = cleaned
 
-        return data_transformed or None
+        if isinstance(data_transformed, list):
+            return data_transformed
+        else:
+            return data_transformed or None
+
     elif isinstance(data, list):
         data_transformed = [
             cleaned
             for item in data
             if (cleaned := _remove_empty_elements(item)) is not None  # type: ignore[assignment]
         ]
-        return data_transformed or None
+
+        if isinstance(data_transformed, list):
+            return data_transformed
+        else:
+            return data_transformed or None
     else:
         return data
 
@@ -308,10 +315,6 @@ def _clean_mongo_steps(
                 step[key] = val
             else:
                 step[key] = _clean_mongo_steps(val)  # type: ignore[assignment]
-                # FIXME: This may happens for more other keys but on mongo
-                # 'pipeline' cannot be None/null, it should be an empty []
-                if key == "pipeline" and step[key] is None:
-                    step[key] = []  # type:ignore[assignment]
 
         return _remove_empty_elements(step)
     elif isinstance(mongo_steps, list):
--- a/server/src/weaverbird/pipeline/pipeline.py
+++ b/server/src/weaverbird/pipeline/pipeline.py
@@ -291,10 +291,7 @@ def _remove_empty_elements(data: Any) -> Any:
             if (cleaned := _remove_empty_elements(item)) is not None  # type: ignore[assignment]
         ]
 
-        if isinstance(data_transformed, list):
-            return data_transformed
-        else:
-            return data_transformed or None
+        return data_transformed
     else:
         return data
 
--- a/server/src/weaverbird/pipeline/pipeline.py
+++ b/server/src/weaverbird/pipeline/pipeline.py
@@ -279,6 +279,9 @@ def _remove_empty_elements(data: Any) -> Any:
                 if (cleaned := _remove_empty_elements(v)) is not None:
                     data_transformed[k] = cleaned
 
+                if cleaned == [] and k in ["$or", "$nor", "$and"]:
+                    data_transformed[k] = None
+
         if isinstance(data_transformed, list):
             return data_transformed
         else:
--- a/server/tests/test_pipeline.py
+++ b/server/tests/test_pipeline.py
@@ -285,16 +285,7 @@ def test_skip_void_parameter_from_variables_for_mongo_steps():
                 }
             }
         ]
-    ) == [
-        {
-            "$match": {
-                "$and": [{"$or": [{"property2": "value2"}]}, {"$nor": []}],
-                "$ne": None,
-                "$nor": [{"property9": {"$and": []}}],
-                "$or": [],
-            }
-        },
-    ]
+    ) == [{"$match": {"$and": [{"$or": [{"property2": "value2"}]}], "$ne": None}}]
 
     assert remove_void_conditions_from_mongo_steps(
         [
--- a/server/tests/test_pipeline.py
+++ b/server/tests/test_pipeline.py
@@ -285,7 +285,16 @@ def test_skip_void_parameter_from_variables_for_mongo_steps():
                 }
             }
         ]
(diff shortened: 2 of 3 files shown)


server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -280,14 +279,22 @@ def _remove_empty_elements(data: Any) -> Any:
...
             cleaned
             for item in data
             if (cleaned := _remove_empty_elements(item)) is not None  # type: ignore[assignment]
         ]
-        return data_transformed or None
+
+        if isinstance(data_transformed, list):
+            return data_transformed
+        else:
+            return data_transformed or None
```
lukapeschke, line 297: The `isinstance` check is redundant. The comprehension could be returned directly here

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -198,7 +198,6 @@ def dict(self, *, exclude_none: bool = True, **kwargs) -> dict:
     "prevRank",  # for the rank step, this needs to be kept
     "$ne",  # for isnotnull (None -> null for mongo),
     "$eq",  # for isnull (None -> null for mongo)
-    "pipeline",
```
lukapeschke, line 201: Why has this behaviour been changed ?

server/tests/test_pipeline.py
```diff
@@ -285,7 +285,16 @@ def test_skip_void_parameter_from_variables_for_mongo_steps():
...
             }
         ]
-    ) == [{"$match": {"$and": [{"$or": [{"property2": "value2"}]}], "$ne": None}}]
+    ) == [
+        {
+            "$match": {
+                "$and": [{"$or": [{"property2": "value2"}]}, {"$nor": []}],
+                "$ne": None,
+                "$nor": [{"property9": {"$and": []}}],
+                "$or": [],
```
lukapeschke, line 294: Isn't this match always false ?

server/src/weaverbird/pipeline/pipeline.py
Sanix-Darker: ah yes, true. updated here : https://github.com/ToucanToco/weaverbird/pull/1838/commits/75bdda95

server/src/weaverbird/pipeline/pipeline.py
Sanix-Darker: because the same behaviour with pipeline is the one i was covering here : https://github.com/ToucanToco/weaverbird/blob/42c08d9a704e593fc8ddd4199ea92c432bbfd808/server/src/weaverbird/pipeline/pipeline.py#L311-L314

server/tests/test_pipeline.py
Sanix-Darker: and why that please ? 🤔 
This is just a dummy test on how the removal behave on empty dict/list...

lukapeschke: Because this would try to match an `$or` condition with an empty predicate list. If mongo's `$or` operator is defined as "at least one predicate is true", then this would always be false

lukapeschke: If the list is empty, it's probably because a predicate containing a `__VOID__` was removed. So it's not necessarily up to the AB

Sanix-Darker: mhhh... no 🤔 
please check again the assertion instruction, there is no `__VOID__` over there, I don't get well your point here !
```python
    assert remove_void_conditions_from_mongo_steps(
        [
            {
                "$match": {
                    "$ne": None,
                    "$and": [
                        {"$or": [{"property2": "value2"}, {"property3": {}}]},
                        {"$nor": [{"property4": {}}, {"property5": {}}]},
                    ],
                    "$or": [],
                    "$nor": [
                        {
                            "property9": {
                                "$and": [
                                    {
                                        "property7": {},
                                    }
                                ]
                            }
                        }
                    ],
                }
            }
        ]
    ) == [{"$match": {"$and": [{"$or": [{"property2": "value2"}]}], "$ne": None}}]
    ) == [
        {
            "$match": {
                "$and": [{"$or": [{"property2": "value2"}]}, {"$nor": []}],
                "$ne": None,
                "$nor": [{"property9": {"$and": []}}],
                "$or": [],
            }
        },
    ]
```

any suggestion could help 🙏🏽

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -280,14 +279,22 @@ def _remove_empty_elements(data: Any) -> Any:
                 if (cleaned := _remove_empty_elements(v)) is not None:
                     data_transformed[k] = cleaned
 
-        return data_transformed or None
+                if cleaned == [] and k in ["$or", "$nor", "$and"]:
+                    data_transformed[k] = None
```
lukapeschke, line 283: Shouldn't it just be omitted in case the key is `$or`, `$and` or `$nor` ? I don't think `$and: null` is valid mongo syntax

Sanix-Darker: 🤔 We will never have the case of `$and: null` at the end of the transformation !
See the updated test, please!
```python
    assert remove_void_conditions_from_mongo_steps(
        [
            {
                "$match": {
                    "$ne": None,
                    "$and": [
                        {"$or": [{"property2": "value2"}, {"property3": {}}]},
                        {"$nor": [{"property4": {}}, {"property5": {}}]},
                    ],
                    "$or": [],
                    "$nor": [
                        {
                            "property9": {
                                "$and": [
                                    {
                                        "property7": {},
                                    }
                                ]
                            }
                        }
                    ],
                }
            }
        ]
    ) == [{"$match": {"$and": [{"$or": [{"property2": "value2"}]}], "$ne": None}}]
```

lukapeschke: Woops sorry :facepalm:

lukapeschke (approved):
:ok_hand:

PR #1845, opened 2023-07-31
feat(ui): Added rank step to pypika translator

lukapeschke:


davinov (approved):
Moarf, nice catch. Hopefully this will be generated in a near future!

PR #1846, opened 2023-08-01
chore: add fixture to reproduce [TCTC-5400]

fspot:
(works for mongo and pandas, not for pypika)

### DIFF
--- a/server/tests/backends/fixtures/aggregate/multiple_aggregations_with_window_func_without_on_pypika.json
+++ b/server/tests/backends/fixtures/aggregate/multiple_aggregations_with_window_func_without_on_pypika.json
@@ -1,0 +1,105 @@
+{
+  "exclude": [
+    "snowflake"
+  ],
+  "step": {
+    "pipeline": [
+      {
+        "name": "aggregate",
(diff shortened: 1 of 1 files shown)


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -434,12 +434,20 @@ def _build_window_subquery() -> Any:
...
+                    .inner_join(all_windows_subquery)
+                    .on_field(*step.on)
+                )
+            else:
+                # If there is no `step.on` columns to join, just put the 2 subqueries side by side:
+                merged_query = (
+                    self.QUERY_CLS.from_(agg_query)
+                    .from_(all_windows_subquery)
+                    .select(*merged_selected)
+                )
```
lukapeschke, line 450: Nice :+1: I think you'll have to do a modification like that l.463 as well: In case `keep_original_granularity` is specified, there's also a join on `step.on`

fspot: done in 4f9b6e1

PR #1855, opened 2023-08-09
feat(sql): change how first and last aggregation are translated [TCTC-6537]

davinov:
There is no easy FIRST and LAST aggregation functions in SQL. The current approach with window functions is very inefficient, and causes queries to time out. For now, we suggest a quick fix: to use min and max for first and last respectively. This is way faster is many situations.

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog (weaverbird python package)
 
+## Unreleased
+
+### Changed
+
+- SQL translation of the `first` and `last` aggregations was implemented with window functions. However, this is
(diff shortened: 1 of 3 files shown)


PrettyWood (approved):

fspot (approved):

Sanix-Darker (approved):
Thanks for this update, if possible, can you also add a new entry in the CHANGELOG.md ? 🙏🏽

PrettyWood (approved):

davinov:
After discussion, let's avoid merging this for now. We'll see how behave our uses cases with less of these aggregations before taking such a radical measure ;)

PR #1856, opened 2023-08-10
feat: refactoring pipeline with references

Sanix-Darker:
- merge pipelineRefs with the actual pipeline module
- update all dependencies related to that change


davinov (changes requested):
Thanks for the suggestion, however I'm not sure to see what it brings in terms to consumers.

To me, the separation of concerns between the pipeline and the references module seems an advantage.

I'm really open to change my mind if you bring relevant arguments to this or if most of others devs agree that this is a good way forward.

Sanix-Darker:
let's keep it like that 😉  + this may add more changes to all repos that depends on weaverbird !

PR #1857, opened 2023-08-10
fix: adapt custom date formats for SQL engines [TCTC-6547]

PrettyWood:


fspot (approved):
:+1:

davinov (approved):
:+1: top! Makes it easy to declare a new mapping for another system :+1:

PR #1859, opened 2023-08-11
fix(sql): if then else step can accept numerical values

davinov:
I noticed that ifthenelse steps with numbers as results were not accepted by the Pypika translators with the following errors: `('Could not translate parameters to a SQL query', AttributeError("'int' object has no attribute 'encode'"))``

Example of such a step:
```js
{
  name: 'ifthenelse',
  if: { column: 'order', operator: 'eq', value: 'first' },
  then: 0,
  else: {
    if: { column: 'order', operator: 'eq', value: 'second' },
    then: 1,
    else: {
      if: { column: 'order', operator: 'eq', value: 'third' },
      then: 2,
      else: 3,
    },
  },
  newColumn: 'order',
}
```

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Fix[PyPika]: ithenelse step supports numerical values
+- Fix[PyPika]: `ifthenelse` step supports numerical values
 
 ### Changed
 
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog (weaverbird python package)
 
+## Unreleased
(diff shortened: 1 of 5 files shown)


fspot (approved):

server/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog (weaverbird python package)
 
+## Unreleased
+
+### Fixed
+
+- Fix[PyPika]: ithenelse step supports numerical values
```
PrettyWood, line 7:
```suggestion
- Fix[PyPika]: `ifthenelse` step supports numerical values
```

PrettyWood (approved):

PR #1873, opened 2023-08-30
fix(server/pypika): Various improvements on the fromdate step TCTC-6156

lukapeschke:
Work items:

* Duplicated all existing fixtures for PyPika translators. Ensure formattings are the same across all
  PyPika backends.

* Postgres: Do not pad Months to 9 chars

* Athena: Set FROM_DATE_OP to an existing function

* GBQ: Implement a FORMAT_DATE function (neither DATE_FORMAT nor TO_CHAR are supported)

* Redshift: Make formatting consistent with other backends

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -1200,6 +1200,8 @@ class SQLTranslator(ABC):
                 convert_fn = DateFormat
             case FromDateOp.TO_CHAR:
                 convert_fn = functions.ToChar
+            case FromDateOp.FORMAT_DATE:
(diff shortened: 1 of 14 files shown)


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1800,6 +1802,12 @@ def __init__(self, term: str | Field, date_format: str, alias: str | None = None
         super().__init__("DATE_FORMAT", term, date_format, alias=alias)
 
 
+# Of course GBQ must have a different name AND inverted arguments
```
PrettyWood, line 1805: google

PrettyWood (approved):

PR #1878, opened 2023-09-05
feat(server/pypika): Add support for the duration step TCTC-6158

lukapeschke:
closes TCTC-6485

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -1,4 +1,4 @@
-from abc import ABC
+from abc import ABC, abstractmethod
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
 from datetime import datetime, timezone
@@ -928,10 +928,9 @@ class SQLTranslator(ABC):
         return StepContext(query, columns + [step.new_column_name])
 
     @classmethod
+    @abstractmethod
     def _interval_to_seconds(cls, value: Selectable) -> functions.Function:
-        raise NotImplementedError(
-            f"[{cls.DIALECT}] Converting an INTERVAL to seconds is not supported"
-        )
+        """Converts an INTERVAL SQL type to a duration in seconds"""
 
     def duration(
         self: Self,
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -49,6 +49,7 @@ from weaverbird.pipeline.dates import RelativeDate
 from weaverbird.pipeline.pipeline import Pipeline
 from weaverbird.pipeline.steps import ConvertStep, DomainStep, FilterStep, TopStep
 from weaverbird.pipeline.steps.date_extract import DATE_INFO
+from weaverbird.pipeline.steps.duration import DURATIONS_IN_SECOND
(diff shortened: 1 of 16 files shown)


lukapeschke:
depends on #1877

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -925,6 +927,30 @@ def duplicate(
...
+        prev_step_table: str,
+        columns: list[str],
+        step: "DurationStep",
+    ) -> StepContext:
+        the_table = Table(prev_step_table)
+        as_seconds = self._interval_to_seconds(
+            self._cast_to_timestamp(the_table[step.end_date_column])
+            - self._cast_to_timestamp(the_table[step.start_date_column]),
+        )
+        new_column = (as_seconds / DURATIONS_IN_SECOND[step.duration_in]).as_(step.new_column_name)
```
davinov, line 949: Can the division output decimal values?
Would be nice to add a test case for that (like: an duration of 1 hour and a half should appear as 1.5 hours)

ui/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 ### Added
+- NativeSQL translator: Added support for the duration step
```
davinov, line 6: Maybe we should precise Postgres, Athena and BigQuery?
If I understand correctly, the other engines will fail during `_interval_to_seconds` because they don't override it.

davinov (changes requested):
Thanks & nice!

Here a just a few notes:

server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -925,6 +927,30 @@ def duplicate(
         )
         return StepContext(query, columns + [step.new_column_name])
 
+    @classmethod
+    def _interval_to_seconds(cls, value: Selectable) -> functions.Function:
+        raise NotImplementedError(
+            f"[{cls.DIALECT}] Converting an INTERVAL to seconds is not supported"
+        )
```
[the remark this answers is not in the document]
lukapeschke: I hesitated because I thought this could be common to all translators with one or two specializations, but they're all too different. So you're right, I'll make it abstract :slightly_smiling_face:

server/src/weaverbird/backends/pypika_translator/translators/base.py
lukapeschke: You're right, I'll see if I can craft such as fixture

server/src/weaverbird/backends/pypika_translator/translators/base.py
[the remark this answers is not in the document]
lukapeschke: [053c872](https://github.com/ToucanToco/weaverbird/pull/1878/commits/053c872a474d211988f4061077d256c6e4f5ab39)

server/src/weaverbird/backends/pypika_translator/translators/base.py
lukapeschke: [84d9c03](https://github.com/ToucanToco/weaverbird/pull/1878/commits/84d9c037045648df588e14983555a47d2867d937)

ui/CHANGELOG.md
davinov: Ah ye! Thanks for the clarification :+1:

davinov (approved):

PR #1881, opened 2023-09-07
feat(ui): Disable customsql step in pypika translator TCTC-6722

lukapeschke:
closes TCTC-6722


davinov (approved):

lukapeschke:
> Actually, don't we use a custom SQL step as first steps for some pipelines? As long as it doesn't break this, it's fine !

Nope, the back-end denies it if it is not the very first step of a pipeline (you can either have a domain step + connectionUid **or** a customsql step) :slightly_smiling_face:

PR #1900, opened 2023-09-26
fix(ui): remove relative date filtering feature flag and related code [TCTC-6755]

Align-center:
## What
- Removing the `RELATIVE_DATE_FILTERING` feature flag that was only used in the `FIlterSimpleCondition` component and was preventing the display of the right operators when having a `dateTime` column selected. 
- Remove related code since this was the only place where the FF was used.

### DIFF
--- a/ui/tests/unit/filter-simple-condition-widget.spec.ts
+++ b/ui/tests/unit/filter-simple-condition-widget.spec.ts
@@ -420,7 +420,7 @@ describe('Widget FilterSimpleCondition', () => {
 
   describe('date column and date (using an invalid operator)', () => {
     let wrapper: Wrapper<FilterSimpleConditionWidget>;
-    const createWrapper = (operator: string, isRelativeDateEnabled = false) => {
+    const createWrapper = (operator: string) => {
       setupMockStore({
         dataset: {
           headers: [{ name: 'columnA' }, { name: 'columnB' }, { name: 'columnC' }],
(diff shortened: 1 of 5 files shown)


ui/tests/unit/filter-simple-condition-widget.spec.ts
```diff
@@ -539,28 +440,8 @@ describe('Widget FilterSimpleCondition', () => {
...
-      createWrapper('until');
-      await wrapper.vm.$nextTick();
-      expect(wrapper.emitted().input).toHaveLength(1);
-      // should emit new value with 'le' operator replacing invalid 'until' operator
-      expect(wrapper.emitted().input[0]).toEqual([
-        { column: 'columnA', value: '', operator: 'le' },
-      ]);
-    });
-
     it('should transform an invalid "le" operator when relative date are enabled', async () => {
```
davinov, line 443: Maybe we can remove the "when relative date are enabled" from test titles ?

davinov (approved):
Awesome !
Just a minor comment :

Thanks for cleaning this 👍👍👍

ui/tests/unit/filter-simple-condition-widget.spec.ts
Align-center: Yes you're right ! Fixed in [44bb611](https://github.com/ToucanToco/weaverbird/pull/1900/commits/44bb61154b74948aef85831742a9b376b2d83711) 🙏

alice-sevin (approved):
Topissime, merci beaucoup pour le cleaning :pray:

PR #1920, opened 2023-10-12
fix(server/waterfall): Make the first milestone the first element of the dataset

lukapeschke:
Rather than having both milestones as the last elements, the first milestone is now the first element,
and the last one is the last element


davinov (approved):
Very clear :+1: Thanks!

PR #1948, opened 2023-11-16
feat(sql): customsql step can contains a column list TCTC-7280

davinov:
Avoid to rely on a `__custom_sql_table__`, which would be overridden when there is multiple customsql steps

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Added
+
+- Pypika translator: added an option `columns` property to CustomSQL step. This columns list is used when translating
+  the query, and avoid relying on a fake table in `table_columns`. Prevents name conflicts when joining multiple
+  pipelines with different CustomSQL steps.
+
 ## [0.38.1] - 2023-10-13
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


server/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Added
+
+- Pypika translator: added an option `columns` property to CustomSQL step. This columns list is used when translating
```
```suggestion
- Pypika translator: added an optional `columns` property to CustomSQL step. This column list is used when translating
```

server/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
...
+  the query, and avoid relying on a fake table in `table_columns`. Prevents name conflicts when joining multiple
```
lukapeschke, line 8:
```suggestion
  the query, and avoids relying on a fake table in `table_columns`. This prevents name conflicts when joining multiple
```

ui/src/components/stepforms/schemas/customsql.ts
lukapeschke, line 1: Doesn't the `CustomSqlStep` type in `ui/src/lib/steps.ts` need to be updated as well ?

lukapeschke (changes requested):
Thanks for taking care of this :pray:

ui/src/components/stepforms/schemas/customsql.ts
davinov: Indeed! Thanks

PR #1978, opened 2023-12-27
feat(steps/filter): Allow using variables in RelativeDate TCTC-7581

lukapeschke:
relates to TCTC-7581

### DIFF
--- a/server/tests/steps/test_filter.py
+++ b/server/tests/steps/test_filter.py
@@ -2,11 +2,13 @@ import json
 from datetime import datetime
 from os.path import dirname
 from os.path import join as path_join
+from typing import Any
 from zoneinfo import ZoneInfo
 
 import pytest
 from pandas import DataFrame, read_json
 from pandas.testing import assert_frame_equal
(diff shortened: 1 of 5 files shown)


server/tests/steps/test_filter.py
```diff
@@ -373,3 +375,55 @@ def test_date_filter(date_df: DataFrame, expected_date_filter_result: DataFrame)
...
+                        },
+                    ]
+                },
+            ]
+        },
+    }
+
+    step = FilterStepWithVariables(**raw)
+    assert step.condition.or_[0].and_[0].value == "<%=appRequesters.date.start%>"
+    assert all(isinstance(cond.value, RelativeDateWithVariables) for cond in step.condition.or_[1].and_)
```
PrettyWood, line 429: can you test with `render` ?

PrettyWood (changes requested):

server/tests/steps/test_filter.py
lukapeschke: 296cbc279

PR #1982, opened 2024-01-06
Remove event listeners before component unmounts

Arooba-git:
Hi! 👋 
The event listeners added in ResizablePanels and StepForm are never removed once added. Even after the component unmounts, the event handlers remain bound to the events. This would lead to an accumulation of such listeners each time a component is mounted, leading to a memory leak. This PR fixes this issue:)


davinov:
Closing as we're gonna rewrite the UI part at some point

PR #2001, opened 2024-01-25
fix(server): accept variables in combined pipelines TCTC-7763

davinov:
In append steps, I noticed that sub-pipelines containing variables were not considered valid.

### DIFF
--- a/server/src/weaverbird/pipeline/steps/utils/combination.py
+++ b/server/src/weaverbird/pipeline/steps/utils/combination.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, Annotated, Literal
 from pydantic import BaseModel, BeforeValidator, TypeAdapter
 
 if TYPE_CHECKING:
-    from weaverbird.pipeline.pipeline import PipelineStep, PipelineStepWithRefs
+    from weaverbird.pipeline.pipeline import PipelineStep, PipelineStepWithRefs, PipelineStepWithVariables
 
 
 class Reference(BaseModel):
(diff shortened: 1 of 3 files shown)


server/src/weaverbird/pipeline/steps/utils/combination.py
```diff
@@ -55,10 +55,12 @@ def _pipelinestep_adapter() -> TypeAdapter["str | list[PipelineStep]"]:
...
 @cache
-def _pipelinestepwithref_adapter() -> TypeAdapter["str | list[PipelineStepWithRefs | PipelineStep]"]:
-    from weaverbird.pipeline.pipeline import PipelineStep, PipelineStepWithRefs
+def _pipelinestepwithref_adapter() -> (
+    TypeAdapter["str | list[PipelineStepWithRefs | PipelineStepWithVariables | PipelineStep]"]
+):
+    from weaverbird.pipeline.pipeline import PipelineStep, PipelineStepWithRefs, PipelineStepWithVariables
 
-    return TypeAdapter(str | list[PipelineStepWithRefs | PipelineStep])  # type: ignore[arg-type]
+    return TypeAdapter(str | list[PipelineStepWithRefs | PipelineStep | PipelineStepWithVariables])  # type: ignore[arg-type]
```
lukapeschke, line 63: Could you please add a comment indicating that `PipelineStepWithVariables` must be last, otherwise we might end up with `WithVariables` variants rather than regular `PipelineStep`s ?

davinov: Ah yes ideed !

davinov: What do you think of:
```py
# Note: PipelineStep must appear _before_ PipelineStepWithVariables, to avoid fields that could contain variables
# to always be matched as strings
```

lukapeschke (approved):

PR #2037, opened 2024-03-11
fix(date-inputs): use correct variables delimiters for default variables [TCTC-8076]

alice-sevin:
In a previous PR: https://github.com/ToucanToco/weaverbird/pull/1988, we updated the delimiters for relative variables but we forgot to do it for default variables (TODAY, YESTERDAY), picked in variables list not in relative date form.
We add logic to retrieve correct delimiters depending on if the variable is trusted or not.

### DIFF
--- a/ui/CHANGELOG.md
+++ b/ui/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+### Fixed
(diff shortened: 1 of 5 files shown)


davinov (approved):
:+1: thx!

PR #2041, opened 2024-03-12
fix(pypika translator): cannot select null in filterstep with operators in / nin / eq / ne [TCTC-7943]

fspot:
### in / nin

```sql
WHERE column IN (NULL, "foo", "bar")
```
does not work. Replace it with

```sql
WHERE column IS NULL OR column IN ("foo", "bar")
```

### eq / ne

```sql
WHERE column=NULL
```
does not work. Replace it with

```sql
WHERE column IS NULL
```

### DIFF
--- a/server/src/weaverbird/pipeline/steps/filter.py
+++ b/server/src/weaverbird/pipeline/steps/filter.py
@@ -1,4 +1,4 @@
-from typing import Literal
+from typing import Any, Literal
 
 from weaverbird.pipeline.steps.utils.base import BaseStep
 from weaverbird.pipeline.steps.utils.render_variables import StepWithVariablesMixin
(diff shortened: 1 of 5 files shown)


server/src/weaverbird/pipeline/steps/filter.py
```diff
@@ -13,3 +13,8 @@ class FilterStep(BaseStep):
 
 class FilterStepWithVariables(FilterStep, StepWithVariablesMixin):
     condition: ConditionWithVariables  # type:ignore[assignment]
+
+    def model_dump(self, *, exclude_none: bool = False, **kwargs) -> dict[str, Any]:
+        # we don't want to drop fields with value=None like:
+        #     {'column': 'nullable_name', 'operator': 'eq', 'value': None}
+        return super().model_dump(exclude_none=exclude_none, **kwargs)
```
lukapeschke, line 20: isn't `exclude_none=False` already the default behavior ? https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump

lukapeschke (approved):
Just a note on the new `model_dump` method, LGTM apart from that :+1:

server/src/weaverbird/pipeline/steps/filter.py
fspot: we have it overriden with exclude_none=True in BaseStep

lukapeschke: :bulb:

PR #2059, opened 2024-04-03
fix(steps): remove custom and custom sql deprecated steps from action menu [TCTC-7178]

alice-sevin:
We remove the deprecated 'Custom sql' and 'Custom' steps from action menu.
It won't be usable anymore with hybrid pipeline

:information_source: We don't remove steps code and pipeline listing because we may have migrated apps with custom steps so we want them to be removed before cleaning the code
The customs steps are still accessible (to be deleted) if a step has been created earlier but app builder won't be able to create new ones

### DIFF
--- a/ui/src/components/constants.ts
+++ b/ui/src/components/constants.ts
@@ -36,8 +36,6 @@ export const ACTION_CATEGORIES: ActionCategories = {
     { name: 'text', label: 'Add text column' },
     { name: 'formula', label: 'Add formula column' },
     { name: 'ifthenelse', label: 'Add conditional column' },
-    { name: 'custom', label: 'Add custom step' },
-    { name: 'customsql', label: 'Add custom sql step' },
   ],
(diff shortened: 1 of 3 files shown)


ui/src/components/constants.ts
```diff
@@ -36,8 +36,6 @@ export const ACTION_CATEGORIES: ActionCategories = {
     { name: 'text', label: 'Add text column' },
     { name: 'formula', label: 'Add formula column' },
     { name: 'ifthenelse', label: 'Add conditional column' },
-    { name: 'custom', label: 'Add custom step' },
-    { name: 'customsql', label: 'Add custom sql step' },
```
davinov, line 40: Does that also hides them from search?

alice-sevin: Ils n'étaient déjà pas visibles

davinov (approved):

PR #2072, opened 2024-04-16
fix(server): correctly generate one parent and one child row even if labels are the same TCTC-8331 

davinov:
Mongo and pandas waterfall steps had slightly different behaviors when handling rows that have the same value for group and label.
Fixes this difference in favor of producing one parent row and one child row.


lukapeschke (approved):
LGTM, could you please add an entry to the changelog as well ?

PR #2088, opened 2024-04-29
chore(server/pypika): removed join and append steps

lukapeschke:


davinov (changes requested):
If we combine 2 datasets coming from the same db, it's way better to delegate this join to this db, isn't it?
Wouldn't that PR prevent this feature, and force a retrieval in memory of both tables?

PR #2097, opened 2024-05-06
fix(server/pypika): fix _id method

lukapeschke:
The _id method appears to be flaky: it relies on the id() built-in, which must return a unique identifier for an object. In CPython, this is the object's address.

This causes an issue in the `append` method of the PyPika translator when several pipelines are appended: an instance of the translator class is created everytime we need to process a pipeline. We assume the id of every instance to be unique to generate unique human-readable step names in CTEs.

However, when multiple pipelines are appended, an instance might be garbage-collected between two iterations of the for-loop, which can resulting in the next instance (or the one after that) to be allocated in the address as the garbage-collected one. This results in both instances having the same id, and thus generating the same step names.

This issue is fixed by generating a unique id in the classe's __init__ method, and rely on this id rather than the object's pythonic id when generating step names.

### DIFF
--- a/server/src/weaverbird/backends/pypika_translator/translators/base.py
+++ b/server/src/weaverbird/backends/pypika_translator/translators/base.py
@@ -2,8 +2,8 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
 from datetime import UTC, datetime
-from functools import cache
 from typing import TYPE_CHECKING, Any, Literal, TypeVar, Union, cast, get_args
+from uuid import uuid4
 
 from dateutil import parser as dateutil_parser
 from pypika import (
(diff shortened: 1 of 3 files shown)


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -207,20 +207,20 @@ def __init__(
         self._known_instances: dict[int, str] = known_instances or {}
         self._step_count = 0
         self._source_rows_subset = source_rows_subset
+        self.__id = uuid4().int
 
     def __init_subclass__(cls) -> None:
         ALL_TRANSLATORS[cls.DIALECT] = cls
 
-    @cache  # noqa: B019
```
davinov, line 214: Why cas this cached? it doesn't see very costly ^^

davinov (approved):

server/src/weaverbird/backends/pypika_translator/translators/base.py
lukapeschke: no idea :man_shrugging:

PR #2107, opened 2024-05-15
fix(formula): handle single quotes in formula and conditional steps [TCTC-8504]

fspot:
Single quotes were not handled properly in ast parser, resulting in formula and conditional steps not working properly.

### DIFF
--- a/server/src/weaverbird/pipeline/formula_ast/types.py
+++ b/server/src/weaverbird/pipeline/formula_ast/types.py
@@ -3,8 +3,6 @@ from typing import Union
 
 from pydantic import BaseModel
 
-from weaverbird.pipeline.formula_ast.utils import unquote_string
-
 
 class Operator(str, Enum):
     ADD = "+"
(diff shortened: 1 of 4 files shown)


server/src/weaverbird/pipeline/formula_ast/types.py
```diff
@@ -33,41 +31,3 @@ class Operation(BaseModel):
...
-    elif isinstance(expr, ColumnName):
-        return f"{column_start_seq}{expr.name}{column_end_seq}"
-    elif isinstance(expr, str):
-        # Unquoting
-        return f"{str_quote_seq}{unquote_string(expr)}{str_quote_seq}"
-    # bool is a subtype of int, but we don't want to have 1 and 0 replaced with 'true' or 'false'
-    elif isinstance(expr, int | float) and not isinstance(expr, bool):
-        return expr
-    elif isinstance(expr, bool):
-        return expr if bools_as_py else ("true" if expr else "false")
```
lukapeschke, line 73: c'était pour debugger et mettre au propre des formulas ça, on peut éventuellement le garder avec un commentaire ?

lukapeschke (approved):
:ok_hand:

server/tests/backends/fixtures/ifthenelse/quotes.json
```diff
@@ -17,7 +17,7 @@
           "operator": "gt"
         },
         "newColumn": "COND",
-        "then": "'tintin'",
+        "then": "\"tin'tin\"",
         "else": "some ' stuff with \" annoying 'quotes'"
```
davinov, line 21: Shouldn't this else example be invalid? (there is column `some` nor `stuff`, etc...)

davinov (changes requested):
Thanks :+1: 

We also may be a bit laxist with some formulas. Can we prevent invalid ones? (invalid = identifers (tokens without quotes) that doesn't refer to a column)

server/src/weaverbird/pipeline/formula_ast/types.py
fspot: c'est du code mort

server/tests/backends/fixtures/ifthenelse/quotes.json
PrettyWood: 100% aligné. IMO il faut qu'on rende bcp plus robuste notre formula step. Et plus tôt on le fait sur la v3 mieux on se portera

PR #2177, opened 2024-07-17
fix(server/pandas): reindex rows at the end of rollup step (TCTC-8689)

julien-pinchelimouroux:
## Why

 - Can raise  `*** ValueError: cannot reindex on an axis with duplicate labels` if the next step assigns a new column

### DIFF
--- a/server/CHANGELOG.md
+++ b/server/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


davinov (approved):
OK! I guess that it has potential perf implications (could be a bit long on a big dataset?), but I don't see how we could avoid it.
Maybe if there is a method to determine if the index has duplicates?

Otherwise, looks good. Thanks!

julien-pinchelimouroux:
@davinov If the rollup has multiple columns configured, it will always creates duplicates indexes
We can at least check on this condition before apply the index re-computation

PR #2217, opened 2024-08-27
fix(server/pypika): support datetimes in filter step with Athena [TCTC-9146]

lukapeschke:


julien-pinchelimouroux (approved):
ok pour moi :+1:

PR #2223, opened 2024-09-03
fix: mongo left outer join runs same logic than left join (TCTC-8310)

julien-pinchelimouroux:


server/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
```
lukapeschke, line 5:
```suggestion
### Changed
```

I don't think it will impact people, but it technically is a breaking change

server/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+
+- Mongo: `left` and `left outer` join step return the same translation
```
```suggestion
- Mongo: The `join` step now has the same behavior for the `left` and `left outer` join methods
```

lukapeschke (approved):

PR #2230, opened 2024-09-06
fix(server): prevent custom SQL queries finishing with comments to fail

davinov:
As noticed by @ThGa-Droid , custom queries that ended with comments failed.
That's because we are encapsulating
 them like this:
```sql
WITH __step_0_googlebigquerytranslator__ AS (SELECT
  *
FROM `something`
  -- some comment) SELECT * FROM `__step_0_googlebigquerytranslator__` LIMIT 50
```
which cases the last line to be ignored.

To avoid this, I suggest we add a new line to all custom queries, like:
```sql
WITH __step_0_googlebigquerytranslator__ AS (SELECT
  *
FROM `something`
  -- some comment
) SELECT * FROM `__step_0_googlebigquerytranslator__` LIMIT 50
```


lukapeschke (approved):
Good catch :ok_hand:

fspot (approved):

PR #2234, opened 2024-09-10
fix(ui): allow typing column name where it was forgotten TCTC-9216

davinov:
In #2229 , I forgot some widgets where we also need to be able to type column names

### DIFF
--- a/ui/src/components/stepforms/HierarchyStepForm.vue
+++ b/ui/src/components/stepforms/HierarchyStepForm.vue
@@ -22,6 +22,7 @@
       :automatic-new-field="false"
       data-path=".hierarchy"
       :errors="errors"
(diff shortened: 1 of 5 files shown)


ui/src/components/stepforms/HierarchyStepForm.vue
```diff
@@ -22,6 +22,7 @@
       :automatic-new-field="false"
       data-path=".hierarchy"
       :errors="errors"
+      :componentProps="{ allowCustom: true }"
```
alice-sevin, line 25: Tiens c'est drole, pourquoi on a pas le même naming ici? C'est un composant dynamique?

alice-sevin (approved):

PR #2240, opened 2024-09-17
fix: ifthenelse step must return an unique list of column names (TCTC-9145)

julien-pinchelimouroux:


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1468,7 +1468,7 @@ def ifthenelse(
                 case_=Case(),
             ).as_(step.new_column),
         )
-        return StepContext(query, columns + [step.new_column])
+        return StepContext(query, list(set(columns + [step.new_column])))
```
lukapeschke, line 1471: LGTM but could we avoid the `set` usage to preserve column ordering ? Something like:
```suggestion
        return StepContext(query, columns + [step.new_column] if step.new_column not in columns else columns)
```

lukapeschke (approved):
LGTM, thanks

PR #2248, opened 2024-09-23
chore(ui): remove store dependency for step form [TCTC-9179]

alice-sevin:
1. Export some constants and getters/actions to be able to move dataviewer component to another repository
2. Remove store dependency from stepform component to be able to use it with query builder or without it.

Now stepform can be used as a dumb component using props or plug to store

### DIFF
--- a/ui/src/components/stepforms/StepFormComponent.vue
+++ b/ui/src/components/stepforms/StepFormComponent.vue
@@ -1,0 +1,116 @@
+<template>
+  <component
+    key="stepForm"
+    :is="formComponent"
+    ref="step"
+    :translator="translator"
+    :initialStepValue="initialStepValue"
+    :isStepCreation="isStepCreation"
+    :columnTypes="columnTypes"
+    :backendError="backendError"
+    :availableVariables="availableVariables"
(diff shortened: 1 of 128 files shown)


ui/src/components/stepforms/StepFormComponent.vue
davinov, line 1: :+1:

ui/stories/StepForm.stories.ts
davinov, line 1: Thanks for including the stories :+1:

ui/src/main.ts
```diff
@@ -33,6 +44,7 @@ export {
   NewDateInput,
   DateRangeInput,
   Pagination,
+  StepFormComponent,
```
davinov, line 47: Maybe this could be called `StepForm` only?

davinov (approved):
Awesome! I've not read in details all the modifications to each step form, but the API of the StepForm component is clear. Thanks for this work :clap:

ui/src/main.ts
alice-sevin: There is already a StepForm component (it's the base form component used by each steps :wink: )

PR #2251, opened 2024-09-27
fix(step-form-component): provide step form defaults and preselect column if necessary in complex steps [TCTC-9437]

alice-sevin:
Form was not working as expected with defaults values (partial config), due to missing props propagation
We also need to add a selectedColumn property because some complex steps (filter, rename...) preselect column not based on defaults but on selectedColumns logic


davinov (approved):
:ok_hand:

PR #2256, opened 2024-10-02
fix(server): cast filler value in fillna step [TCTC-9419]

lukapeschke:


lukapeschke:
fixed by https://github.com/ToucanToco/toucan-connectors/pull/1784

PR #2261, opened 2024-10-09
fix(step-form-component): handle reactivity for selected column

alice-sevin:
I missed reactivity for selected column in weaverbird data viewer migration, it results to selected column not working as expected in migrated code

Watch update state value
Key update form component (because it not watch anything, it only apply state on init based on selected column so we need a full refresh)


davinov (approved):
Vu en visio

PR #2296, opened 2024-12-05
fix(server/pypika): reverse OFFSET and LIMIT for Athena translator [TCTC-9905]

lukapeschke:
…CTC-9905]


julien-pinchelimouroux (approved):
:ok_hand:

PR #2300, opened 2024-12-10
fix: fix pypika replace step for substrings (TCTC-9903)

julien-pinchelimouroux:


server/src/weaverbird/backends/pypika_translator/translators/base.py
```diff
@@ -1681,15 +1681,16 @@ def replace(
...
-
         # Do a nested `replace` to replace many values on the same column
         replaced_col = col_field
+
+        case_ = Case()
         for old_name, new_name in step.to_replace:
-            replaced_col = functions.Replace(replaced_col, old_name, new_name)
+            case_ = case_.when(col_field == old_name, functions.Replace(replaced_col, old_name, new_name)).else_(
+                col_field
+            )
```
lukapeschke, line 1692:
```suggestion
            case_ = case_.when(col_field == old_name, LIteravalValue(new_name)).else_(col_field)
```

We don't want to replace text here

server/tests/backends/fixtures/replace/with_variables.yaml
```diff
@@ -0,0 +1,115 @@
+---
+exclude:
+- mongo
+- pandas
+- snowflake
+step:
+  pipeline:
```
lukapeschke, line 7: Could you please add a `select` step to have case to have only the columns we're interested in ? Could you also add a replace case on "iple" to ensure "Triple" does not get modified ?

lukapeschke (changes requested):
thanks!

server/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Pypika: `replace` doesn't replace substring value even if it is matching the value to replace.
```
Fanaen, line 7: I was confused about this, so here's another suggestion. I'm not sure its better 😄 
```suggestion
- Pypika: `replace` doesn't do partial match anymore, as intended.
```

Fanaen (approved):
Thanks 👍

server/src/weaverbird/backends/pypika_translator/translators/base.py
Fanaen: I'm not sure about `LIteraval` 😄 
But true, I missed that!

server/tests/backends/fixtures/replace/with_variables.yaml
julien-pinchelimouroux: `Ale` case is already here for that ;)

server/CHANGELOG.md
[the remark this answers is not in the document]
julien-pinchelimouroux: you don't like my english style :surfer: ?

server/tests/backends/fixtures/replace/with_variables.yaml
lukapeschke: Not really, since "India Pale Ale"  is a superset of the match (both contain "Ale"). With the previous behaviour, results could vary depending on the order of the replaces :man_shrugging: To completely validate that this is not the case anymore, we need a completely different match

server/CHANGELOG.md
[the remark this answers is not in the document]
Fanaen: No, that's a meaning thing. "La step replace ne remplace pas la partie de texte même si cela correspond à ce qu'il faut remplacer" is a pretty convoluted way of saying "On veut juste une correspondance complète"

server/tests/backends/fixtures/replace/with_variables.yaml
julien-pinchelimouroux: There is a second kind which contains Ale : 
`beer_kind: Belge blonde forte & Golden Ale`
This one isn't replaced

lukapeschke: ah true, nevermind

server/CHANGELOG.md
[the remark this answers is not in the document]
julien-pinchelimouroux: That's a joke :smiley:

lukapeschke (approved):

PR #2337, opened 2025-01-29
chore(server): v0.49.0

lukapeschke:


julien-pinchelimouroux (approved):
Tu peux faire un `uv sync` histoire d'avoir la bonne version dans le uv.lock aussi ? 
Merci!

lukapeschke:
weird, `uv sync` did not update the lockfile, I had to manually remove the version from the lockfile for it to work

lukapeschke:
hmm updating uv from 0.5.20 -> 0.5.25 did the trick

julien-pinchelimouroux:
I'm in 0.5.24, it also works

PR #2341, opened 2025-02-06
fix: dont remove inclusion condition with an empty list of values (TCTC-10114)

julien-pinchelimouroux:


server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -257,10 +257,12 @@ def _remove_void_from_condition(condition: Condition) -> Condition | None:
         if condition.column == VOID_REPR or condition.value == VOID_REPR:
             return None
     elif isinstance(condition, InclusionCondition):
-        condition_values = [v for v in condition.value if v != VOID_REPR]
-        if len(condition_values) == 0:
-            return None
-        condition.value = condition_values
+        # We dont want to remove the condition if the list of values is empty
```
Fanaen, line 260: **suggestion(non-blocking):** it may be useful to re-explain what cases means. What do you think of this?
```suggestion
        # [] aka *empty list* means we want an empty list.
        # [__VOID__] means we don't want to filter.
        # If there are other values in the list, we only want to ignore those __VOID__
```

Fanaen (approved):
👍

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -257,10 +257,12 @@ def _remove_void_from_condition(condition: Condition) -> Condition | None:
...
+        if len(condition.value) > 0:
+            condition_values = [v for v in condition.value if v != VOID_REPR]
+            if len(condition_values) == 0:
+                return None
```
lukapeschke, line 264: I'm not sure we still want this... Should check if it happens in the front-end

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -257,10 +257,12 @@ def _remove_void_from_condition(condition: Condition) -> Condition | None:
...
+            condition.value = condition_values
```
lukapeschke, line 265:
```suggestion
            return condition.model_copy(updates={"value": condition_values})
```

lukapeschke (changes requested):
thx

server/src/weaverbird/pipeline/pipeline.py
julien-pinchelimouroux: > In other words, we don't want to
> remove an empty object that was initially there but if, by removing `__VOID__`
> entries, the object becomes empty, we do want to remove it completely.

This comes from the front-end (tucana)

I think we can keep that behaviour

server/src/weaverbird/pipeline/pipeline.py
julien-pinchelimouroux: [21f60bd](https://github.com/ToucanToco/weaverbird/pull/2341/commits/21f60bd8de258d5f0b8ccf198775b7dc0fb1d868)

Fanaen (approved):

server/src/weaverbird/pipeline/pipeline.py
```diff
@@ -257,11 +257,13 @@ def _remove_void_from_condition(condition: Condition) -> Condition | None:
         if condition.column == VOID_REPR or condition.value == VOID_REPR:
             return None
     elif isinstance(condition, InclusionCondition):
-        condition_values = [v for v in condition.value if v != VOID_REPR]
-        if len(condition_values) == 0:
+        # "__VOID__" means we don't want to filter
+        # [] aka *empty list* or ["__VOID__"] means we want an empty list.
+        # If there are other values in the list, we only want to ignore those "__VOID__"
```
lukapeschke, line 262: Isn't the function's name self-explanatory ? the comment is confusing me :sweat_smile:

lukapeschke (approved):
thx

server/src/weaverbird/pipeline/pipeline.py
Fanaen: You're right, it is now! When I suggested this comment, it did not entirely match the name. Now it does, so we can remove the comment if it's confusing.

PR #2370, opened 2025-03-19
chore(server): v0.52.0

julien-pinchelimouroux:


server/uv.lock
```diff
@@ -1,5 +1,4 @@
 version = 1
-revision = 1
```
lukapeschke, line 2: what uv version do you have ? Seems weird that stuff gets removed here

julien-pinchelimouroux: oh you're probably right : `0.5.24`

lukapeschke: Could you please `uv self update` and bump `UV_VERSION` in `.github/` as well ?

lukapeschke (approved):

PR #2402, opened 2025-04-14
feat(aggregate): add a new countNulls parameter [TCTC-10573]

lukapeschke:


server/tests/backends/fixtures/aggregate/count_nulls_false_pypika.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - nullable_name: Brugse Zot blonde
+      nullable_name_count: 1
+    - nullable_name: Ninkasi Ploploplop
+      nullable_name_count: 1
+    - nullable_name: Pauwel Kwak
+      nullable_name_count: 1
+    - nullable_name: Weihenstephan Hefe Weizen Alcoholarm
+      nullable_name_count: 1
+    - nullable_name: null
+      nullable_name_count: 0
```
Fanaen, line 34: We keep this to make sure we don't break any existing pipeline, right?

Fanaen (approved):
Thanks! 👍

server/tests/backends/fixtures/aggregate/count_nulls_false_pypika.yaml
lukapeschke: yup

PR #2418, opened 2025-04-25
feat(server, dategranularity): support variables in `granularity` field

Fanaen:


lukapeschke (changes requested):
needs to be added to `PipelineStepWithVariable`

Fanaen:
@lukapeschke As I thought, [they already are](https://github.com/ToucanToco/weaverbird/blob/e8cea0c33dfa1ef0098abbeeaa577a8269d61544/server/src/weaverbird/pipeline/pipeline.py#L205). The problem was I failed to properly build a new test version.

lukapeschke (approved):
:bulb:

PR #2421, opened 2025-04-29
fix(mongo, todate): support dates with two-digit years [TCTC-10523]

Fanaen:


Fanaen:
Right! Thanks for the code snippet 🙏 
The test is now more extensive to make sure to get it right (and in `yaml`, since we're at it).

server/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+- Mongo: step `todate` supports two-digit years (`%y`) like pandas.
```
```suggestion
- Mongo: step `todate` properly supports two-digit years (`%y`).
```

server/src/weaverbird/backends/mongo_translator/steps/todate.py
```diff
@@ -7,13 +7,20 @@ def translate_todate(step: ToDateStep) -> list[MongoStep]:
 
     col_as_string = {"$toString": f"${step.column}"}
 
+    steps: list[dict] = []
+    should_guess_year = False
```
lukapeschke, line 11:
```suggestion
    should_guess_century = False
```

We're guessing the century, right ?

lukapeschke (approved):
Looks good, thanks!

PR #2422, opened 2025-04-30
feat(ifthenelse, formula): allow `null` as value

Fanaen:
In some `ifthenelse` cases, it is preferable to simply output `null` in the `then` or `else`.
It is indeed a breaking change.

### DIFF
--- a/server/src/weaverbird/backends/pandas_executor/steps/ifthenelse.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/ifthenelse.py
@@ -16,7 +16,7 @@ def _execute_ifthenelse(ifthenelse: IfThenElse, df: DataFrame, new_column) -> Da
 
     then_branch = eval_formula(df, str(ifthenelse.then))
 
-    # If one of the branch is filled with null, we cast it to the non-null type to avoid date issues
+    # If one of the branches is null, we cast it to the non-null type to avoid typing issues
     else_branch_is_null = isinstance(else_branch, NullSeries)
     then_branch_is_null = isinstance(then_branch, NullSeries)
 
--- a/server/src/weaverbird/backends/pandas_executor/steps/ifthenelse.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/ifthenelse.py
@@ -1,7 +1,7 @@
 import numpy as np
 from pandas import DataFrame
 
-from weaverbird.backends.pandas_executor.steps.utils.formula import eval_formula
+from weaverbird.backends.pandas_executor.steps.utils.formula import NullSeries, eval_formula
 from weaverbird.backends.pandas_executor.types import DomainRetriever, PipelineExecutor
 from weaverbird.pipeline.steps.ifthenelse import IfThenElse, IfthenelseStep
 
@@ -15,6 +15,16 @@ def _execute_ifthenelse(ifthenelse: IfThenElse, df: DataFrame, new_column) -> Da
         else_branch = eval_formula(df, str(ifthenelse.else_value))
 
     then_branch = eval_formula(df, str(ifthenelse.then))
+
+    # If one of the branch is filled with null, we cast it to the non-null type to avoid date issues
+    else_branch_is_null = isinstance(else_branch, NullSeries)
+    then_branch_is_null = isinstance(then_branch, NullSeries)
+
+    if else_branch_is_null and not then_branch_is_null:
+        else_branch = else_branch.astype(then_branch.dtype)
+    elif then_branch_is_null and not else_branch_is_null:
+        then_branch = then_branch.astype(else_branch.dtype)
+
     return df.assign(**{new_column: np.where(apply_condition(ifthenelse.condition, df), then_branch, else_branch)})
 
 
--- a/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
@@ -16,7 +16,10 @@ _OP_MAP = {
 
 
 class NullSeries(Series):
-    """This type allow us to know that a series is filled with None without re-checking all its content."""
+    """A Series containing only null values.
+    
+    This type allows us to know that a Series is filled with null values without having to check its content
+    """
 
     def __init__(self, index: DataFrame.index):
         Series.__init__(self, index=index, dtype="object")
--- a/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
@@ -15,6 +15,16 @@ _OP_MAP = {
 }
 
 
+class NullSeries(Series):
+    """
+    When the expr is "null", this is a float series filled with NaN.
+    This subclass allow us to keep this information for necessary casts, e.g in ifthenelse step.
+    """
+
+    def __init__(self, index: DataFrame.index):
+        Series.__init__(self, index=index, dtype="object")
+
+
 def _eval_operation(df: DataFrame, op: Operation) -> Series:
     return _OP_MAP[op.operator](_eval_expression(df, op.left), _eval_expression(df, op.right)).replace(
         [np.inf, -np.inf], np.nan
@@ -22,7 +32,9 @@ def _eval_operation(df: DataFrame, op: Operation) -> Series:
 
 
 def _eval_expression(df: DataFrame, expr: Expression) -> Series:
-    if isinstance(expr, Operation):
+    if expr is None:
+        return NullSeries(df.index)
+    elif isinstance(expr, Operation):
         return _eval_operation(df, expr)
     elif isinstance(expr, ColumnName):
         return df[expr.name]
--- a/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
@@ -17,7 +17,7 @@ _OP_MAP = {
 
 class NullSeries(Series):
     """A Series containing only null values.
-    
+
     This type allows us to know that a Series is filled with null values without having to check its content
     """
 
--- a/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
+++ b/server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
@@ -16,10 +16,7 @@ _OP_MAP = {
 
 
 class NullSeries(Series):
-    """
-    When the expr is "null", this is a float series filled with NaN.
-    This subclass allow us to keep this information for necessary casts, e.g in ifthenelse step.
-    """
+    """This type allow us to know that a series is filled with None without re-checking all its content."""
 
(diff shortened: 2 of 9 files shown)


server/src/weaverbird/pipeline/formula_ast/eval.py
```diff
@@ -153,13 +153,13 @@ def _parse_expr(self, expr: ast.expr) -> types.Expression:
...
             case ast.Constant(value=value):
-                if isinstance(value, bool | int | float | str):
+                if isinstance(value, bool | int | float | str | None):
                     return value
                 else:
                     raise UnsupportedConstant(f"Unsupported constant '{expr}' of type {type(value)}")
             # Column name
             case ast.Name(id=name):
-                return self._build_name(name)
+                return None if name == "null" else self._build_name(name)
```
lukapeschke, line 162: Could you add a case to `_build_name` instead ? So we can have the branches for `null` values and booleans in the same place

server/tests/steps/test_ifthenelse.py
```diff
@@ -67,6 +67,68 @@ def test_then_should_support_formulas():
     assert_dataframes_equals(result_df, expected_df)
 
 
+def test_simple_condition_with_null():
```
julien-pinchelimouroux, line 70: Can you use yaml fixtures instead ? 
This allows to test a step config on each kind of translators (sql, pandas, mongo)

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
```diff
@@ -15,14 +15,26 @@
...
 
 
+class NullSeries(Series):
+    """
+    When the expr is "null", this is a float series filled with NaN.
+    This subclass allow us to keep this information for necessary casts, e.g in ifthenelse step.
+    """
+
+    def __init__(self, index: DataFrame.index):
+        Series.__init__(self, index=index, dtype="object")
```
lukapeschke, line 25:
```suggestion
        super().__init__(self, index=index, dtype="object")
```

Why not an `object` Series filled with `None` ? You're already using the `object` dtype. I don't see the advantage of a float series here

server/src/weaverbird/pipeline/formula_ast/eval.py
julien-pinchelimouroux, line 162: `"null"` can be a constant like `__VOID__`
its easier to retrieve usages later ;)

server/src/weaverbird/backends/pandas_executor/steps/ifthenelse.py
```diff
@@ -15,6 +15,16 @@ def _execute_ifthenelse(ifthenelse: IfThenElse, df: DataFrame, new_column) -> Da
         else_branch = eval_formula(df, str(ifthenelse.else_value))
 
     then_branch = eval_formula(df, str(ifthenelse.then))
+
+    # If one of the branch is filled with null, we cast it to the non-null type to avoid date issues
```
lukapeschke, line 19:
```suggestion
    # If one of the branches is null, we cast it to the non-null type to avoid typing issues
```

julien-pinchelimouroux (commented):
Thanks !

server/tests/steps/test_ifthenelse.py
```diff
@@ -67,6 +67,68 @@ def test_then_should_support_formulas():
...
+        }
+    )
+    result_df = execute_ifthenelse(step, sample_df)
+
+    expected_df = DataFrame(
+        {"a_str": ["test", "test", "autre chose"], "a_date": dates, "test": [dates[0], dates[1], NA]}
+    )
+
+    print(result_df.dtypes)
+    print(expected_df.dtypes)
```
lukapeschke, line 107: to remove

server/tests/steps/test_ifthenelse.py
```diff
@@ -67,6 +67,68 @@ def test_then_should_support_formulas():
...
+            "then": "null",
+            "else": "[a_date]",
+        }
+    )
+    result_df = execute_ifthenelse(step, sample_df)
+
+    expected_df = DataFrame({"a_str": ["test", "test", "autre chose"], "a_date": dates, "test": [NA, NA, dates[2]]})
+
+    print(result_df.dtypes)
+    print(expected_df.dtypes)
```
lukapeschke, line 128: to remove

server/tests/steps/test_ifthenelse.py
lukapeschke, line 1: Could you please add yaml fixtures to target mongo and pypika as well ?

lukapeschke (changes requested):
Thanks!

server/tests/steps/test_ifthenelse.py
Fanaen: Oh, I thought it was the case, given it's not in a pandas-specific path. Damn.

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
Fanaen: Ah, my bad, I forgot the `None`. 
What do you mean by `a float series`?

lukapeschke: it's what you said in the docstring:
> When the expr is "null", this is a float series filled with NaN.

Fanaen: Oh, good catch! Docstring is lying 😄 

In old pandas version, building a Series with None create a float series by default, with a warning to tell us dtype="object" is now the way to go. 

This is exactly what I did, and it even fixed a cast issue. 

Fixed the docstring in [2496b3e](https://github.com/ToucanToco/weaverbird/pull/2422/commits/2496b3e1ad052aadd2b27068f786c21a90b355a4).

---

The suggestion about `super()` gives this error: 'NullSeries' object has no attribute '_name'
I'm not proficient enough in python to know what is going on.

server/tests/steps/test_ifthenelse.py
Fanaen: Removed the old test and created a yaml fixture instead: [ceb2f9f](https://github.com/ToucanToco/weaverbird/pull/2422/commits/ceb2f9f783c6ff81c989ae4dfd4dc4f4b6d6ee4b)

server/tests/steps/test_ifthenelse.py
Fanaen: Removed the old test and created a yaml fixture instead: [ceb2f9f](https://github.com/ToucanToco/weaverbird/pull/2422/commits/ceb2f9f783c6ff81c989ae4dfd4dc4f4b6d6ee4b)

server/tests/steps/test_ifthenelse.py
Fanaen: Removed the old test and created a yaml fixture instead: [ceb2f9f](https://github.com/ToucanToco/weaverbird/pull/2422/commits/ceb2f9f783c6ff81c989ae4dfd4dc4f4b6d6ee4b)

server/tests/steps/test_ifthenelse.py
Fanaen: Good idea: [f15cef1](https://github.com/ToucanToco/weaverbird/pull/2422/commits/f15cef12b73f488a538cb17c91e7f217da6146cd)

server/src/weaverbird/pipeline/formula_ast/eval.py
Fanaen: Good idea! [f15cef1](https://github.com/ToucanToco/weaverbird/pull/2422/commits/f15cef12b73f488a538cb17c91e7f217da6146cd)

server/src/weaverbird/pipeline/formula_ast/eval.py
```diff
@@ -115,16 +115,18 @@ def _operator_from_ast_op(op: ast.operator) -> types.Operator:
             case _:
                 raise UnsupportedOperator(f"Unsupported operator: {op}")
 
-    def _build_name(self, name: str) -> types.ColumnName | bool:
+    def _build_name(self, name: str) -> types.ColumnName | bool | None:
         """Builds a ColumnName from a raw name.
 
         In case the name is "true" or "false", returns the equivalent boolean
-        (users do not expect to have to capitalize booleans).
+        (users do not expect to have to capitalize booleans). Same for "null".
```
lukapeschke, line 122:
```suggestion
        (users do not expect to have to capitalize booleans). 
        In case the name is "null", returns None.
```

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
```diff
@@ -15,14 +15,23 @@
 }
 
 
+class NullSeries(Series):
+    """This type allow us to know that a series is filled with None without re-checking all its content."""
```
lukapeschke, line 19:
```suggestion
    """A Series containing only null values.
    
    This type allows us to know that a Series is filled with null values without having to check its content
    """
```

https://peps.python.org/pep-0257/

server/src/weaverbird/backends/pandas_executor/steps/utils/formula.py
lukapeschke: Ah, `Series` probably has a metaclass. You can leave it as it is

server/tests/backends/fixtures/ifthenelse/simple_with_dates_and_nulls.yaml
```diff
@@ -0,0 +1,89 @@
...
+      then: "[date_a]"
+      else: "null"
+    - name: ifthenelse
+      if:
+        column: index
+        operator: eq
+        value: 5
+      newColumn: date_c
+      then: "null"
+      else: "[date_b]"
```
lukapeschke, line 25: could you remove the brackets on this one ? Just so we have both cases (with and without brackets) in the fixture

server/tests/backends/fixtures/ifthenelse/simple_with_dates_and_nulls.yaml
lukapeschke, line 1: Could you please add a fixture for pypika as well ?

server/tests/backends/fixtures/ifthenelse/simple_with_dates_and_nulls.yaml
Fanaen: Good idea! Here it is [48c0477](https://github.com/ToucanToco/weaverbird/pull/2422/commits/48c04772f41df00b1e6564b4a19804deb0d43930).
I've done the same in pypika test: [3c8b95f](https://github.com/ToucanToco/weaverbird/pull/2422/commits/3c8b95f1185a1d80d16242180e18c6e2cdc8df5f)

lukapeschke (approved):
thanks

server/tests/backends/fixtures/ifthenelse/simple_with_dates_and_nulls.yaml
lukapeschke: Yeah you can add it, thanks

PR #2428, opened 2025-05-12
feat: remove playground feature

julien-pinchelimouroux:
It removes playground from weaverbird's backend and documentation. This is not used anymore since front-end playground have been removed.


lukapeschke (approved):
:broom: :sparkles:

PR #2429, opened 2025-05-14
fix(server/mongo):  step adds empty columns to values without match (TCTC-10693)

julien-pinchelimouroux:


server/src/weaverbird/backends/mongo_translator/steps/join.py
```diff
@@ -63,12 +75,28 @@ def translate_join(step: JoinStep) -> list[MongoStep]:
...
+                "newRoot": {
+                    "$mergeObjects": [
+                        {
+                            "$arrayToObject": {
+                                "$map": {
+                                    "input": {"$objectToArray": {"$first": "$_vqbJoinKeyMetadata"}},
+                                    "as": "metadata",
+                                    "in": {"$setField": {"field": "v", "input": "$$metadata", "value": None}},
+                                }
+                            }
```
lukapeschke, line 88: Could you please add a comment explaining what this does ? It's kinda confusing

It looks to me like you're setting a hardcoded `v` field ?

lukapeschke (commented):
Thx! looks good overall, I just need a bit of explaining :)

server/src/weaverbird/backends/mongo_translator/steps/join.py
julien-pinchelimouroux: Yes, objectToArray generates a list of `[{"k": "key name", "v": "value"}, ...]`

julien-pinchelimouroux: [316aa29](https://github.com/ToucanToco/weaverbird/pull/2429/commits/316aa298bf2096482c2e89071cbf8a9b86f55eb5)

lukapeschke (approved):

server/src/weaverbird/backends/mongo_translator/steps/join.py
lukapeschke: thx

Fanaen (approved):
Thanks for the fix and comments 🙏

PR #2446, opened 2025-06-26
feat(server/pandas): support french months in todate step [TCTC-10758]

lukapeschke:


server/src/weaverbird/backends/pandas_executor/steps/todate.py
```diff
@@ -3,6 +3,39 @@
...
+    ("septembre", "september"),
+    ("octobre", "october"),
+    ("novembre", "november"),
+    ("décembre", "december"),
+    ("decembre", "december"),
+]
+
+_FR_TO_EN_MONTHS_SHORT = [
+    ("janv", "jan"),
+    ("févr", "feb"),
```
Fanaen, line 25: I see in mongo, we support `fevr` as well. Is this relevant?

server/src/weaverbird/backends/pandas_executor/steps/todate.py
```diff
@@ -3,6 +3,39 @@
 from weaverbird.backends.pandas_executor.types import DomainRetriever, PipelineExecutor
 from weaverbird.pipeline.steps import ToDateStep
 
+_FR_TO_EN_MONTHS_LONG = [
+    ("janvier", "january"),
+    ("février", "february"),
```
Fanaen, line 8: For consistency, we probably should do like december and support without accent here too.
Or remove both, I don't see the support for those in mongo translator.

Fanaen (commented):
Thanks for the tests and extra fixes 🙏

server/src/weaverbird/backends/pandas_executor/steps/todate.py
lukapeschke: [3840820](https://github.com/ToucanToco/weaverbird/pull/2446/commits/3840820a799e1b4d875d964a558ab791918d3e85)

